PR #2, opened 2018-03-20
:sparkles: Add examples

Gavlooth:
I added  reverse and comp functions, as well as if-let macro experimentally. If you think it is appropriate, you can move them to boot.dst.
I will also implement let as monads (as it is originally is). Tell me your opinion.
I like the iter approach you  implemented there :)

### DIFF
--- a/examples/utils.dst
+++ b/examples/utils.dst
@@ -1,0 +1,56 @@
+(defn- reverse-array [t]
+  (var  n (dec  (length t))  )
+  (var reversed [])
+  (while (>= n 0)
+    (array-push  reversed (get t n))
+    (-- n))
+   reversed)
+
+
+(defn- reverse-tuple [t]
+  (def max (length t))
+  (var  n  0)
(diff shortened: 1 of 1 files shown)


bakpakin:
I don't see the if-let, but thanks for more examples! I have merged but have some suggestions.

Lastly, tuple-prepend is an O(n) operation that copies the whole tuple, unlike lisp cons, which is why I really only use it in macros. The tuple reversing function would work best by simply using the array-reverse function and converting to a tuple.

I'm not sure why you would want to have let as a monad; it would be much less efficient. It would be an interesting exercise, though. I admit I am a bit fuzzy on the formalities of monads, though.

Gavlooth:
Sorry, I mispelled, i meant if-not
I was under the impression that let was a monand due to https://github.com/khinsen/monads-in-clojure/blob/master/PART1.md i read  a cuple of years ago. Search the source code of clojure and let expands to an expression of let*.
Thanks for the review.

PR #4, opened 2018-03-22
Some issues

Gavlooth:
I notice that   (type :=) => symbol. That implies that type doesn't return the type 
binded to that symbol (cfunction maybe?).  How does binding to symbols work? Seems that := are
special symbols that are parsed differently from the other symbols ? I like the symbol only approach by the way.
My implementation of iter2lazyseq causes a segmentation fault, not sure why. Try it with the function filter2
Do you know  if lazyseq would be more efficient if it memoized more elements? I know that the clojure 
lazy-seq always realizes 32.
Also i am curious, have you in mind something similar to deftype/defprotocol ? Or maybe you have something other in mind more suitable for a lua/clojure (seems to me like that) hybrid?

### DIFF
--- a/examples/lazyseqs.dst
+++ b/examples/lazyseqs.dst
@@ -24,17 +24,17 @@ body once, and then memoizes the result."
 (def HEAD :private 0)
 (def TAIL :private 1)
 
-(defn empty-seq 
- "The empty sequence." 
+(defn empty-seq
+ "The empty sequence."
  [] nil)
 
-(defmacro cons 
+(defmacro cons
  "Create a new sequence by prepending a value to the original sequence."
  [h t]
  (def x (tuple h t))
  (fn [] x))
 
-(defn empty? 
+(defn empty?
  "Check if a sequence is empty."
  [s]
  (not (s)))
@@ -49,7 +49,7 @@ body once, and then memoizes the result."
  [s]
  (get (s) TAIL))
 
-(defn range2 
+(defn range2
  "Return a sequence of integers [start, end)."
  [start end]
  (if (< start end)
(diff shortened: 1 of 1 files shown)


bakpakin:
Any symbol that begins with the colon character is not resolved during compilation, it behaves as a keyword. The := keyword is both a special form and a keyword style symbol. The fact that it is a keyword is an unfortunate side effect that doesn’t seem to have too much of a drawback.

I will investigate the segfault when I have time later. I’m also interested in what system you are running as it may help me debug.

bakpakin:
As for defprotocol, deftype, I have been thinking about how to implement custom types and or multimethods. I am undecided about how to best implement polymorphism, and whether to implement a more Lua style (prototypes and meta tables), or a clojure, functional style (multimethods, protocols). Tables can have a prototype set with the getproto and setproto functions for simple prototypical inheritance. They are fairly straightforward to use

```
(def t1 @{:a 1 :b 2})

(def tchild (setproto @{:c 3} t1))

(get tchild :c) #> 3
(get tchild :b) #> 2
(rawget tchild :a) #> nil

```

Gavlooth:
I see, sounds reasonable. I can say that it feels comfortable  to use mutation to encapsulate local state, we do it all the time in clojure with atoms, and is also reasonable to keep things simple. that being Said, in my experience protocols and especially multimethods are a very handy tool, more  than prototypes were when i coded in javascript, but i have no experience with lua prototypes. lastly deftype and defrecord are rarely used in clojure .

Gavlooth:
I run on Debian GNU/Linux buster/sid 
 AMD FX(tm)-6300 Six-Core Processor   
with 2 8gb kingston DDR3-1600

bakpakin:
I just push a fix to a small bug I found with closure creation that had to do with byte offsets (it caused problems on my computer only when I built a 32 bit binary). Try merging from master and seeing if that helps.

I couldn't get the segfault you described on my machine, could you send the exact script that caused the segfault if the above doesn't help?

PR #5, opened 2018-03-23
Replicate segmentation fault

Gavlooth:
In the file lazyseqs.dst  I added a run-me function that  produces a segmentation fault. 
It seems also that the implementation of filter has some quirks (filter pos?  [0 -1 -2 33 -3 0 302 -3 2 8  54 3 -2 0]) will return an iterator with the first element nil and last element false

### DIFF
--- a/examples/lazyseqs.dst
+++ b/examples/lazyseqs.dst
@@ -109,15 +109,14 @@ body once, and then memoizes the result."
 #Iterators is a conscept that looks a lot like lazy seq
 #The following functions turn iterators to lazy seq and vice versa
 
-(defn- iter-self
-  [next]
-  (delay (tuple (next) (iter-self next))))
 
 (defn iter2lazy
-"Create a lazy sequence froma an iterator"
+"Create a lazy sequence from  an iterator"
   [iter]
   (def  {:more more :next next} iter)
-  (iter-self next))
+  (if (more)
+    (delay (tuple  (next) (iter2lazy iter)))
+    empty-seq))
 
 (defn lazy2iter
   "turn a lazy-seq to an iterator"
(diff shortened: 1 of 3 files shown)


bakpakin:
So running the run-me function through valgrind shows that this is a stack overflow in the garbage collector. The garbage collector uses the C stack to traverse memory (like emacs lisp) which is simple not entirely correct.

Considering the size of stuff we are working with, though, so this suggests the Garbage collector is in a loop (it is marking the children of an item that it has not yet marked). I will continue working on this.

Also, I do not see the behavior of filter that you have mentioned.

From my repl
```lisp
> (filter pos? [0 -1 -2 33 -3 0 302 -3 2 8 54 3 -2 0])
{:more <function 0x5555557E1360> :next <function 0x5555557E1330>}
> (foreach _ print)
33
302
2
8
54
3
nil
>
```

Gavlooth:
From mine:
> (def arr [0 -1 -2 33 -3 0 302 -3 2 8  54 3 -2 0])
nil
> (def filtered (filter pos? arr))
nil
> (def next (get filtered :next))
nil
> (next)
nil
> (next)
33
> (next)
302
> (next)
2
> (next)
8
> (next)
54
> (next)
3
> (next)
false
> (next)
nil
> (next)
nil
> (next)
nil
> (next)
nil
>

bakpakin:
Iterators are currently under specified I guess. It is assumed that you will always call :more to check if the iterator is finished before calling next. But this is a quirk, yes.

Gavlooth:
Got it, thanks.

Gavlooth:
What functions/features do you think there should be in the core? I could try and write some of them.

bakpakin:
It seems there is no loop in the GC - every stack frame is a unique dst object. It seems the implementation of lazy2iter is doing infinite recursion via closures. If the GC were corrected and dst didn't segfault, it would eventually run out of memory. I would guess you have tried to realize an infinite sequence.

bakpakin:
I think you can move all of the non iterator/seq functions in util.dst into boot.dst. The lazyseq and iterator functions still need work, so for now I would like to leave them out of boot.dst.

Gavlooth:
Actually is iter2lazy and yes you are write. The way it is implemented it doesnt check if the iterator is
producing nils.

Gavlooth:
Excellent!

PR #6, opened 2018-03-24
Some thoughts 

Gavlooth:
Hi, here are some thoughts

 I've been reading about metatables and i find them an interesting concept. Can the mechanism be used to 
create a more "formal" type/object ? Your thoughts on metatables make sense since there is no host language
I am curious, what is the motivation of using :keywords ? Since there is no IFn protocol, wouldn't 
a quote suffice ? like 'symbol. Or ''symbol if you want macros. Nobody uses them that way but 'symbol in clojure maps behave like :keyword (and also implement IFn protocol).
Sorry for bothering you with questions, thank you in advance.
 And something else i have in mind. It might be difficult if someone read the code to distinguish between mutable from immutable data types. The same thing happens in clojure a lot with atoms, but  a naming convention is usually enough to avoid any pitfalls and you only get dereference errors either way. Making the same mistakes with mutable data might give you more unexpected results

### DIFF
--- a/examples/lazyseqs.dst
+++ b/examples/lazyseqs.dst
@@ -109,15 +109,14 @@ body once, and then memoizes the result."
 #Iterators is a conscept that looks a lot like lazy seq
 #The following functions turn iterators to lazy seq and vice versa
 
-(defn- iter-self
-  [next]
-  (delay (tuple (next) (iter-self next))))
 
 (defn iter2lazy
-"Create a lazy sequence froma an iterator"
+"Create a lazy sequence from  an iterator"
   [iter]
   (def  {:more more :next next} iter)
-  (iter-self next))
+  (if (more)
+    (delay (tuple  (next) (iter2lazy iter)))
+    empty-seq))
 
 (defn lazy2iter
   "turn a lazy-seq to an iterator"
@@ -138,8 +137,7 @@ body once, and then memoizes the result."
 #data structures as their values are references to this
 #data structures. Same is true for iterators
 
-(defn filter2 [pred coll]
-  (tail (iter2lazy (filter pred coll))))
+(defn filter2 [pred coll]  (iter2lazy (filter pred coll)))
 
 (def arr [0 -1 -2 33 -3 0 302 -3 2 8  54 3 -2 0])
 
--- a/src/compiler/boot.dst
+++ b/src/compiler/boot.dst
@@ -1,13 +1,13 @@
 # Bootstrap the dst environment
 # Copyright 2018 (C) Calvin Rose
 
-(var *env* 
+(var *env*
  "The current environment."
  _env)
 
 (def defn :macro
  "Define a function"
-  (fn [name & more] 
+  (fn [name & more]
    (def fstart (fn recur [i]
     (def ith (ast-unwrap1 (get more i)))
     (def t (type ith))
@@ -23,7 +23,7 @@
  "Define a macro."
  (do
   (def defn* (get (get _env 'defn) :value))
-  (fn [name & more] 
+  (fn [name & more]
    (def args (array-concat [] name :macro more))
    (apply1 defn* args))))
 
@@ -86,7 +86,7 @@
 
 (defmacro when
  "Evaluates the body when the condition is true. Otherwise returns nil."
- [cond & body] 
+ [cond & body]
  (tuple 'if cond (tuple-prepend body 'do)))
 
 (defmacro cond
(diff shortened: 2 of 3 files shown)


bakpakin:
Don't worry about questions, I am more than happy to answer :smile: 

Metatables (or prototype tables) can indeed be used to create a more formal object or type system. This is how languages like Lua can implement complete OO or other type systems through metatables. Dst prototypes are much simpler than Metatables at the moment to keep things simple in the implementation, but I rather like the simplicity. However, I would definitely like to postpone this, as one idea I have been bouncing around for dst is a gradual type system for the compiler. If we go such a route, I am not sure what role metatables or prototypes would actually play in the type system, so I am keeping them very simple and just capable enough for what I need now.

Keywords are used because they evaluate to themselves, so you never need nested quotes. They are not strictly necessary, even for performance, but I think they make code clearer. Many languages have them and they don't add much complexity to the implementation or I think to the programmer.

The difference of vars vs defs is a good point. I don't think there is too much of an issue with local vars having no naming convention, as they are local and hidden, but for global vars I prefer the \*double-star\* syntax, like clojure's dynamic defs.

Lastly, dst does not intend to have the full immutability support of clojure. By making core all data structures immutable, clojure sacrifices serious performance, and more critically generates a lot of garbage. As dst will have many more mutable data types and mutating functions than clojure, it seems overwrought to suffix all such functions with a bang.

As for data literals, mutable types are in general prefixed with a @, except for arrays, which seem
to be a commonly used datatype.
```
# Mutable types
@"Hello" # buffer
@{1 2 3 4} # table
[1 2 3] # array

# Immutable types
"Hello" # string
{1 2 3 4} # struct
(1 2 3) # tuple
```

To make arrays no longer the exception, we could modify the parser to make arrays also require the @ prefix.

Proposal:
```
@"Hello" # buffer
@{1 2 3 4} # table
@[1 2 3] # array
@(1 2 3) # Also array

# Immutable types
"Hello" # string
{1 2 3 4} # struct
(1 2 3) # tuple
[1 2 3] # Also tuple
```
This seems a more consistent design. Thoughts?

Gavlooth:
Yeah i had in mind to use different function for immutable and mutable data but this implies some (a lot maybe?) unnecessary code duplication. What you suggest is clean and with var and def, could  help someone to pinpoint and identify data structures  a lot easier.
What  is still unclear to me is if nested mixtures of tables/structs are prone to errors, but you can't control
everything :p

As for keywords, i don't completely agree. In Clojure keywords are handy and convenient because they look-up themselves in a map. This is especially helpful when you have big S-exps and you are forced to 
use threading macros -> ->> as-> e.t.c. E.g. (->> data :labels :strings   (map #(zipmap labels % )... ). and this happens a lot. Of course it is true that is of no cost, and give a nice clean look in the code. 
I also noticed something else when i tried to implement memoization. {[1 2] "string"} is permitted, but is otherwise ) inaccessible (with get at least)

bakpakin:
> As for keywords, i don't completely agree. In Clojure keywords are handy and convenient because they look-up themselves in a map. This is especially helpful when you have big S-exps and you are forced to
use threading macros -> ->> as-> e.t.c. E.g. (->> data :labels :strings (map #(zipmap labels % )... ). and this happens a lot. Of course it is true that is of no cost, and give a nice clean look in the code.

I believe symbols in clojure also do self look up for their implementation of IFn. Not to say that making keywords act like functions isn't useful, but it is not the only reason for their existence.

> I also noticed something else when i tried to implement memoization. {[1 2] "string"} is permitted, but is otherwise ) inaccessible (with get at least)

Lastly, mutable types are reference types - they have equality semantics that are defined by their internal pointers. Immutable types have the expected functional semantics and work as table and struct keys as expected.

I have also just pushed the change I suggested above to master, making array literals require a leading @. This make break some code, but I think it makes the syntax more orthogonal.

Gavlooth:
Are there examples on buffers/strings?

bakpakin:
In boot.dst, the function run-context does some stuff with buffer manipulation.

You can push bytes sequences (strings, buffer, symbols) to a buffer, clear the buffer, and extract bytes from the buffer with get. Also keep in mind that, like other mutable data structures, buffers have pointer equality. You can also use string-slice and buffer-slice to get sub strings from a char-sequence as a string or buffer respectively.

One thing I would really like is something like the Lua pattern library, or maybe the [parse](http://www.red-lang.org/2013/11/041-introducing-parse.html) module from the Red programming language.

Lastly, you can create buffers, string, and symbols with the functions `buffer`, `string`, and `symbol`.

Clearly I need to write some documentation.

Gavlooth:
Is lua library more powerful than  regex? (possix or otherwise). From lua i have only worked with 
LPEG (which is awesome).

PR #10, opened 2018-03-25
Fix if-let when-let and add put-in (assoc-in)

Gavlooth:
There seems to be some problem with ast-unwrap(1) and recursion. It seems that subsequent calls and changes the argument of ast-unarwap doesn't work but i have not a clear mind right now. I might test it later.
I added a put-in (it took me more that expected to implement it due to if-let). 
Before you push your implementations of array literals, maybe we could use [] to  avoid (tuple ...) when writing macros? I think it would be very convenient.

### DIFF
--- a/examples/utils.dst
+++ b/examples/utils.dst
@@ -1,0 +1,22 @@
+
+(defn put-in [coll keys val]
+  (defn assoc [the-coll n]
+    (if-let [current-key (get keys n)
+             current-val (get the-coll current-key)]
+          (put the-coll current-key (assoc current-val (inc n)))
+      val))
+  (assoc coll 0))
+
+
+(defn update-in [coll keys an-fn]
+  (def new-keys (array-slice  coll 0 -2) )
+  (def last-key (get (array-slice  coll -1 -2) 0))
+  (defn assoc [the-coll n]
+    (if-let [current-key (get keys n)
+             current-val (get the-coll current-key)]
+      (put the-coll current-key (assoc current-val (inc n)))
+      ( update  the-coll last-key an-fn )))
+  (assoc coll  new-keys   0))
+
+
+;; (defn update-in-test [ ] (update-in @{:a "x" :b {:y {"pipa" 3}}} [:b :y "pipa"] type))
--- a/src/compiler/boot.dst
+++ b/src/compiler/boot.dst
@@ -367,13 +367,16 @@ If no match is found, returns nil"
   (tuple 'when (tuple not condition) exp-1))
 
 (defmacro if-let
-"Takes the first one or two forms in a vector and if true binds
+"Takes the first one or two forms in a vector and if both are  true binds
  all the forms with let and evaluates the first expression else
  evaluates the second"
   [bindings then else]
-  (def head (ast-unwrap1 bindings))
-  (tuple 'let head
-         (tuple 'if (and (get head 1) (if (get head 2) (get head 3) true))
+  #(tuple 'print   (tuple '> (tuple 'length bindings) 2))
+  (tuple 'let bindings
+         (tuple 'if (tuple 'and (tuple 'get bindings 1)
+                           (tuple 'if
+                                  (tuple '> (tuple 'length bindings) 2)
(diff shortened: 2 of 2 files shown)


bakpakin:
As you may have noticed, arguments to macros maybe wrapped in an abstract type parse.ast, which contains source mapping information. This is a a result of implementation, ideally, we would have a clojure-like metadata attribute for all values, but this is inefficient (it requires boxing primitives to support pointer equality).

ast-unwrap1 simply gets the data inside the parse.ast. ast-unwrap does the sam but then recursively unwraps data structures as well (arrays, tuple, structs, tables).

> Before you push your implementations of array literals, maybe we could use [] to avoid (tuple ...) when writing macros? I think it would be very convenient.

It would be convenient, but I don't think it is a good idea for a number of reasons
- It would mean square brackets could not be used for fn forms, let, etc.
- Dst syntax would have that much more "magic"

What I think would be even nicer is syntax quoting and unquoting, which can be implemented via macros.

I also don't see put-in in the changed files.

Gavlooth:
I see. Ok yes Rich Hickey must have new what he was doing with clojure macro implementation :P 
I created a new file examples/core.dst, which apparently is gitignored :S I renamed  it to util.dst again

bakpakin:
> Ok yes Rich Hickey must have new what he was doing with clojure macro implementation :P

Yeah :P. In thoughts.md, one of the topics is how to remove parse.ast while still retaining the ability to do source mapping with macros. I think it's possible, but will require some changing of the compiler.

Gavlooth:
About the metadata.  Clojure supports metadata on symbols and data structures, symbols are not primitives so a presume  (I don't have good knowledge of the jvm or vm's anyway) that are already boxed and stored to the heap? On the other hand lua's (and yours) vm is registered based, and all data structures are  stored in registers?

bakpakin:
In the JVM, everything is an object except primitives, which all have corresponding boxed representation. When the clojure parser parses source code, it uses these boxed representations of integers, doubles, booleans, nil, strings, etc. This allows every form to be a unique instance, and so the meta information can just be kept in an external data structure that takes any Object as a key and returns its metadata (I’m just guessing here, but it seems likely). In dst, the parser returns numbers in an unboxed form, so after parsing, two instances of the same number are completely identical - when generating bytecode source mapping for a number, symbol, or Boolean, the compiler has no way to tell where in the source code that particular primitive was created, unless we pass around the source mapping information with said primitive (hence the parse.ast wrapper).

There are ways around this, for example only keeping track of source mapping information for data structures (arrays, tuples, structs, tables), and being slightly less precise when mapping each bytecode instruction to a source code location. I would like to try and make this change in the compiler and see if I can remove the parse.ast abstraction.

While source mapping is not currently seen in the repl when errors occur, if you disassemble the bytecode of a function in boot.dst (using disasm), you should hopefully see that there are a pair of byte offsets (start and end offsets) for each bytecode instruction that indicate which form in the source file emitted that instruction during compilation.

Gavlooth:
I see, so we should be able to add metadata to function arguments, for example but not to the symbols. Would too much of an overhead if you added the metadata to symbols, like in clojure? (of course you have already thought of that but i would like to understand the motivation on the decisions)

bakpakin:
Symbols are all interned for speed, as they are also used like keywords. This causes the same problem as adding metadata to primitives like numbers.

For a more concrete example, see the following snippet:

```clojure
(defn fun [x y]
  (if x x y))
```

In the above snippet, there are 3 instances of the symbol x. Because symbols are interned, after parsing, all x symbols are simply pointers to one location, 0x1234; they are identical. When the current compiler is compiling an 'x' symbol, there is no way of telling where in the source this X symbol came from because macros can freely transform the source code, and all x's are exactly the same, so they cannot have unique metadata without a wrapper.

To allow metadata, each symbol would need to be a pointer to a unique location, say 0x2222, 0x3333, and 0x4444, and equality checking suffers a little bit. It wouldn't be too much overhead, but I would rather not as this would require many changes to the runtime. I think there is a better solution to source mapping that I will pursue.

PR #31, opened 2019-02-09
Crazy brackets

bakpakin:
Allow using square brackets as a tuple constructor rather than a tuple literal.


honix:
I like this decision. And i don't like clojure's random ( ) and [ ] syntax.

PR #40, opened 2019-02-16
string/format  

jfcap:
Adding string/format function 

This patch provides string/format function.
A snprintf like function with some specialization for Janet.
%p (for pretty printing Janet values) etc.
This code is mostly copy/paste from lua/lstrlib.c

I wrote this code because I'm really need it.
But I'm not sure it should be included in core janet source.
(a dedicated extern module can do the job of course.)

```
janet:329:> (def o {:k 1 :v [1 2 3]})
{:k 1 :v (1 2 3)}
janet:402:> (print (string/format @"" "pi=%8.5f\nobject : %.2p" math/pi o))
pi= 3.14159
object : {:k 1 :v (...)}
nil
```

### DIFF
--- a/src/core/buffer.c
+++ b/src/core/buffer.c
@@ -317,6 +317,14 @@ static Janet cfun_buffer_blit(int32_t argc, Janet *argv) {
     return argv[0];
 }
 
+static Janet cfun_buffer_format(int32_t argc, Janet *argv) {
+    janet_arity(argc, 2, -1);
+    JanetBuffer *buffer = janet_getbuffer(argv, 0);
+    const char *strfrmt = (const char *) janet_getstring(argv, 1);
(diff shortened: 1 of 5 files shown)


bakpakin:
This is great! I wanted to add this for a while but never got around to it - it should definitely be in the core source. Interfaces like `string/number` were easy to write in a safe manner but are a pain to use and too verbose.

bakpakin:
Perhaps there should be two functions - `string/format` and `buffer/format` so we don't need to pass a buffer argument to string/format unless we actually want to use the buffer. `string/format` could basically wrap buffer/format but provide the empty buffer and convert to a string.

jfcap:
It will be better. Feel free to modify it as you like.

PR #56, opened 2019-02-27
Register core/file abstract type

jfcap:
Register `core/file` abstract type to allow acccess to janet file object in native C module.

For example in a native module we can do something like :

```C
struct IOFile {
    FILE *file;
    int flags;
};

FILE * get_janet_iostream(const Janet *argv, int32_t n) {
  const JanetAbstractType *iof_at=janet_get_abstract_type(janet_ckeywordv("core/file"));
  if (iof_at==NULL)
    janet_panic("core/file abstract type  not registered");
  struct IOFile * iof =(struct IOFile *)janet_getabstract(argv,n,iof_at);
  return iof->file;
}
```

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -416,4 +416,7 @@ void janet_lib_io(JanetTable *env) {
     janet_core_def(env, "stdin",
                    makef(stdin, IO_READ | IO_NOT_CLOSEABLE | IO_SERIALIZABLE),
                    JDOC("The standard input file."));
+
+    janet_register_abstract_type(&cfun_io_filetype);
+    
 }
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -400,6 +400,14 @@ static const JanetReg io_cfuns[] = {
     {NULL, NULL, NULL}
 };
 
+/* C API */
+
+FILE *janet_getfile(const Janet *argv, int32_t n, int *flags) {
+    IOFile *iof = janet_getabstract(argv, n, &cfun_io_filetype);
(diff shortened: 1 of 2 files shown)


bakpakin:
I thought `janet_register_abstract_type(&cfun_io_filetype)` registered a type for marshalling, and we don't want to be able to marshal files, as they contain too much state that we can't see. If we want to expose the FILE * in IOFile structs, we can just write a new function `janet_getfile(const Janet *argv, int32_t n)` in io.c that wraps `janet_getabstract`.

jfcap:
Of course, my objective is not to marshal file objects.

In my point of view the abtract type "registering mecanism" provides some abstraction benefits not directly related to marshaling.
One of these is the ability give access in a clean and generic way to the JanetAbstractType descriptor to C native module developper  without exposing an adhoc specific API.

Of course, in the case of janet io file oject , adding just a simple API function to get the FILE* descriptor
as you propose will certainly do the job (BTW, an other function to access the flags would be handy as well).

But in more general case (for example when an Abstract type is defined in an extern native module), it will not be possible without providing a specific extern C API (with an adhoc header file, etc...)

But again, in the specific case of janet io file oject, it's more "a matter of taste".
I just find "elegant" to provide access to module developpers to all the interface with only one call to `janet_register_abstract_type()` without the need to extend de C API endlessly.

Nevertheless, for my specific need here your proposition is certainly ok.
Feel free to decline this PR.

jfcap:
I've changed the PR to meet your proposal.

PR #71, opened 2019-03-14
Bigint 

### LINKED ISSUE #67: uint64 & int64 Typed Arrays
Since 0382dc9 commit, uint64_t & int64_t typed arrays support have been removed.
Why ?
I know we don't have (yet ?) full proper way to manipulate 64bit integers directly in janet (without truncation) but a lot of numerical C libraries use 64bit integers arrays and it's handy (and efficient) to be able to pass directly such binary data as "standard" janet objects to them.

jfcap:
Fixes #67 

Bigint for Janet implementation proposal

- added core/int64 & core/uint64 bigint boxed types with basic arithmetic and logic methods.
- added (u)int64 typed arrays back.


bakpakin:
Another case that triggers an exception is INT64_MIN / -1, which affects all signed integer types on most architectures - we should probably try to detect this.

jfcap:
Fixed in https://github.com/jfcap/janet/commit/c5217da10906b36ad3727098e43fbbc110b5787d

```
janet:0:> (def min (bigint/int64 "-0x8000000000000000"))
-9223372036854775808
janet:47:> (:/ min -1)
error: INT64_MIN divided by -1
  in <cfunction>
  in _thunk [repl] (tailcall) at (48:58)
janet:73:> (:% min -1)
error: INT64_MIN divided by -1
  in <cfunction>
  in _thunk [repl] (tailcall) at (74:84)
```

bakpakin:
Thanks for all the changes! I will give these changes a spin soon and probably merge everything if there are no immediate issues.

Just a minor thing about names -- "bigint" usually implies python/GMP style integers of unlimited precision. These 64 bit integers have a different role, mainly providing an interface to C APIs where 64 bits integers are used.  Perhaps we could rename the module to just `int64`, rename the constructors `s` and `u` (signed and unsigned), and name the types something like `s64` and `u64`, linux kernel / rust style? I think it would better signal the intent of the module to newcomers.

jfcap:
Feel free to rename everything as you think it's better.

bakpakin:
A bit of an explanation for the changes I have made:

We can still use and create typed arrays with 64 bit integers, even if JANET_NO_INT_TYPES (previously  JANET_NO_BIGINT) is set. However, we can't get and set in them from Janet without the `int` library enabled. This also means that the marshaling format is consistent, regardless of how compiler defines are set.

I have also made typed array views all share a single abstract type. This is kind of a large change for this branch, but it made code cleaner. The main difference for a user is typed arrays now all print out as `<ta/view 0x12345678>` when printed and share a type `:ta/view`. This has ups and downs, but I want to avoid creating many types, especially in the core.

jfcap:
Thank you or the merge.

After your changes there is some "dead code" in pp.c

```C
        case JANET_ABSTRACT: {
#ifdef JANET_BIGINT
            JanetBigintType bt = janet_is_bigint(x);
            if (bt == JANET_BIGINT_TYPE_int64) {
                int64_to_string_b(buffer, *(int64_t *)janet_unwrap_abstract(x));
                return;
            }
            if (bt == JANET_BIGINT_TYPE_uint64) {
                uint64_to_string_b(buffer, *(uint64_t *)janet_unwrap_abstract(x));
                return;
            }
#endif
```

The good way to correct this is probably to implement your proposal  #73

PR #84, opened 2019-05-03
Add string/has-prefix? and string/has-suffix?.

andrewchambers:


andrewchambers:
Don't mind if they don't pass the usefulness bar, but I tend to use this quite a lot doing string munging.

bakpakin:
I actually use this a lot in mendoza but do it manually for checking file extension. Could also b ednoe with pegs, but this is more efficient.

PR #85, opened 2019-05-08
Add buffer/push-buffer.

andrewchambers:


andrewchambers:
I just found out push-string can be used for this purpose, perhaps we could alter the docs to explain this?

PR #97, opened 2019-05-15
Add support for Haiku

Crestwave:
Other defaults could be set in the Makefile, like the default installation path, but it would probably mess up the structure. This gets it to build successfully by default on 64-bit Haiku (instructions for 32-bit are included).


bakpakin:
Neat!

PR #102, opened 2019-05-16
Add static library

PaulBatchelor:
This PR adds some recipes to the makefile to produce a static library archive via ar. I figured it would be a useful thing to have.

Let me know if there are any problems with it.


bakpakin:
Looks good to me, I have verified it links properly for a simple program.

PR #107, opened 2019-05-19
Add a 'compile-only' flag to the command line

ALSchwalm:
This allows syntax checkers like the emacs [flycheck](https://www.flycheck.org/en/latest/) to check the source without side effects. Currently, I am just not executing the REPL when this option is supplied, but I am open to suggestions on what to do there.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1458,6 +1458,7 @@
   :env - the environment to compile against - default is the current env\n\t
   :source - string path of source for better errors - default is \"<anonymous>\"\n\t
   :on-compile-error - callback when compilation fails - default is bad-compile\n\t
+  :compile-only - only compile the souce, do not execute it - default is false\n\t
   :on-status - callback when a value is evaluated - default is debug/stacktrace\n\t
   :fiber-flags - what flags to wrap the compilation fiber with. Default is :ia."
   [opts]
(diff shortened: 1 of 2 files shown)


src/boot/boot.janet
```diff
@@ -1458,6 +1458,7 @@
   :env - the environment to compile against - default is the current env\n\t
   :source - string path of source for better errors - default is \"<anonymous>\"\n\t
   :on-compile-error - callback when compilation fails - default is bad-compile\n\t
+  :compile-only - only compile the souce, do not execute it - default if false\n\t
```
kurayama: default `is` false

ALSchwalm: Good catch. Fixed and rebased.

bakpakin:
Looks like a good start. There are some practical limits to this as is, but the general idea should work fairly well.

For example:

```clojure
(defn func1
  [x]
  (+ x x 2))

(defn func2
  [y]
  (* (func1 y) (func1 y)))
```

Will trigger a compilation error with the current implementation as func1 is never added to the environment, which happens at runtime. 

I will merge this for now, and I have some ideas on how to make this more useful but I'm not sure what is the best way to proceed.

bakpakin:
It seems I spoke too soon, the above example does work. However, imports or anything like that will not work.

PR #114, opened 2019-05-28
Add api for checking build compatibilty.

andrewchambers:
The api is still not used automatically by modules as that would be a breaking change.

### DIFF
--- a/src/core/capi.c
+++ b/src/core/capi.c
@@ -26,6 +26,16 @@
 #include "fiber.h"
 #endif
 
+static JanetBuildConfig *api_build_config = &(JanetBuildConfig){
+    .api_version = JANET_API_VERSION,
+    .single_threaded = JANET_SINGLE_THREADED_BIT,
+    .nanbox = JANET_NANBOX_BIT
+};
+
+const JanetBuildConfig *janet_build_config() {
+    return api_build_config;
+}
+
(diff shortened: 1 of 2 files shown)


andrewchambers:
The logic behind the api flag is in the rare case symbols don't change but the Janet structs do.

bakpakin:
There seems to be multiple GCCisms in this patch (I'm hesitant to add very non-standard C) . We should probably switch to a plain int and use bitflags instead of GCC bitfields.

```c
#define JANET_NANBOX_BIT 0x1
#define JANET_SINGLE_THREADED_BIT 0x2
// ect.

typedef struct {
    int api_version;
    int flags;
} JanetBuildConfig;

#define janet_api_compatible() \
  ((janet_api_build_config()->api_version == JANET_API_VERSION) \
   && (janet_api_build_config()->flags == (JANET_NANBOX_BIT | JANET_SINGLE_THREADED_BIT)))

// ...
```
As for this construct:

```c
static JanetBuildConfig *api_build_config = &(JanetBuildConfig){
    .api_version = JANET_API_VERSION,
    .single_threaded = JANET_SINGLE_THREADED_BIT,
    .nanbox = JANET_NANBOX_BIT
};
```
I don't believe it will compile in MSVC.

PR #123, opened 2019-06-07
Add file/{fdopen,fileno} functions.

andrewchambers:


bakpakin:
Looks pretty good. However, windows docs say to use _fdopen and _fileno respectively. The non prefixed versions are deprecated.

andrewchambers:
Those seemed to be windows C++ docs, but I wasn't sure. _fdopen didn't compile on my machine.

bakpakin:
Yes, it will probably need an #ifdef JANET_WINDOWS

andrewchambers:
fixed

xigoi:
Why were these functions removed?

sogaiu:
I don't know but for reference, here's the commit where they were removed: https://github.com/janet-lang/janet/commit/b8d530da36d01cfd8e5b05879bfb5a51d8a1ce05

andrewchambers:
because they are useless without additional C functions to provide file descriptors.they don't make sense standalone.

Issue #121, opened 2019-06-04
API for extracting partially parsed state from parser.


andrewchambers:
Currently we are able to get what nesting level the parser is at, when implementing repls and completions it would be useful to be able to inspect the partially parsed form values themselves.

e.g. be able to inspect what the current function symbol is

ALSchwalm:
I'm interested in implementing this. Is it acceptable to break the interface for `parser/state` or would this be another function?

Given a parser buffer of "((", I'm imagining something like:

```
> (parser/state p)
@{:wrappers "((somthing" :start 2}
```

So `start` returns the starting position of the body of the inner most s-expression (or array, string, etc). It is absent when ':wrappers' is the empty string. Alternately, it could be an array of the starting position of each 'wrapper'. Either way, returning a table here would also allow us to return more parser state info in the future without breaking the interface.

I'm not sure that actually trying to return the first symbol itself is necessary, as that could sometimes be another s-expression. I'm really kind of just basing this off of emacs `parse-partial-sexp` because I'm familiar with it. Definitely open to other ideas.

bakpakin:
I like the idea of extending `parser/state` with more info. I think it would make sense for it to take a second argument, a set of flags that indicate what state to get. If a single flag is passed, return that state value only, otherwise a table of all parser state. Also, `(parser/status p)` could be an alias for `(parser/state p :s)`. Another alternative is to use full keywords like `(os/stat)`

ALSchwalm:
:+1: I like the idea of passing optional keywords. Currently I'm thinking of:

```
:wrappers - The current value returned by parser/state. Not sure about the name here
:start - The 'start' of the innermost s-expression (one past the opening paren, bracket, etc)
:prefix - The modifying character (if any) before the innermost s-exp (so a `'` for quoted s-expt, '@' for tables, etc)
:in-comment - Returns whether the parser is in a comment
```

Note that I haven't really looked at the difficulty of implementing these other than 'start' which is trivial. I may want to drop some of these if it would require large changes to the parser.

bakpakin:
I don’t think any of those would require much change to the parser. However, I think the state could be much more general, such as getting the current characters in the buffer, all current values in the current form, the kind of value currently being parsed, etc.

ALSchwalm:
That seems doable. There are a few areas where we will need to make decisions though. For example, if the parse is fed something like:

```
@{ :foo
 ```

[...]

PR #124, opened 2019-06-13
Add support for getting more detailed parser state 

ALSchwalm:
Implements #121

I think this could use some more tests, but I wanted to go ahead and get it up for review. I'm not 100% sure the `memcpy` on 877 is sound, as I'm not exactly sure of the gc rules for Janet. Also not sure about the key names, if you have an suggestions, I'm open.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1788,7 +1788,7 @@
   (default chunks (fn [buf p] (getline (string "repl:"
                                                (parser/where p)
                                                ":"
-                                               (parser/state p) "> ")
+                                               (parser/state p :delimiters) "> ")
                                        buf)))
   (default onsignal (fn [f x]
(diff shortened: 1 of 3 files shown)


bakpakin:
There is a good bit changed here so give me some time before I merge it, but the memcpy is indeed sound.

ALSchwalm:
:+1: Sure thing, no rush

bakpakin:
Looks really good!.  As for names, I prefer something like  `:delimiters` or `:opened-delimiters` instead of `:wrappers`, and `:frames` is fine. Aside from that, this seems ready to merge, so update that and I can merge this.

bakpakin:
Also, install `astyle` and run `make format` on your changes.

ALSchwalm:
Thanks for the review! Fixed and rebased

PR #139, opened 2019-07-09
Remove amalg.janet dependency on os/date

Barakat:
When compiling Janet with `JANET_REDUCED_OS`, `os/date` will not be available which breaks the tool amalg.janet. One can check file modification time on the filesystem instead.


bakpakin:
Fair enough, version and build info is more important anyways.

PR #147, opened 2019-07-28
Update documentation

curist:
* Remove repeated `bridging`
* Update `math/log` docstring


bakpakin:
Thanks for the fix! Those typos must have been there for a while. I think it should be "supports bridging" or something like that.

PR #153, opened 2019-08-05
Update several docstrings.

curist:
docstrings for take-while / take-until and drop-while / drop-until seems to be inverted?


bakpakin:
Thanks, merging.

PR #154, opened 2019-08-05
Add take/drop.

curist:
Adds `take` and `drop` functions.
**take**
When taking number bigger than `xs` length, it returns all elements in `xs`.
When n less or equal to zero, returns empty values.
**drop**
Drop number bigger than `xs` length, returns empty value.
Dropping n less or equal to zero, returns the copy of original `xs`.

Both will error out when taking non-indexed types.

## motivation
These 2 functions are likely to be common operations on indexed types.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -702,6 +702,20 @@
   (def i (find-index pred ind))
   (if (= i nil) nil (get ind i)))
 
+(defn take
+  "Take first n elements in an indexed type. Returns new indexed instance."
+  [n xs]
+  (def [f empty] (case (type xs)
+                   :array [array/slice @[]]
+                   :buffer [buffer/slice @""]
+                   :string [string/slice ""]
+                   :tarray [tarray/slice @[]]
+                   :tuple [tuple/slice (if (= :parens (tuple/type xs)) '() [])]
+                   (error "should provide an indexed type")))
+  (def len (if (= :tarray (type xs)) (tarray/length xs) (length xs)))
+  (def end (if (pos? n) (min n len) 0))
+  (f xs 0 end))
+
 (defn take-until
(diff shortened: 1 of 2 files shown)


bakpakin:
I have no problem with adding something like this, just a few things.

1. Unfortunately, we can't use `tarray/*` functions like this in boot.janet. Since janet can be compiled without typed array support, these functions are not always defined (similar for `peg/*`, most `os/*` functions, etc). You should probably remove the typed array support in the functions. (In the future, typed arrays may support `length`, as well as a `:slice` method so support could more easily be added without needing to reference the functions directly).

2. We can also easily support keywords and symbols as well (they are also indexed types). Most functions in Janet's core that can accept strings can also accept symbols and keywords. Most functions in the core library don't bother preserving the exact input type. It would be idiomatic to do
something like (and less code):

```clojure
(defn take
  "Take first n elements in an indexed type. Returns new indexed instance."
  [n xs]
  (def use-str (bytes? xs))
  (def f (if use-bytes string/slice tuple/slice))
  (def empty (if use-bytes "" ()))
  ...
```

(Other functions like `drop-while`, etc. should probably do this too! See `partition` for an example).

3. No need to check brackets on tuples. In fact, almost all functions should completely ignore that
there are different types of tuples. By default, all tuples have parens. The cases where it matters (macros), we should be explicit about it (`tuple/brackets`). I kind of regret having multiple tuple types, but oh well, they have they're uses. This is related to point 2.

curist:
That's some great feedback; I will need some time to tidy this up.

curist:
One question, I see generally you use `ind` as arg name when referring to `indexed type`, is that the preferred name? And is there a more general term to refer `indexed type`, in `clojure` they seem to call it `coll` or something.

PR #156, opened 2019-08-06
Take and drop

curist:
Cont #154 
Did something stupid to my fork branch.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -704,20 +704,22 @@
 
 (defn take
   "Take first n elements in an indexed type. Returns new indexed instance."
-  [n xs]
-  (def use-str (bytes? xs))
+  [n ind]
+  (def use-str (bytes? ind))
   (def f (if use-str string/slice tuple/slice))
   # make sure end is in [0, len]
-  (def end (max 0 (min n (length xs))))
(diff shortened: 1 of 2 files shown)


curist:
I checked how `partition` is implemented, but I believe using `slice` will always get us a new array?
And, you probably already well aware of this, this PR will change the behaviors for `take-while` and friends, thus is a breaking change.

bakpakin:
Yes, `*/slice` will always create a new array/string/buffer/etc, even if nothing was cut-off. I don't think this is a big deal. The changes to take-while, etc. also look acceptable.

curist:
there is one thing that I would love to have your input
```
(assert (deep= (take-until |(= $ 115) "books") "book") "take-until 6")
```
`find-index` function uses `(get ind i)`, and for string, that get us the char code of the character; is there a better way to do this? `((string/bytes "s") 0)` won't look too great imo, but probably better than that literal `115` right there?

bakpakin:
You can index into strings, buffers, keywords, and symbols directly to get at the contained bytes.

```clojure
(assert (deep= (take-until |(= $ ("s" 0)) "books") "book") "take-until 6")
```

curist:
Cool, didn't know this trick!

PR #157, opened 2019-08-08
Feature/alternative splice char

ahungry:
I have strong doubts you would consider this request, as I'm sure the original decisions that went into parse syntax were a combination of your personal preference + technical reasons, but I'd like to propose Janet re-work the parse syntax to be more lisp-like as defined in this pull request.

The reasons in favor (in my opinion) are:

- Strong / better integration with existing "lisp" (Emacs lisp, Common Lisp, Clojure, Scheme) like tooling systems
- More familiar / less jarring syntax for people coming from said lisp environments
- More readable / less surprising symbols for new-comers (in most non-lisps, semi-colon tends to delimit a line or statement, and in lisps it is the comment character.  In both cases the semi itself has very little "effect" on the evaluation results of the code.  In Janet, it is very impactful.

The overall change set should be as such:

- Splice character changed from `;` to `@` (parity with *-lisp)
- Mutable data type prefix changed from `@` to `#` (parity / similarity with other lisps using that as a value to indicate what succeeds it is a unique datatype or form)
- Comment character changed from `#` to `;` (much more lisp-like)

Thoughts?

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -11,7 +11,7 @@
   "(defn name & more)\n\nDefine a function. Equivalent to (def name (fn name [args] ...))."
   (fn defn [name & more]
     (def len (length more))
-    (def modifiers @[])
+    (def modifiers #[])
     (var docstr "")
     (def fstart
       (fn recur [i]
@@ -62,14 +62,14 @@
   "Dynamically create a global def."
   [name value]
   (def name* (symbol name))
-  (setdyn name* @{:value value})
+  (setdyn name* #{:value value})
   nil)
 
 (defn varglobal
   "Dynamically create a global var."
   [name init]
   (def name* (symbol name))
-  (setdyn name* @{:ref @[init]})
+  (setdyn name* #{:ref #[init]})
   nil)
 
 ; Basic predicates
@@ -197,7 +197,7 @@
   (if (odd? (length bindings)) (error "expected even number of bindings to let"))
   (def len (length bindings))
   (var i 0)
-  (var accum @['do])
+  (var accum #['do])
   (while (< i len)
     (def {i k (+ i 1) v} bindings)
     (array/push accum (tuple 'def k v))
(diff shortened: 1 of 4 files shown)


ahungry:
Also, the crux in wanting to make this set of changes is that Emacs "Lispy" mode (a context based paredit-like) plays extremely well with Emacs Lisp, Common Lisp, Clojure, Scheme, Hy (and other lisp-likes) but falls apart badly on Janet.

While that one package can likely be made compatible (if I make a PR or the author chooses to) I think running into walls with Janet syntax being distinct, where this altered syntax would "just work" with existing lisp-like ecosystems/tools.

bakpakin:
If you want to change something like this, you also have to change all of the examples, existing documentation, libraries, etc. Also, emacs is certainly capable of handling a few different characters in the syntax file. There a few Janet modes for emacs out there already.

I personally don’t use emacs, but I can appreciate the desire to have Janet just work with emacs and existing modes. The # for comments is just to make #! work better for scripts without special handling in the  parser. The changes you have proposed will break tools like jpm unless special handling is added. It’s sort of the standard for comments in the shell/Unix world, rather than the Lisp and assembly worlds.

You are right, I’m probably not going to merge this. Having Janet work out of the box with scheme or Common Lisp syntax would be nice but was not really a goal.

At the end of the day, all lisps are pretty different. Pretty much every dialect is going to want their own syntax files (modes) anyways. People will complain always complain about things like indexing from 0-1, tabs vs spaces, why does Lua use two dashes for comments, etc. I don’t really think it matters that much.

bakpakin:
Janet does have support for multiple language frontends, much like Racket's #lang directive. These can be used to specify an alternative syntax, like you described, as well as provide different methods for loading these files. For example, you could make a loader that parsed .lisp (really Janet flavored lisp) files into Janet data structures and evaluated them. See [this doc page](https://janet-lang.org/docs/modules.html) on modules and specifically on custom loaders. For a very complete example of this, see [mendoza](https://github.com/bakpakin/mendoza/blob/master/mendoza/markup.janet) for an example of a custom DSL loader that let's normal Janet code import parsed documents from .mdz files just using `(import mydoc)`. A .lisp parser could probably be the same size or even smaller than the linked implementation.

I'm just pointing out that this is possible, not recommending it.

PR #164, opened 2019-09-09
Defuse tarbomb: wrap tar file contents in directory, fix #163

### LINKED ISSUE #163: Tar archives expand without containing directory (tarbomb)
Just tried the latest release on Linux and noticed that the tar file expands its files in the current directory instead of into a new subdirectory.
It might be better to wrap it.

I’ve prepared a pull request that I’ll submit after this report.

AlbertoGP:
https://en.wikipedia.org/wiki/Tar_(computing)#Tarbomb
http://www.linfo.org/tarbomb.html

The `--transform` option requires GNU tar: BSD tar uses the option `-s` instead, and its value must omit the initial `s`.
If GNU tar is not an acceptable requirement, another way would be to create a temporary `janet-<version>` directory and put the files there before archiving it.

Edit: add link to issue #163

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,7 +255,10 @@ build/janet-%.tar.gz: $(JANET_TARGET) \
 	src/include/janet.h src/conf/janetconf.h \
 	jpm.1 janet.1 LICENSE CONTRIBUTING.md $(JANET_LIBRARY) $(JANET_STATIC_LIBRARY) \
 	build/doc.html README.md build/janet.c
-	tar -czvf $@ $^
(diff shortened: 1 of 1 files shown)


bakpakin:
Try to make you tar command work for BSD and macos. bsdtar has different options than GNU tar. Another option is to manually extract the needed files into a directory like `build/janet-blah/...`, cd into build, and create an archive there.

AlbertoGP:
All right, the current version works on Linux with both tar and bsdtar (libarchive-tools package in Ubuntu), and on MacOS X.

bakpakin:
Looks good, merging.

PR #167, opened 2019-09-11
Correct old name all-symbols to all-bindings in README

ato:
```
janet:0:> (all-symbols)
compile error: unknown symbol "all-symbols" at (1:13) while compiling repl
```


bakpakin:
Thanks, good catch.

PR #176, opened 2019-11-17
Add support for delete key in repl

staab:
Just a tiny case statement that I noticed was missing from the repl when I went to use the delete key.


bakpakin:
Looks good to me, thanks. Merging.

PR #177, opened 2019-11-23
Minor fixes for parser.

andrewchambers:
Check length before dereferencing buffer in tokenchar.
Check keywords are valid utf-8.
Fix minor typos.


bakpakin:
Looks good. However, I don’t think there is a need to check for the empty symbol in the parser. How could that possibly be triggered?

andrewchambers:
Yeah, I had wondered if it was just dead code.

bakpakin:
Yeah, I think it was just dead code. By checking if the first character is a colon, we are assuming length > 0. This makes sense, as some character needs to be read to get the parser to this state.

bakpakin:
Otherwise, utf-8 checking looks good. Please remove the old, dead code and adding a test case for parsing invalid utf-8 in a keyword in one of the test suites would be great.

andrewchambers:
Cool, will do.

andrewchambers:
Added the tests.

bakpakin:
Solid, thanks!

PR #182, opened 2019-11-28
Add an optional finalizer to scratch resources.

andrewchambers:
A finalizer can be attached to scratch allocations efficiently at any point in
it's lifecycle via janet_sfinalizer. Care was taken to keep allocations aligned
with  platform alignment requirements.

A big drawbacks to this approach is the waste of up to 16 bytes per scratch
allocation in the case the scratch memory does not require a finalizer.


bakpakin:
I like it - I don’t think the overhead is really that bad. The bump allocator might make sense, though, to avoid hitting malloc so much.

PR #185, opened 2019-11-29
Add afl fuzzing helpers.

andrewchambers:


andrewchambers:
Not fussed if this isn't merged, just thought I'd put it out there.

andrewchambers:
Perhaps these helpers would work nicely as a jpm library that can be used across multiple projects. (I also want to fuzz the json parser.)

bakpakin:
Thanks a bunch, it’s already proven itself in the parser so there fools are certainly useful.

PR #191, opened 2019-11-29
Rename quickbin to quick-bin for consistency.

andrewchambers:
Take it or leave it, I can see this might be confusing if it produces a quick bin, or it quickly produces a bin.


bakpakin:
You should also update jpm.1 with change.

bakpakin:
Also tests seem to fail when testing ‘jpm quickbin’, so the Makefile and build_win.bat probably need updating

andrewchambers:
will fix eventually haha.

PR #192, opened 2019-11-30
Add os/cryptorand

andrewchambers:


andrewchambers:
I was only able to test on linux, so forgive any failed builds.

src/core/os.c
```diff
@@ -497,6 +498,45 @@ static Janet os_cwd(int32_t argc, Janet *argv) {
...
+    /* We should be able to call getrandom, but it doesn't seem  to be uniformly supported
+       on linux distros... */
+    int randfd = open("/dev/urandom", O_RDONLY);
+    if (randfd < 0)
+        janet_panic(errmsg);
+    size_t remaining = buffer->count;
+    while (remaining) {
+        ssize_t n = read(randfd, buffer->data + (remaining - buffer->count), remaining);
+        if (n <= 0) {
+            close(randfd);
```
andrewchambers: @bakpakin To be honest, I'm not sure what to do in the case of EINTR on these calls, especially close, if its ok to just ignore them.

andrewchambers:
The OS file is getting a bit messy with ifdefs, I wonder if theres a better way.

bakpakin:
os.c is definitely the messiest file, as we are trying to abstract a bunch of random os level apis from different systems into one api.

You can add .c files to the source code, you just need to update things in 2 places, The makefile and meson.build. The amalgamation should pick these up too.

.h files are similar but tricker ,you also need to update build_win.bat for the windows amalgamation. Also it’s tricky because the headers need to be in the correct order for that.

I have optimized the flow more for end users than for devs, apologies for that. It would be nice to go fully to meson, but sometimes Make is just so convenient and easy.

andrewchambers:
to be honest, I think a handwritten and short configure script is far nicer than meson. meson is not really a mainstream tool yet and I would advice against relying on it 100 percent.

bakpakin:
> to be honest, I think a handwritten and short configure script is far nicer than meson. meson is not really a mainstream tool yet and I would advice against relying on it 100 percent as I think you will lose users,

Yup, precisely why I haven't fully switched to meson. Having written a bash script, a Makefile, a CMakeLists.txt, and meson.build file, the meson build file took far less time than any of the others and offers the most features (perhaps besides CMake), including correct cross compilation and the equivalent of a configure script.

However, a ./configure script would be nice. The problem is making it short :).

bakpakin:
Have you seen the windows api function `rand_s`? https://docs.microsoft.com/en-us/previous-versions/sxtz2fa8(v%3Dvs.140)

It might make things work without binding to crypt.lib or whatever, and docs say its cryptographically secure and uses entropy from the OS.

andrewchambers:
Nice, I can give it a try.

andrewchambers:
@bakpakin worked perfectly.

bakpakin:
Looks pretty good, but I think there is no need in include wincrypt.h anymore? Also, you test in the suite is incorrect (`inc` is a pure function, to update a var use `++`).

PR #193, opened 2019-12-01
Add truthy? to core.

andrewchambers:


bakpakin:
Seems pretty benign, and useful for coercing to a boolean. I would usually use something like `|(not (not $))`, but this is cleaner.

andrewchambers:
Yeah, I figured if 'true?' is there, 'truthy?' seems more useful. Actually I made an error while programming earlier assuming 'true?' WAS 'truthy?'

PR #195, opened 2019-12-01
Correct 'in' documentation.

andrewchambers:
(side notes)
- It is very confusing that fact that the OP_GET actually implements in, and not get.
- Not a massive fan of the 'in' name to be honest, to me it sounds like something that should check for set membership and return a boolean (I suppose it sort of works like that for tuples and structs, but the default value makes much less sense for that interpretation too).


andrewchambers:
Closing because I have a better idea.

PR #197, opened 2019-12-01
Remove 'in' core function.

andrewchambers:
The 'in' core function was added as a replacement for the new 'get'
semantics, however it is actually an almost entirely redundant and
confusing function.

The reason the function is redundant is because it can nearly be trivially
replaced by just indexing on the datastructure directly.

The reason the function is confusing/not named well is because the name does not
match the underlying bytecode, 'in' was implemented by the 'get' bytecode
and get was implemented by the get c function, a very confusing situation.

This pull request simply removes 'in' because it is not needed.

### DIFF
--- a/examples/debugger.janet
+++ b/examples/debugger.janet
@@ -21,22 +21,22 @@
   "Show a stack frame"
   [&opt n]
   (def stack (debug/stack (.fiber)))
-  (in stack (or n 0)))
+  (stack (or n 0)))
 
 (defn .fn
   "Get the current function"
   [&opt n]
-  (in (.frame n) :function))
+  ((.frame n) :function))
 
 (defn .slots
   "Get an array of slots in a stack frame"
   [&opt n]
-  (in (.frame n) :slots))
+  ((.frame n) :slots))
 
 (defn .slot
   "Get the value of the nth slot."
   [&opt nth frame-idx]
-  (in (.slots frame-idx) (or nth 0)))
+  ((.slots frame-idx) (or nth 0)))
 
 (defn .quit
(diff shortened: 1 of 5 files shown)


andrewchambers:
Note, this is WIP until I understand that XXX comment.

bakpakin:
While semantically redundant, the compiler can generate better bytecode with an explicit use of it. This is because indexing into a data structure directly is actually a function call, while a call to in can be inlined by the compiler with instruction JOP_GET (which is obviously confusingly named, and was named before `in`was added)

bakpakin:
We could also change the name of the opcode.

bakpakin:
Also I don’t think there is a need to change gendoc.janet?

andrewchambers:
The reason for changing gendoc is because a key was passed that is a number. I wonder if that is another bug, CI fails without that change.

My argument against the 'in' function is that the name really doesn't make sense to me at all.
```
(in table :key) # makes total sense
(in buffer 0) # ambiguous and confusing, could be checking if 0 is in buffer.
(in buffer 0 1) # even more confusing 
```
I'm totally ok with it under a different name, even get* makes more sense to me than in.

bakpakin:
The reason it’s not passing is probably somehow your change caused a number to leak into the root environment as a key. Changing gendoc is not the solution here

bakpakin:
We could definitely change the name, perhaps to `index` or something. But also continuous breaking changes for aesthetics is not a fun road. But better now when the feature is young.

andrewchambers:
ah sorry for that foolish breaking change. Let me rework and comeup with something nicer,

PR #198, opened 2019-12-01
Integers convert to plain number strings.

andrewchambers:
A user can use (type n) to find the true type, the old behavior did not
seem useful for most uses of the string function.


bakpakin:
Looks ok to me. I originally wanted to not have people fooled into thinking they had a normal number type at the repl, but it's probably ok for now. I think i need to look into Appveyor (which has been the worst of the CI products I'm using).

andrewchambers:
Yeah, your reasoning is sound, is maybe we can add an L suffix or something the common integer parsers understand.

PR #199, opened 2019-12-01
Rename in to ind

andrewchambers:
Why:

Consider (in t k), it could be considered by a user to be set membership
at first glance, but actually is indexing and lookup. This may be a
source of confusion for many people in the future we can avoid with some
changes now.

Notes:

Unfortunately some code in boot uses the name ind, in the cases it
conflicted the variables were renamed to indexed.

Note this change still leaves an inconsistency where the ind function
is mostly implemented by the OP_GET opcode. It's possible to rename to
OP_IND, but then we should also rename janet_get() too. It also leaves
the question about abstract types having a get method and not an ind
method... Overall those issues seem less of a concern.

### DIFF
--- a/examples/debugger.janet
+++ b/examples/debugger.janet
@@ -21,22 +21,22 @@
   "Show a stack frame"
   [&opt n]
   (def stack (debug/stack (.fiber)))
-  (in stack (or n 0)))
+  (ind stack (or n 0)))
 
 (defn .fn
   "Get the current function"
   [&opt n]
-  (in (.frame n) :function))
+  (ind (.frame n) :function))
 
 (defn .slots
   "Get an array of slots in a stack frame"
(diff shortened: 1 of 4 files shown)


andrewchambers:
I also considered 'lookup' as an alternative name, its longer but I kind of like it.

andrewchambers:
The more I think about this, the less sure I am.

```
(in {:key true} :key) # great works well.
(in {:foo false :bar :ok} k) # sort of confusing when value is false, but maybe just an idiosyncracy.
(in t :key bar) # Somewhat confusing.
(in @"buffer" 10) # Somewhat confusing.
                           # Can be read as 'item 10 in buffer'
```

andrewchambers:
I'm overall just not sure anymore.

PR #201, opened 2019-12-01
Update documentation for in and get builtins.

andrewchambers:
Try to clarify documentation and teach users the correct
way to read the 'in' so it is less likely to be confused
with python's usage of the keyword.


andrewchambers:
@bakpakin Once I rewired by brain a bit, I understand your logic far more, hopefully this conveys your intention in the documentation.

By the way, I wonder if 'in' really needs a default value at all, it can't be used with buffers anyway.

bakpakin:
Looks good to me, merging.

PR #204, opened 2019-12-02
New capi janet_get_permissive

andrewchambers:
The janet_get_permissive function implements the core semantics
of the 'get' function. The original janet_get implements the semantics of
the 'in' function and also the OP_GET opcode. This slight oddity is
to avoid a backwards incompatible change.


bakpakin:
This looks good to me. I think it can be the start of moving the `get` function back into an opcode + some inline Janet assembly rather than a C function. This would then allow the current compiler to more easily inline `get` with two arguments to a single opcode.

PR #208, opened 2019-12-03
Explain the logic behind negative slice indices.

andrewchambers:
This wastes a few bytes, but might save an argument or two in the future if we avoid someone getting burnt by accident.


bakpakin:
Wasting a few bytes on docstrings is totally fine by me (you can configure Janet to use no docstrings if you want to save a few bytes). Thanks!

PR #209, opened 2019-12-03
Add os/cryptorand.

andrewchambers:


bakpakin:
LGTM, merging. To bad getrandom() doesn't work cross distro.

andrewchambers:
Yeah, maybe we can work out the right #ifdefs to detect it in the future.

PR #222, opened 2019-12-06
Add oob protection to unmarshal (slightly breaking for code using marshal).

andrewchambers:
Add OOB checking to unmarshal (slightly breaking).

The new field was added to the end to hopefully
avoid breaking most sensible libs. I think unmarshalling
is rarely used in the package ecosystem so far.


andrewchambers:
It should be noted that until janet can also handle per thread memory allocation limits and recover from JANET_OOM, passing an untrusted buffer to unmarshal is always going to cause issues.

andrewchambers:
Closing because we had a better idea.

PR #226, opened 2019-12-09
Abstract type getters can indicate key absence.

andrewchambers:
This change to the c api allows abstract types to indicate
to the runtime if a key was absent, or if it meant to return nil.


andrewchambers:
I wonder if the method function can be adjusted to work better with the new api.

PR #229, opened 2019-12-28
Resolve #225 - add file/open buffer parameter.

### LINKED ISSUE #225: No way to disable or configure buffering of a FILE
Related to #224 , but it would also be nice to be able to just disable the buffer, or set the buffer size from janet code.

KrzysztofSzewczyk:
TODO:
 - [X] Add the parameter to `file/popen`
 - [X] Add the parameter to `file/fdopen`
 - [X] Update the documentation entries.

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -581,11 +581,14 @@ static const JanetReg io_cfuns[] = {
     },
     {
         "file/open", cfun_io_fopen,
-        JDOC("(file/open path &opt mode)\n\n"
-             "Open a file. path is an absolute or relative path, and "
-             "mode is a set of flags indicating the mode to open the file in. "
-             "mode is a keyword where each character represents a flag. If the file "
-             "cannot be opened, returns nil, otherwise returns the new file handle. "
+        JDOC("(file/open path &opt mode &opt buf)\n\n"
+             "Open a file. path is an absolute or relative path, "
+             "mode is a set of flags indicating the mode to open the file in."
+             "mode is a keyword where each character represents a flag."
+             "buf is the buffer instance to be used for output caching."
+             "if buf's capacity is equal to zero, buffering is disabled."
+             "If the file cannot be opened, returns nil, otherwise returns the new file"
+             "handle. "
              "Mode flags:\n\n"
              "\tr - allow reading from the file\n"
              "\tw - allow writing to the file\n"
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -342,9 +342,15 @@ static Janet cfun_io_fflush(int32_t argc, Janet *argv) {
 static int cfun_io_gc(void *p, size_t len) {
     (void) len;
     IOFile *iof = (IOFile *)p;
+    
     if (!(iof->flags & (JANET_FILE_NOT_CLOSEABLE | JANET_FILE_CLOSED))) {
         return fclose(iof->file);
     }
+    
+    if(NULL != iof->buf) {
+        free(iof->buf);
+    }
+    
     return 0;
 }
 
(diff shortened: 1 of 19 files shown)


KrzysztofSzewczyk:
CC @bakpakin @andrewchambers

bakpakin:
The biggest problem here is that janet buffers are dynamically resizable. If someone were to push to a buffer after using it as a vbuf, it could realloc, and cause a segfault. Once some memory is set as a file's vbuf, it should not be freed or realloced until the file is destroy. This kind of makes it impossible to use buffers directly as the vbuf.  Instead, I think passing in a size to use for the vbuf would make more sense, and allocate ourselves. In the `core/file` finalizer, we would then free that memory.

KrzysztofSzewczyk:
Hm, yeah, I'll get it sorted out soon.

KrzysztofSzewczyk:
I tested it and (it seems) to work.

src/core/io.c
```diff
@@ -127,35 +129,80 @@ static Janet cfun_io_popen(int32_t argc, Janet *argv) {
...
+        if(argc == 3) {
+            int bufsiz = janet_getinteger(argv, 2);
+            
+            if(bufsiz <= 0 && setvbuf(f, NULL, _IONBF, 0))
+                return janet_wrap_nil();
+            
+            buffer = malloc(bufsiz);
+            
+            if(NULL == buffer) {
+                janet_panic("memory exhausted");
```
andrewchambers: Memory exhaustion is should not be handled like this, there is a specific out of memory define you should use.

src/core/io.c
```diff
@@ -127,35 +129,80 @@ static Janet cfun_io_popen(int32_t argc, Janet *argv) {
 #define popen _popen
 #endif
     FILE *f = popen((const char *)fname, (const char *)fmode);
-    if (!f) {
+    
+    if(NULL == f) {
         return janet_wrap_nil();
+    } else {
```
andrewchambers: I think instead of copy pasting the same code 3 times, a static utility could be shared.

src/core/io.c
```diff
@@ -36,6 +36,7 @@ typedef struct IOFile IOFile;
 struct IOFile {
     FILE *file;
     int flags;
+    char * buf;
```
andrewchambers: There is a code formatter script that can be run in the tools directory.

src/core/io.c
andrewchambers: or part of the logic could be in makef

src/core/io.c
KrzysztofSzewczyk: Done.

src/core/io.c
KrzysztofSzewczyk: Done.

src/core/io.c
```diff
@@ -286,6 +316,9 @@ static Janet cfun_io_fclose(int32_t argc, Janet *argv) {
         janet_panic("file is closed");
     if (iof->flags & (JANET_FILE_NOT_CLOSEABLE))
         janet_panic("file not closable");
+    if (NULL != iof->buf) {
```
bakpakin: The buffer should probably be freed _after_ the file is actually closed, since closing a file can cause it to flush, which references this buffer. As is, this causes UB.

andrewchambers:
I still don't necessarily agree with this change, things are going to get worse If I add file/pipe, and file/temp doesn't support this option either.

andrewchambers:
@KrzysztofSzewczyk why did it reformat so many files? I suspect your version of astyle may differ to ours?

src/core/io.c
KrzysztofSzewczyk: Yes, you're right. I missed that

KrzysztofSzewczyk:
My build works as a branch, but for some reason it's broken here.

bakpakin:
Closing for now, as further discussion makes this seem a possibly dubious/unnecessary feature.

PR #230, opened 2019-12-29
Extend file api to allow creating and checking.

andrewchambers:


andrewchambers:
I needed this api for a process library I'm finishing up. The basic idea is the C api can be used to create whatever crazy FILE you want, then you can create the abstract value.

bakpakin:
Looks clean to me.

PR #235, opened 2020-01-02
Fix faults in Make build

theosotr:
Hello

This commit adds some missing Make dependencies so that the targets are regenerated whenever there are updates to the corresponding source files.


bakpakin:
Thanks, I don't edit these files that often.

PR #236, opened 2020-01-02
Add scratch calloc.

andrewchambers:


andrewchambers:
The overflow check comes from how musl calloc does it.

andrewchambers:
Don't understand the failure.

bakpakin:
I think it’s just appveyor being finicky, don’t worry about it.

PR #237, opened 2020-01-03
Compiling Janet to wasm using wasienv to run on wasmer

lsevero:
Greetings, I've successfully compiled janet to wasm using wasienv adding some preprocessor guards to remove unsupported features like setjmps, pipes and termios.

To compile just uncomment the defines `JANET_SINGLE_THREADED`, `JANET_NO_DYNAMIC_MODULES`, `JANET_REDUCED_OS`, and the newly created ones `JANET_NO_SETJMP`, `JANET_NO_PIPES`, `JANET_NO_TERMIOS`.

I can't really how much the integrity of janet is compromised without setjmps because it is used for error handling. Maybe there is a better way of dealing with setjmps on targets that do not support them. Threads are very experimental on wasm, and they were not compiling in janet because of three internal functions (pthread_getspecific, pthread_key_create, pthread_setspecific), hopefully it will be fixed in the future.

Wasmer could definitely improve janet user experience, providing portable binaries, increase the availability using wapm and https://webassembly.sh/

To compile with the above configs I just typed `wasimake make repl` on janet root folder.

Edit:
To run the janet.wasm I was using:
`wasmer run build/janet.wasm --dir=examples/ -- examples/primes.janet`

### DIFF
--- a/src/conf/janetconf.h
+++ b/src/conf/janetconf.h
@@ -35,7 +35,7 @@
 /* #define JANET_BUILD "local" */
 
 /* These settings all affect linking, so use cautiously. */
-/* #define JANET_SINGLE_THREADED  */
+/* #define JANET_SINGLE_THREADED */
 /* #define JANET_NO_DYNAMIC_MODULES */
 /* #define JANET_NO_NANBOX */
 /* #define JANET_API __attribute__((visibility ("default"))) */
--- a/src/conf/janetconf.h
+++ b/src/conf/janetconf.h
@@ -35,7 +35,7 @@
 /* #define JANET_BUILD "local" */
 
 /* These settings all affect linking, so use cautiously. */
-/* #define JANET_SINGLE_THREADED */
+/* #define JANET_SINGLE_THREADED  */
 /* #define JANET_NO_DYNAMIC_MODULES */
 /* #define JANET_NO_NANBOX */
 /* #define JANET_API __attribute__((visibility ("default"))) */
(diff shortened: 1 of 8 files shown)


andrewchambers:
Like you said, janet totally relies on setjmp for error handling, it is totally useless without that feature.

andrewchambers:
@lsevero  I'm not sure just commenting out critical sections of code is a wise strategy. I am against merging this PR.

bakpakin:
It's good work, but as @andrewchambers said, without setjmp/longjmp, Janet is basically useless, much like Lua. There are ways to implement compiling without setjmp in a reasonable way, such as always aborting the whole program on panics, which might make sense in some limited use cases, or being able to hook in a custom exception handling mechanism. However, for full use of Janet, non-local returns are absolutely needed, it's baked into the C API.

I really want to get Janet compiling easily to WASM, but I think I will wait until web assembly has support for non-local returns (setjmp/longjmp) before trying to support it in the repo. See [the proposal](https://github.com/WebAssembly/exception-handling) here.

Lastly, janetconf.h should not do anything with code in `src/mainclient`. Instead, the idea for getting Janet running in new, non-posix environments is  to write a new client main function, and compile it with the amalgamated build, like `clang --target=wasm32 -std=c99 -O2 ... janet.c mymain.c -o webjanet.wasm` (using whatever toolchain you like). This means the JANET_NO_TERMIOS is not needed. I also think we can just turn off popen support if JANET_REDUCED_OS rather than add a new define. The main reason for the REDUCED_OS option is to allow compiling Janet for less feature rich platforms that a posix operating system or Windows.

I can't merge this, as simply getting something to compile doesn't imply that it actually works. I'm certain that running even a simple test with error handling would segfault (or whatever wasmer would do). `examples/primes.janet` works of course because there are no errors!.

PR #247, opened 2020-01-06
Optional default value for os/getenv.

andrewchambers:


bakpakin:
Looks pretty good to me, seems Appveyor is having another stroke. Merging, thanks!

PR #257, opened 2020-01-16
Use siphash for string hashing.

andrewchambers:
The hash key still needs to randomly initialized
for the security advantage, but this patch is a
step closer to avoiding hash based DOS.

Further work may including exposing the raw hash
function for use by abstract types who also choose to
implement hash.

### DIFF
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -94,14 +94,110 @@ const char *const janet_status_names[16] = {
     "alive"
 };
 
+/*
+  Public domain siphash implementation sourced from:
+
+  https://raw.githubusercontent.com/veorq/SipHash/master/halfsiphash.c
+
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks good, it doesn't look like SipHash affects performance of normal programs very much. However, I will add an option to disable at compile time for various reasons. Some seem to think that this (using keyed hash functions in hash tables) is an insecure patch over a bad hash table implementations with O(n) worse case performance for insert, and that the correct solution is to use a data structure with O(log(n)) worst case performance, like a red black tree. However, I am loathe to move away from open hashing in the current implementation, so I think this is a good compromise. If someone could come up with an open hashing solution (one contiguous block of allocated memory) with O(log(n)) worst-case insertion time, that would be amazing. Seems unlikely, though.

PR #266, opened 2020-01-21
Make file/close idempotent.

andrewchambers:
It is easier to use constructs like defer
with complex control flow if it is safe to close
a file twice.


pepe:
I am doing this on closing LevelDB. But I am eager to learn if it is good/bad :-).

andrewchambers:
@pepe So go allows multiple closing. It's especially useful if closing does both cleanup and notifies you of success of a write.

pepe:
Makes sense

bakpakin:
This is totally reasonable, and probably a good pattern to adopt everywhere.

PR #268, opened 2020-01-22
Make REPL key bindings more similar to those on GNU readline.

crocket:
* I deleted Alt-H and Alt-L because Ctrl-F and Ctrl-B serve the same roles.
* Ctrl-W, Alt-D, Alt-F, and Alt-B behave more similarly to the same key bindings on GNU readline.
* Improved documentation of REPL keybindings on man page.
* Home and End keys now work on more terminal environments.
* Removed bindings for `Esc OH` and `Esc OF` because andrewchambers doesn't need those bindings and the bindings don't seem to make much sense for Home and End. `Esc O` is Single Shift Select of G3 Character Set in xterm. https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

### DIFF
--- a/janet.1
+++ b/janet.1
@@ -29,68 +29,88 @@ most new platforms.
 .SH REPL KEY-BINDINGS
 
 .TP 16
-.BR Home/Ctrl\-A
-Move cursor to the beginning of input line
+.BR Home
+Move cursor to the beginning of input line.
 
 .TP 16
 .BR End
-Move cursor to the end of input line
+Move cursor to the end of input line.
 
 .TP 16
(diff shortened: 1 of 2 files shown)


bakpakin:
Just a note, I prefer `for (;;) {` over `while (true) {`, and none of the other modules use stdbool.h.

crocket:
I reflected your preference in the code.

bakpakin:
Ok, I will merge this. Thanks for your work here, it looks pretty good.

crocket:
Thanks for merging fast.

PR #280, opened 2020-02-03
Fix next function arity

### LINKED ISSUE #279: Argument key is not optional for next function
When calling `(next (range 0 10))` I get:

```
compile error: <function next> expects at least 2 argument, got 1 on line 2, column 1 while compiling repl
```

`(next (range 0 10) 0)` works as expected.

pepe:
This fixes #279. But I am not sure, how deep this gets (`quick_asm` suffix is scary for me).


bakpakin:
Looks good to me.

PR #298, opened 2020-03-06
Make alt-backspace behave like ctrl-w.

leahneukirchen:
Another common binding in readline and its clones.

### DIFF
--- a/src/mainclient/shell.c
+++ b/src/mainclient/shell.c
@@ -765,6 +765,9 @@ static int line() {
                         case '.': /* Alt-. */
                             historymove(-JANET_HISTORY_MAX);
                             break;
+                        case 127: /* Alt-backspace */
+                            kbackspacew();
+                            break;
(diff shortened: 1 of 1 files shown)


bakpakin:
Lgtm, thanks! There a whole bunch of keybindings that we can add, but just the most common readline bindings are a good start

leahneukirchen:
Just adding the ones I find myself missing. :)

One thing is that the readline is currently ignoring all non-7-bit chars, which makes Unicode input impossible. I've looked at a few utf8-capable linenoise forks, but they add substantial complexity.

bakpakin:
The hardest part with general utf-8/multibyte input is determining character widths in an acceptable way, as you need to have a sophisticated way of determining the width of glyphs in number of character boxes, which depend on the terminal, font, etc. Not to mention that unicode codepoints != unicode glyphs.

PR #312, opened 2020-03-14
Expose PEG types to C api

andrewchambers:


andrewchambers:
Hmm think this can be done slightly better...

PR #318, opened 2020-03-18
os/date fixes and os/mktime

leahneukirchen:
The first two patches are bugfixes, the third patch adds a new function os/mktime that does the inverse of os/date, i.e. assemble a date from a broken-down date struct.

This is very useful because it also normalizes the date, so you can easily add days to :month-day,
roundtrip, and get the new date, without implementing a whole calendar system.

Unfortunately, `timegm` is not POSIX, but all systems we check for JANET_POSIX except for Solaris have it. Currently we just panic on Solaris, but I could also prepare a MIT-compatible fallback function based on musl libc that does the same thing.

Windows support is untested and just googled.

I'll add unit tests if you think this is a useful addition.

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -621,8 +621,6 @@ static Janet os_date(int32_t argc, Janet *argv) {
     struct tm *t_info = NULL;
     if (argc) {
         int64_t integer = janet_getinteger64(argv, 0);
-        if (integer < 0)
-            janet_panicf("expected non-negative 64 bit signed integer, got %v", argv[0]);
         t = (time_t) integer;
     } else {
         time(&t);
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -69,6 +69,13 @@ extern char **environ;
 void arc4random_buf(void *buf, size_t nbytes);
 #endif
 
+/* Not POSIX, but all Unixes but Solaris have this function. */
+#if defined(JANET_POSIX) && !defined(__sun)
+time_t timegm(struct tm *tm);
+#elif defined(JANET_WINDOWS)
+#define timegm _mkgmtime
+#endif
+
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks good to me. As much as I’d like to support Solaris, I’ve had much trouble in the past emulating a SPARC system, so I’m fine with the omission of SPARC/Solaris. I have not tried Solaris on x86-64, though.

Unless we can actual test on Solaris, ideally in CI, I think it’s a fools errand to try and make sure that it works. I would like to be sure that we could accept a patch to fix it, though.

leahneukirchen:
Well, the code to emulate it is a bunch of portable functions, we can test than on any system.

bakpakin:
In that case, it would be appreciated. I would like to start testing on SPARC someday if I can get things working.

leahneukirchen:
It's essentially this code: https://github.com/leahneukirchen/mblaze/blob/master/mytimegm.c
Where should it go?

leahneukirchen:
Why is the CI failing btw?

bakpakin:
Seems LGTM is not using a new version of Meson or something, I can update the build file. Appveyor sometimes just times out sometimes for reasons unrelated to what we are doing.

bakpakin:
I will merge this anyways, and hopefully LGTM will catch up. I have fixed it on master.

andrewchambers:
Is there an alternative to true/false, if there is a timezone, can't we specify it?

bakpakin:
Since this is mostly posix c stdlib stuff, you would need to set locales.

andrewchambers:
Oh I see, a flags arg is another way to do it, this is also open to some extension.

leahneukirchen:
Even worse, you need to change `$TZ` which is process global.

PR #319, opened 2020-03-18
os/lstat and os/readlink

leahneukirchen:


bakpakin:
Looks good to me, thanks! These are definitely useful for scripting.

PR #328, opened 2020-03-30
Fix build and tests on big and little endian PowerPC systems

q66:
Little endian systems were mistakenly being detected as big endian, which was breaking number serialization/deserialization.

Big endian systems had a typo in the unmarshalling code, which was creating subtly broken results.

This fixes both, so test suite passes on little and big endian ppc64 systems, and most likely other big endian systems.


bakpakin:
Looks good to me, thanks for the fix. I will probably merge this back into the 1.8 branch and make a patch release.

PR #332, opened 2020-04-02
Fix doc typo in os/mkdir.

andrewchambers:


bakpakin:
Fixed  in another commit, closing.

PR #344, opened 2020-04-13
Added a fuzzer and integration with OSS-Fuzz.

DavidKorczynski:
Hi,
I starting fuzzing Janet and have created an initial integration with OSS-Fuzz. The fuzzer is in the pull request and the integration with OSS-Fuzz will occur in their own repository (https://github.com/google/oss-fuzz). The build itself occurs within OSS-Fuzz and the code will be continuously fuzzed. If you are interested then please merge the request and provide me with an email to which bug reports will be sent - notice this email will be publicly visible. I am happy to continue creating fuzzers for the project if you would like but the current fuzzer covers a good amount of the code so this should be fairly useful in finding bugs.

David

### DIFF
--- a/test/fuzzers/fuzz_dostring.c
+++ b/test/fuzzers/fuzz_dostring.c
@@ -1,0 +1,22 @@
+#include <stdint.h>
+#include <string.h>
+#include <janet.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+	char *new_str = (char *)malloc(size+1);
+	if (new_str == NULL){
+		return 0;
+	}
+	memcpy(new_str, data, size);
+	new_str[size] = '\0';
+
+	/* janet logic */
+	janet_init();
+	JanetTable *env = janet_core_env(NULL);
+	janet_dostring(env, new_str, "main", NULL);
+	janet_deinit();
+
+	free(new_str);
+	return 0;
+}
+


bakpakin:
Thanks for your input here! This looks like a good start, although is that really ok to add to OSSFuzz? I mean, it could do things like (os/shell “rm -rf /“), although I would imagine google locks things down fairly well.

I guess you could just look for segfaults and ignore hangs, as an infinite loop is pretty easy to add.

DavidKorczynski:
Thanks for the quick feedback! I am not entirely sure what their response will be, however, if you are happy to integrate with OSS-Fuzz then I will try and merge with them and hear their views. If they are open to integrating the project but not in the current state then I can integrate the fuzzers deeper in the project, e.g. the parser and similar.

andrewchambers:
I think you will get better results fuzzing in a more controlled way.

bakpakin:
I will merge this for now in case it turns up something interesting - I think we can try it and see what happens.

andrewchambers:
Yeah, I think we can unify the fuzzing with my afl stuff, delete the old stuff, then get google to fuzz everything for us. It will be better this way in the long run.

DavidKorczynski:
Cool - I will do some more work on this and also integrate fuzzers deeper into the project.

andrewchambers:
@DavidKorczynski have a look inside tools/afl . I am especially interested in getting google to do fuzzing of marshal/unmarshal.

DavidKorczynski:
Looks good - I will fix this up tomorrow! Will do the parser as well.

PR #345, opened 2020-04-14
Check some *alloc return values

sogaiu:
May be these are unnecessary?


bakpakin:
Nope, they look good. Thanks!

PR #350, opened 2020-04-18
Updated the libfuzzer to target marshalling.

DavidKorczynski:
Cross-referencing: https://github.com/janet-lang/janet/pull/344

Updated the fuzzer to target unmarshalling, which also helps us avoid executing random commands and infinite loops as discussed above. Do you think this will be better and can I in fact hit the unmarshalling code in this way?

### DIFF
--- a/test/fuzzers/fuzz_dostring.c
+++ b/test/fuzzers/fuzz_dostring.c
@@ -2,45 +2,45 @@
 #include <string.h>
 #include <janet.h>
 
-
-char *cmd_start = "(def v (unmarshal (slurp ((dyn :";
-char *cmd_end = ") 1)) load-image-dict));
-
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-    char *new_str = (char *)malloc(size + strlen(cmd_start) + strlen(cmd_end) + 1);
-    if (new_str == NULL) {
-        return 0;
-    }
-    // Copy start of command into the string we will execute
-    memcpy(new_str, cmd_start, strlen(cmd_start));
 
-    // Copy fuzz-data into the exec string
-    char *datap = new_str + strlen(cmd_start);
-    memcpy(datap, data, size);
-
-    // Copy the end of the command into the exec string
-    end_data = datap + size;
-    memcpy(end_data, cmd_end, strlen(cmd_end));
-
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Also, in the previous PR I asked for an email that I can add to  OSS-Fuzz, which will receive emails with bug reports - do you have one I can use? FYI the email will be public on the oss-fuzz repo.

bakpakin:
I have no idea how your arrived at the source code to use for fuzzing, but it is very wrong - you do not need to use slurp or dyn for anything, there is no need for a leading `:` to a string (that is actually an empty keyword followed by a string), and I don't see what the purpose of indexing is - all you are doing here is fuzzing one of 256 bytes. It almost looks like the output of the fuzzer.

Rather than trying to manipulate lisp source code as a string (a hairy idea at the best of times), you can pass your input bytes directly to janet_unmarshal.

As for the email, the public email on my account will work.

DavidKorczynski:
Thanks for the quick feedback and apologies for the very wrong fuzzer. I was simply trying to follow the logic of the `.sh` script you used with AFL - albeit a bit naive. However, I have now modified it to go more for the parser logic, and this seems to be working well on my system. Could you confirm that the code makes sense?

Regarding the unmarshalling code, I tried the following:
```c++
    janet_init();
    JanetTable *reg = NULL;
    janet_unmarshal(new_str, size+1, 0, reg, NULL);
    janet_deinit();
```
which compiles and runs well initially. However, we run into a ton of issues with execution running into `janet_panicf` and exits. This doesn't really work with the fuzzer as it exits almost instantaneously, so I think we should avoid that part of the code for now.

bakpakin:
The parser stuff looks alright, I will merge that. Yeah, the unmarshal stuff would need to be wrapped to catch the panics, which should be fine from the point of view of the fuzzer.

DavidKorczynski:
Thanks! 
Cool, for FYI the reason we need to avoid fuzzing code that calls exits etc. is that some of the fuzzers used by OSS-Fuzz (LibFuzzer in particular) are in-memory fuzzers. So when the target calls exit the fuzzer exits too, and this will be considered a "crash". As such we can't fuzz it in some of the same ways that is possible with AFL.

andrewchambers:
@DavidKorczynski You can use pcall to catch janet errors from C.

PR #370, opened 2020-05-05
Fix (mostly nonsensible) race condition in multi threaded processes 

andrewchambers:
Fix (mostly nonsensible) race condition in multi threaded processes using os/execute with os/setenv.


andrewchambers:
This race is pretty dumb, you have to be writing a pretty buggy program to encounter it. Its largely unclear to me if its possible to crash the process by doing this.

bakpakin:
Good catch, that was an oversight.

Issue #374, opened 2020-05-05
Leaking of file descriptors to child processes


andrewchambers:
Currently any use of the thread module can easily leak file descriptors on POSIX systems:

Suppose you have thread one:

```
(while true
  (os/execute "do-work"))
```

And another thread:

```
(while true
  (def passf (os/temp))
  (get-passwd f)
  (def conn (connect-to-database t))
  (slurp "secret-data"))
```

The problem is os/execute calls posix_spawn, while file/temp, fopen and other calls do not open file descriptors with the CLOEXEC flag set, instead setting after a delay it later or not at all..

This combination means secret file descriptors or pipes can be leaked to child processes.

andrewchambers:
Arguably the developer should be aware of this. I mitigate this in my process library by doing a well known hack, attempting to close all open file descriptors after fork.

andrewchambers:
I spoke with some people and apparently fopen might be getting an :e flag to cover this case in a backwards compatible way.

leafgarland:
Commit dc3e9fb breaks the build on mac as `SOCK_CLOEXEC` is unidentified. I guess it could be done using `fcntl` same as #375, which does build on mac.

Also, I think that the `SOCK_CLOEXEC` flag should be used with the `type` argument, not the `protocol` argument to `socket()`?

bakpakin:
@leafgarland Thanks for the fix, but I just moved the fcntl stuff into make_stream.

PR #376, opened 2020-05-07
Replace usage of SOCK_CLOEXEC with call to fcntl

leafgarland:
As mentioned in #374. 

The SOCK_CLOEXEC flag is not available on macs, so use fcntl to set the
same option instead. Since we have to conditionally compile this for
Windows anyway (no fcntl) I also removed JSOCKFLAGS.


leafgarland:
Ahh, fixed already. I'll close this.

PR #393, opened 2020-05-11
Fix with-vars doesnt allow fresh variables.

honix:
I found this behavior of ```with-vars``` wrong:
``` lisp
janet:2:> (with-vars [abc 456] (- abc 100))
compile error: unknown symbol abc on line 2, column 1 while compiling repl
```
It only works with names that already declared as variables, but I expect this macro to declare fresh variables too as mentioned in the docs this will work in same logic as ```let``` macro.

I simplify ```with-vars``` macro so it works as I expect it to and added more tests on ```with-vars``` with both fresh variables and variables shadowing.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1084,16 +1084,11 @@
   [vars & body]
   (def len (length vars))
   (unless (even? len) (error "expected even number of argument to vars"))
-  (def temp (seq [i :range [0 len 2]] (gensym)))
-  (def saveold (seq [i :range [0 len 2]] ['def (temp (/ i 2)) (vars i)]))
-  (def setnew (seq [i :range [0 len 2]] ['set (vars i) (vars (+ i 1))]))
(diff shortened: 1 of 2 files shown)


bakpakin:
This really isn’t the point of with-vars, your change actually has different semantics - the new vars don’t update the existing vars. The main idea is to be like with-dyns, such that functions that reference the original vars will be updated as well.

Example:

```
(var a 10)
(def b [] a)
(with-vars [a 20] (b)) # should be 20
```

This change breaks this behavior.  Basically, with-vars is not mutable let, it is for temporary assignment of vars.

honix:
Oh. Now I see, non-trivial macro for me. Thanks for example.

If ```with-vars``` does only variable substitution, I think there will be some version of mutable let.

honix:
I think little examples like this will improve docs quality, espesialy in non trivial ones, as this gives clue what kind of task it meet.

PR #395, opened 2020-05-13
Add snapcraft packaging

zyga:
Snapcraft allows the latest version of Janet to be immediately available
to developers on most popular Linux distributions. This patch provides
the snapcraft.yaml file, which provides information about the package,
how to build it and how to install it.

Both janet and jpm are exposed, though the latter requires a store side
alias to be provided, as snap names also control application names, to
avoid clashes between applications.

For this prototype I've selected classic confinement. Application
language packages are usually provided with classic confinement, as that
is most flexible for developers.

Snaps do not support providing manual pages yet, so no entries for those
were created.

Signed-off-by: Zygmunt Krynicki <me@zygoon.pl>

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,27 @@
+name: janet
+adopt-info: janet
+summary: functional and imperative programming language and bytecode interpreter
+description: |
+    **Janet** is a functional and imperative programming language and bytecode interpreter. It is a
+    modern lisp, but lists are replaced
+    by other data structures (arrays, tables (hash table), struct (immutable hash table), tuples).
+    The language also supports bridging to native code written in C, meta-programming with macros, and bytecode assembly.
+base: core18
+confinement: classic
+grade: stable
+apps:
+    janet:
+        command: usr/bin/janet
+        adapter: none
+    jpm:
+        command: usr/bin/jpm
+        adapter: none
+parts:
+    janet:
+        plugin: make
+        source: .
+        make-parameters: [PREFIX=/usr]
+        build-packages: [build-essential]
+        override-build: |
+            snapcraftctl build
+            snapcraftctl set-version "$($SNAPCRAFT_PART_INSTALL/usr/bin/janet -e '(print "Version: " janet/version)' | awk '{ print $2 }')"


zyga:
Hey there. I found Janet via twitter and found it interesting. As a snapd developer I quickly made a snap package so that I can deploy janet easily to my machines.

To use this packaging you need to register a snap name. You can also obtain an automatic alias so that `jpm` is available after installing `janet`. Using build.snapcraft.io you can estabilish automatic build and publishing pipeline, so that all commits on master get automatically built and published in the snap store, into the `edge` channel of `janet` snap.

I've never contributed snap packaging before so please forgive me for assuming you are familiar with something without properly explaining it. I'm happy to help if you have questions.

bakpakin:
Thanks for adding this - this looks pretty good to me, as long as we don't need to maintain anything besides this file and the pipeline. I am aware of what snap is but not very familiar. A few questions though:

- I don't see any commands like `make` or `make install` in the package, just a make plugin. Will this properly use the Makefile to install things? And can we customize this?
- Can I set up the pipeline to only push and publish tagged releases? I don't want to simply publish the master branch, unless that is the default expected by snapd users.

andrewchambers:
Please don't clutter the main directory if not necessary.

bakpakin:
I personally don't mind cluttering the main directory - also, it seems like snapcraft.io looks  specifically for a snapcraft.yaml file for it's automation. That said, if possible the tools directory is a good place for miscellaneous files.

zyga:
> Thanks for adding this - this looks pretty good to me, as long as we don't need to maintain anything besides this file and the pipeline. I am aware of what snap is but not very familiar. A few questions though:
> 
>     * I don't see any commands like `make` or `make install` in the package, just a make plugin. Will this properly use the Makefile to install things? And can we customize this?

The make plugin is a part of snapcraft and it integrates with typical makefile workflows. You can freely customize it, down to overriding each part but this does "make && make install" with the required options to build the package.
 
>     * Can I set up the pipeline to only push and publish tagged releases? I don't want to simply publish the master branch, unless that is the default expected by snapd users.

Yes you can but the intent is different. By default you can build everything commited and publish it into the *edge* of the *latest* channel. Every snap gets four stability levels by default, *stable*, *candidate*, *beta* and *edge*. It's the expectation that edge is really the CI pipeline.

Developers can either directly publish *existing* builds from edge to one of the other channels or establish separate operations (e.g. only build release branch, only build tags, etc) to build and publis to the stable channel. That part is entirely up to you. I only see snaps that have massive install base so we do many things that are not usually encountered, like complex workflows that move a build from beta, to candidate to the stable channel. In addition we also release a snap to the stable channel in small increments, to lower the load and to get any error reports after a small chunk of the population updates their machines.

In addition, you can have many separate tracks. You can replicate the same edge, beta, candidate and stable in the 1.x/ track as well as a 2.x track (you can request those in the via the store) and publish separately and allow your users to choose which version they get on their systems.

Snap packaging is both easier (packages are genuinely easier to make) and also more complex as there are new concepts (channels, tracks, stability level, hotfix branches and many many more).

zyga:
> I personally don't mind cluttering the main directory - also, it seems like snapcraft.io looks specifically for a snapcraft.yaml file for it's automation. That said, if possible the tools directory is a good place for miscellaneous files.

There are two possible locations and this is more of a compromise between github limitation on APIs and practicality. It's either the top-level directory or the "snap" or "snapcraft" directory (I actually forgot which, sorry). This is well documented on snapcraft.io/docs

You can choose one of those but only those, I'm afraid.

zyga:
As for building, you can either use build.snapcraft.io and get automatic builds for all the supported architectures or build by hand or anything in between. Building is really just "snap install snapcraft && snapcraft". My main point was building is distinct from publishing and when you publish, there are many locations you can publish to (channels/tracks).

zyga:
@bakpakin please let me know if you need help registering the snap name https://snapcraft.io/docs/registering-your-app-name and publishing your builds. I can also help you establish an alias, so that `janet.jpm` is available as just `jpm`.

PR #396, opened 2020-05-13
update janetconf.h path in README embedding section

halfhorst:
`janetconf.h` doesn't exist in `src/include` anymore, it looks like it's in `conf`.


bakpakin:
Thanks, this does need to be updated.

PR #399, opened 2020-05-15
Add make help target

LouisJackman:
Add and document a make target for providing help on the available targets.

This might not be so useful for experienced make users, but could be useful for new contributors who aren't so comfortable with make. I first saw this idea in [Pelican](https://github.com/getpelican/pelican), and have since borrowed it for other projects: https://github.com/getpelican/pelican/blob/bf85991ee648665f2abc224b9d3719c27470e696/pelican/tools/templates/Makefile.jinja2#L61


bakpakin:
Seems useful, thanks!

PR #401, opened 2020-05-16
Add a dbg macro for easy var dumping

LouisJackman:
I've been playing with this language and, as you'd expect, making mistakes. This `dbg` macro has helped to debug them. It dumps values like `pp`, but allows multiple values at once and also displays the variable name.

Sometimes dumping a variable solves a bug more quickly than opening up a full-blown debugger.

```
$ make repl
Janet 1.9.1-d6cd69e  Copyright (C) 2017-2020 Calvin Rose
janet:1:> (def x 42)
42
janet:2:> (def y @[1 :abc 'foo-bar])
(1 :abc foo-bar)
janet:3:> (dbg x "Hello, world!" y)
`x` has value 42
Literal has value "Hello, world!"
`y` has value @[1 :abc foo-bar]
(42 "Hello, world!" @[1 :abc foo-bar])
janet:4:> 
```

If the name is too similar to the existing `debug` function, it could also be named `dump` or `dump-values`.

It is inspired by Rust: https://github.com/rust-lang/rfcs/blob/master/text/2361-dbg-macro.md

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -945,7 +945,7 @@
     (array/push parts (tuple apply f $args)))
   (tuple 'fn (tuple '& $args) (tuple/slice parts 0)))
 
-(defmacro trace-pp
+(defmacro tracev
   "Displays the variables or literals listed, providing both the name and
   and the value for variables. Designed for quick debugging of values. Returns
   the traced forms: nil for none, the form itself for one, and a tuple of the
@@ -955,13 +955,12 @@
     ~(do
       (def ,results @[])
       ,;(map (fn [form]
-               (def preface (if (symbol? form)
-                              (string form " is")
-                              "Expression is"))
                ~(do
                  (def ,var ,form)
-                 (eprintf (string "%s " (dyn :pretty-format "%q"))
-                          ,preface
+                 (eprintf (string (dyn :pretty-format "%q")
+                                  " is "
+                                  (dyn :pretty-format "%q"))
+                          ',form
                           ,var)
                  (eflush)
                  (array/push ,results ,var)))
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -945,6 +945,22 @@
     (array/push parts (tuple apply f $args)))
   (tuple 'fn (tuple '& $args) (tuple/slice parts 0)))
 
+(defmacro dbg
+  "Displays the variables or literals listed, providing both the name and
+  and the value for variables. Designed for quick debugging of values and
+  nothing else."
+  [& forms]
+  ~(do
+    ,;(map (fn [arg]
+            (def preface (if (symbol? arg)
+                           (string "`" arg "` has value ")
+                           "Literal has value "))
+            ~(do
+              (prin ,preface)
+              (pp ,arg)
+              nil))
+          forms)))
+
 (defmacro ->
   "Threading macro. Inserts x as the second value in the first form
   in forms, and inserts the modified first form into the second form
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -957,7 +957,7 @@
       ,;(map (fn [form]
                (def preface (if (symbol? form)
                               (string form " is")
-                              "is"))
+                              "Literal is"))
                ~(do
                  (def ,var ,form)
                  (eprintf (string "%s " (dyn :pretty-format "%q"))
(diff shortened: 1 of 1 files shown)


LouisJackman:
Updated it to return the expressions to, to allow dropping a `dbg` call in the middle of a larger, more complex expression.

bakpakin:
I wouldn't mind merging this, although a couple of things:

- I don't like the `dbg` name because it is too close to `debug`, which does a very different thing. Maybe something like `(trace-pp ...)` would be better, as the functionality is a bit closer to the `trace` and `pp` functions.
- The printing format is not consistent with other janet debug messages - I would omit the backticks wrapping a binding name.
- `pp` prints to stdout - this is probably better printed to stderr, as is this case with most debug printing. `pp` is short form for `(printf (dyn :pretty-format "%q") x)`, and you can replace `printf` with `eprintf`. Probably add `(eflush)` at the end o be safe and flush the error stream.

LouisJackman:
Thanks for the review. The three suggestions have been incorporated. The return values have also been elaborate on within the docstring.

andrewchambers:
My 2 cents: 

Its unfortunate the 'trace' name is taken. I feel like a better naming scheme would have been:

```
(trace-calls fn) # enable tracing on function.
(trace v) # Print a debug value trace.
```

w.r.t the PR.

I also think the 'Expression is' stuff could be changed to something like:

```
(trace-pp (+ a b))
trace: (+ a b) is 3
(trace-pp a)
trace: a is 1
(trace-pp b)
trace: b is 1
```

andrewchambers:
Heres a counter proposal that is backwards compatible:
```
(defmacro tracev
  [f]
  (def s (gensym))
  ~(let [,s ,f]
     (,eprintf "trace: %p is %p" (quote ,f) ,s) ,s))
```

```
janet:10:> (def a 123)
123
janet:11:> (tracev (inc a))
trace: (inc a) is 124
124
```

LouisJackman:
Putting the value at the end of the `let` certainly looks cleaner than manually pushing onto a buffer, although being able to throw multiple values into it is quite handy too, which is what the original buffer pushing is handling.

Is `%p` pretty-printing similarly to the `(dyn :pretty-format "%q")` in my version, or is that functionality not desired here?

Your counter proposal fundamentally does what mine does though, so it's fine for me.

LouisJackman:
Alternatively, my one could be updated to be called `tracev` and display the whole form and not just symbols with `Expression is` for everything else. The format could be updated to `trace: `.

That'd maintain the support for multiple values being passed in and returned while adopting those suggestions.

Of course, your version is simpler, and there's always something to be said for simplicity.

andrewchambers:
what is the benefit of multiple values though? Why not a tuple for that case?

LouisJackman:
@andrewchambers, I've adapted your suggestions into mine. The output looks a lot cleaner now, thanks!

```
Janet 1.9.1-9109e36  Copyright (C) 2017-2020 Calvin Rose
janet:1:> (def x 42)
42
janet:2:> (tracev 42 x [1 2 3])
42 is 42
x is 42
[1 2 3] is (1 2 3)
(42 42 (1 2 3))
janet:3:> (tracev 42)
42 is 42
42
janet:4:> (tracev)
nil
```

LouisJackman:
Just added the "trace " prefix to the message that was requested too.

andrewchambers:
I still don't understand why you need a vector. It seems less efficient and more complicated to me.

bakpakin:
> My 2 cents:
> 
> I think the 'Expression is' stuff could be changed to something like:
> 
> ```
> > (trace-pp (+ a b))
> trace: (+ a b) is 3
> > (trace-pp a)
> trace: a is 1
> > (trace-pp b)
> trace: b is 1
> ```
> 
> @bakpakin I feel like trace might be so infrequently used it is worth repurposing the name.
> 
> Its unfortunate the 'trace' name is taken. I feel like a better naming scheme would have been:
> 
> ```
> (trace-calls fn) # enable tracing on function.
> (trace v) # Print a debug value trace.
> ```
> 
> Anyway, I feel like we can think of a nicer name than trace-pp.

I would consider this - I don't think there is any library (much?) library code that would use trace, as it is intended for debugging, so I don't mind breaking it for backwards compatibility. In that vein, we would also probably want to rename untrace as well.

I also tend to agree with @andrewchambers on having the `tracev` macro take multiple values - I think having it be single arity is simpler and has fewer corner cases, and I think this form needs to . The line number thing is actually a good idea as well, although support would be needed from the macro system, probably communicated to the macro via dynamic bindings.

Regardless I will merge this for now (despite there not being consensus) as I think some variant of this macro should go in the next release.

PR #402, opened 2020-05-18
Format lock files for nicer diffs.

andrewchambers:


bakpakin:
Thanks, looks fine to me.

Issue #397, opened 2020-05-13
Appveyor failures


andrewchambers:
Is there a reason they keep failing on every PR? Should we just disable them?

bakpakin:
The problem is the installer randomly times out when we try to run it and test that installation works. It is annoying but is not something I have been able to figure out. It also doesn't seem to repro on my windows machine. I've written it down to general appveyor/Windows/NSIS flakiness (maybe antivirus)?, but it happens far too often to be that. It does not seem to reproduce on my machine.

Some help on this would be nice - maybe an appveyor/NSIS/Windows expert could help with this. I am loathe to simply disable CI, though, as it is important that installation works correctly.

bakpakin:
Also, rerunning the jobs will eventually get them to pass. I have been scratching my head over this for a while, so if any one can tell me why this is happening and fix it I will get you a beer if we ever meet.

leafgarland:
I saw this issue, and wanted to explore github actions, so I added a workflow to replace appveyor. Unfortunately the installer hangs there too 😞.

leafgarland:
I stripped modern UI "MUI2" from the installer and no repeats of the hanging yet. But it is now a basic installer - instant install with progress bar basically.

bakpakin:
I think it started happening when I upgraded to NSIS 3.05 from 3.04 - so it may be a bug there, although I have a feeling that some dialog box is popping up, and since the install is "silent", it is being hidden. Just a theory

leafgarland:
Might seem a bit extreme but as I haven't yet tracked down what in the NSIS installer is causing the hang, I tried building a msi (via a wix script) instead.

So far, it is has not hung.

The installer is currently pretty basic, it only does per-user installs and has no fancy gui, just the basic msi progress bar, but the installed result is the same.

If you want to have a look you can get the installer from the appveyor artifacts [here](https://ci.appveyor.com/project/leafgarland/janet/build/job/oahnx0eu0t26wu1i/artifacts), and see the [diff](https://github.com/leafgarland/janet/commit/f71b2c706f8dedf440502a75229b4b6ec522c100).

I could improve the wix script to match what the NSIS installer (although wix/msi is a very strange world to operate in).

bakpakin:
Ok, this is great :)

I much prefer the Wix installer as NSIS has good docs but outdated design, and it is possible to really mess up a user’s system if not careful.

[...]

PR #406, opened 2020-05-19
Add Wix/msi installer for Windows

leafgarland:
As discussed in #397, here is a new installer for Windows using Wix. So far it has not shown the hanging issue we have seen with the NSIS one when doing silent installs.

It builds x86/x64 specific installers that can be installed per-user or per-machine. The default is per-user. If you select per-machine then you can also customize the install location.

image
image
image
I would prefer to get rid of the above feature selection step but it requires a lot of customisations in Wix and getting this far was way more trouble than I wanted.
image

The custom images in the dialogs were made with [ronin](https://100r.co/site/ronin.html), if anybody wants to make better ones

```
(resize 493 58)

(clear)

(fill frame 
  (gradient 
    (line 0 0 493 0) 
    ("#09a5b8" "#076591")))

(import "icon_svg.svg" 
  (rect 422 2 70 56))
```

```
(clear)

(resize 493 312)

(fill 
  (rect 0 0 164 312) 
  (gradient 
    (line 0 0 0 493) 
    ("#09a5b8" "#076591")))

(import "janet-w200.png" 
  (rect 10 10 144 144))
```


pepe:
Wow, that seems like a lot of troubles endured :-). Good work.

sogaiu:
Looks nice!

bakpakin:
Wow this is so much better! I had heard Wix was much harder to use than NSIS but the end result looks pretty nice and the source code actually looks not too bad at all. Thanks for taking this on, I wasted a lot of time fighting with NSIS and it’s strange scripting language.

leafgarland:
Wix is not so bad but there is a lot of outdated and missing documentation/examples. I spent a lot of time pairing examples down to find out what was actually needed. I'm pretty pleased to have got it relatively succinct.

andrewchambers:
Thank you, we really appreciate this.

Issue #408, opened 2020-05-19
Implement "->next" for tarray


MikeBeller:
Typed arrays do not support ->next in their JanetAbstractType.
This leads to surprising results (for a new user especially).

```janet
janet:1:> (def a (tarray/new :int32 3))
<ta/view 0x000000D08A10>
janet:2:> (map print a)  #works fine
0
0
0
@[nil nil nil]
janet:3:> (count |(= 0 $) a)  # gives wrong answer
0
janet:4:> (count identity (map |(= 0 $) a))  # gives correct answer
3
```

If I'm not mistaken -- implementing "->next" for tarray in typedarray.c would fix this.  I could submit a PR for it if it's deemed a good thing to have.

bakpakin:
Yes, that would be a good thing to have. This was an omission and should be added, the PR would be welcome.

MikeBeller:
Hey while I was working on this I noticed that tarray/new did not fully check the type of it's fifth argument (i.e. the array/buffer argument).  It checked first whether it was a tarray_view but then if not, it assumed it was a tarray_buffer.  This led to the possibility of doing the following, with result being a segfault:

```janet
    (def ta (tarray/new :int32 10 1 0 (int/u64 0)))
    (put ta 0 99)
```
So I fixed that also in my pull request.  I put in tests for that issue too, but they rely on the presence of the int module.  It looks from the source code like the int module may be optional.  If that is the case perhaps someone can suggest another test which gets at the same problem but doesn't use an optional module.

PR #410, opened 2020-05-20
Fix issue #408 -- make "next" work for typed arrays, and also fix

MikeBeller:
bug where tarray/new failed to fully check the type of it's last
argument.


bakpakin:
Thanks, this looks good to me 👍

PR #414, opened 2020-05-22
adds WiX details to README

roobie:
I added a couple of instructions on how to build the .msi with given the (very appreciated) changes in #406


leafgarland:
The `build_win install` command will install the msi, you only need `build_win dist` to make the msi (assuming you have already built all the files with `build_win build`).

roobie:
:+1 Good catch, I'll rephrase it.

PR #415, opened 2020-05-24
Fix numarray example to work with jpm and latest Janet c-api

leafgarland:


bakpakin:
Thanks, looks good to me!

PR #418, opened 2020-05-28
Fix for double free on fclose due to GC not knowing it failed

ahungry:


bakpakin:
Looks good to me, good catch on the bug and the fix.

PR #420, opened 2020-06-02
Implement os/realpath with _fullpath

leafgarland:
Maybe you had a good reason to not do this. I can see that `_fullpath` is not quite the same as `realpath`, because `realpath` only accepts paths that exist but it still seems like it could be useful.


bakpakin:
This is probably fine for now. While technically Windows 10 supports symbolic links, they are not common in the windows world I think this is close enough in functionality for now. The platforms are different enough that I think it's ok that the window's function works a bit differently.

PR #429, opened 2020-06-25
Tune reverse[d] docstrings

pepe:


bakpakin:
The new reverse will only work on arrays (and buffers) as it needs to be able to mutate the array, not tuples, so that should be included in the docstring.

pepe:
I will add it. Thx

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #431, opened 2020-06-26
Add ppc to os/arch

cellularmitosis:
I noticed that my PowerPC build of janet returned "unknown" as the os/arch.

This PR detects PowerPC via `defined(__ppc__)`, which I verified via the following C program:

```c
int main() {
#if defined(__ppc__)
return 1;
#else
return 2;
#endif
}
```

(compiled with the stock gcc 4.0.1 which ships with OS X Tiger / Xcode 2.5)


bakpakin:
LGTM

PR #432, opened 2020-06-26
Support for systems missing O_CLOEXEC

cellularmitosis:
This PR adds a `JANET_NO_CLOEXEC` define to janetconf.h for systems which don't have `O_CLOEXEC`.

This PR is part of an effort to support janet on OS X Tiger / PowerPC.


bakpakin:
General idea is right, but in this case can’t we just check if O_CLOEXEC is defined? Seems more automatic and straightforward, and no need for a configuration option.

cellularmitosis:
Ah good point!  I will resubmit.

Issue #430, opened 2020-06-26
I got Janet working on OS X Tiger / PowerPC!


cellularmitosis:
There were just a few missing pieces on Tiger:
- `posix_spawn()`
- `clock_gettime()`
- `arc4random_buf()`
- `O_CLOEXEC`

Details: https://gist.github.com/cellularmitosis/a54e8bf4d80fb8e3599551af6416e66a

I feel like Janet's ultra-low start-up latency and good performance make an attractive option for PowerPC Macs.

![Screen Shot 2020-06-25 at 10 23 09 PM](https://user-images.githubusercontent.com/223396/85817556-e7e54980-b733-11ea-8bd8-33ad65a9de1f.png)

Would a PR be welcomed?  Or is Tiger considered too old to officially support?

cellularmitosis:
`tiger.patch`:
```diff
diff -urN janet-1.10.1/Makefile janet-1.10.1.tiger/Makefile
--- janet-1.10.1/Makefile	2020-06-18 19:24:17.000000000 -0500
+++ janet-1.10.1.tiger/Makefile	2020-06-25 21:37:26.000000000 -0500
@@ -259,17 +259,17 @@
 	cp -rf $(JANET_HEADERS) '$(DESTDIR)$(INCLUDEDIR)/janet'
 	mkdir -p '$(DESTDIR)$(JANET_PATH)'
 	mkdir -p '$(DESTDIR)$(LIBDIR)'
-	cp $(JANET_LIBRARY) '$(DESTDIR)$(LIBDIR)/libjanet.so.$(shell $(JANET_TARGET) -e '(print janet/version)')'
-	cp $(JANET_STATIC_LIBRARY) '$(DESTDIR)$(LIBDIR)/libjanet.a'
-	ln -sf $(SONAME) '$(DESTDIR)$(LIBDIR)/libjanet.so'
-	ln -sf libjanet.so.$(shell $(JANET_TARGET) -e '(print janet/version)') $(DESTDIR)$(LIBDIR)/$(SONAME)
+	#cp $(JANET_LIBRARY) '$(DESTDIR)$(LIBDIR)/libjanet.so.$(shell $(JANET_TARGET) -e '(print janet/version)')'
+	#cp $(JANET_STATIC_LIBRARY) '$(DESTDIR)$(LIBDIR)/libjanet.a'
+	#ln -sf $(SONAME) '$(DESTDIR)$(LIBDIR)/libjanet.so'
[...]

PR #436, opened 2020-06-30
Add support for systems which are missing arc4random_buf

cellularmitosis:
~~This PR adds a trivial implementation of `arc4random_buf()` (based on https://stackoverflow.com/a/12956868/558735) for systems which have `arc4random()` but not `arc4random_buf()`.  This includes PowerPC Macs which are stuck on OS X Tiger.~~

~~This is enabled by uncommenting the `JANET_NO_ARC4RANDOM_BUF` define in `janetconf.h`.~~

This PR modifies `os_cryptorand()` to use `/dev/urandom` on Macs prior to OS X 10.7.

context: https://github.com/janet-lang/janet/issues/430

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -70,6 +70,20 @@ extern char **environ;
 void arc4random_buf(void *buf, size_t nbytes);
 #endif
 
+/* arc4random_buf wasn't available in OS X until 10.7. */
+#ifdef JANET_NO_ARC4RANDOM_BUF
+/* Based on https://stackoverflow.com/a/12956868/558735 */
+uint32_t arc4random(void);
+void arc4random_buf(void *buf, size_t nbytes) {
+    size_t entropy_len = (nbytes/4)+1;
+    uint32_t entropy[entropy_len];        
+    for (size_t i = 0; i < entropy_len; i++) {
+        entropy[i] = arc4random();
+    }
+    memcpy(buf, entropy, nbytes);
+}
+#endif
+
 /* Not POSIX, but all Unixes but Solaris have this function. */
 #if defined(JANET_POSIX) && !defined(__sun)
 time_t timegm(struct tm *tm);
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -72,15 +72,20 @@ void arc4random_buf(void *buf, size_t nbytes);
 
 /* arc4random_buf wasn't available in OS X until 10.7. */
 #ifdef JANET_NO_ARC4RANDOM_BUF
-/* Based on https://stackoverflow.com/a/12956868/558735 */
 uint32_t arc4random(void);
 void arc4random_buf(void *buf, size_t nbytes) {
-    size_t entropy_len = (nbytes/4)+1;
-    uint32_t entropy[entropy_len];        
-    for (size_t i = 0; i < entropy_len; i++) {
-        entropy[i] = arc4random();
+    uint32_t *buf_as_words = (uint32_t*)buf;
+    size_t nwords = nbytes / 4;
+    for (size_t i=0; i < nwords; i++) {
+        buf_as_words[i] = arc4random();
     }
-    memcpy(buf, entropy, nbytes);
+
+    size_t tail_len = nbytes % 4;
+    if (tail_len) {
+        uint8_t *tail = buf + nbytes - tail_len;
+        uint32_t rand = arc4random();
+        memcpy(tail, &rand, tail_len);
+    }
 }
 #endif
 
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -39,6 +39,10 @@
 
 #define RETRY_EINTR(RC, CALL) do { (RC) = CALL; } while((RC) < 0 && errno == EINTR)
 
+#ifdef JANET_APPLE
+#include <AvailabilityMacros.h>
+#endif
+
 #ifdef JANET_WINDOWS
 #include <windows.h>
 #include <direct.h>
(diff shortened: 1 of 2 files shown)


cellularmitosis:
oops, I ended up with a conflict after migrating this code.

bakpakin:
This implementation can overflow the stack if size is too large. As a matter fo style and correctness, we don't allocate variable sized arrays on the stack in Janet. For example, `(os/cryptorand 0x7FFFFF)` will probably cause a segfault. Instead, assign directly into the output buffer.

bakpakin:
Also, `(nbytes+3)/4` is a better expression for the number of required arc4random() calls.

cellularmitosis:
Great insights, thanks for taking a look!  I will revise.

cellularmitosis:
How about something like this?

```c
void arc4random_buf(void *buf, size_t nbytes) {
    uint32_t *buf_as_words = (uint32_t*)buf;
    size_t nwords = nbytes / 4;
    for (size_t i=0; i < nwords; i++) {
        buf_as_words[i] = arc4random();
    }

    size_t tail_len = nbytes % 4;
    if (tail_len) {
        uint8_t *tail = buf + nbytes - tail_len;
        uint32_t rand = arc4random();
        memcpy(tail, &rand, tail_len);
    }
}
```

andrewchambers:
I don't know if this is a way to name things, obviously windows has no arc4random_buf but this implementation is not valid for windows.

andrewchambers:
Is there no way to simply check for the maxos version? According to your comment you just need to check if the version is greater than 10.7

src/core/os.c
```diff
@@ -70,6 +70,20 @@ extern char **environ;
 void arc4random_buf(void *buf, size_t nbytes);
 #endif
 
+/* arc4random_buf wasn't available in OS X until 10.7. */
+#ifdef JANET_NO_ARC4RANDOM_BUF
+/* Based on https://stackoverflow.com/a/12956868/558735 */
+uint32_t arc4random(void);
+void arc4random_buf(void *buf, size_t nbytes) {
```
andrewchambers, line 77: I guess you want this to be static.

cellularmitosis:
@andrewchambers  Thanks for taking a look!

> I don't know if this is a good way to name things

I think I misunderstand.  If my system were missing `memcpy()`, and I wanted to provide a fill, wouldn't I be forced to name it `memcpy()`?

> Is there no way to simply check for the macos version

From the command line, yes.  We could check the version of the Darwin kernel via `uname -r`.  Would we add a line to the `Makefile` which would call `uname` and then automatically append `#define JANET_NO_ARC4RANDOM_BUF` to `janetconf.h`?

Or do you mean import some header file from which we can determine the kernel version?

cellularmitosis:
@andrewchambers taking another look at `os_cryptorand()`, the Linux implementation uses `/dev/urandom`, and I just verified that device is also available on Tiger:

```
cell@emac$ uname -a
Darwin emac.home 8.11.0 Darwin Kernel Version 8.11.0: Wed Oct 10 18:26:00 PDT 2007; root:xnu-792.24.17~1/RELEASE_PPC Power Macintosh powerpc

cell@emac$ ls -l /dev/urandom 
crw-rw-rw-   1 root  wheel    8,   1 Dec 31  1969 /dev/urandom
```

So that's another route we could take.

Perhaps `#elif defined(JANET_LINUX)` would become something like `#elif defined(JANET_LINUX) && <thing which checks for OS X < 10.7>`

cellularmitosis:
Ok, it looks like we could use `/usr/include/AvailabilityMacros.h`, which defines a range of OS X versions.

On Tiger, I see:

```
#define MAC_OS_X_VERSION_10_0 1000
#define MAC_OS_X_VERSION_10_1 1010
#define MAC_OS_X_VERSION_10_2 1020
#define MAC_OS_X_VERSION_10_3 1030
#define MAC_OS_X_VERSION_10_4 1040
```

Notably, these definitions stop at `10_4`.

I just checked my Mojave box and verified that the definitions stop with `10_14_6`:

```
#define MAC_OS_X_VERSION_10_14      101400
#define MAC_OS_X_VERSION_10_14_1    101401
#define MAC_OS_X_VERSION_10_14_4    101404
#define MAC_OS_X_VERSION_10_14_6    101406
```

So maybe we could use `!defined(MAC_OS_X_VERSION_10_7)`.

I guess we would also need to guard the include statement:

```
#ifdef __APPLE__
#include <AvailabilityMacros.h>
#endif
```

Using `/dev/urandom` with the above macro approach might be considered cleaner than bodging in an `arc4random_buf` fill.

cellularmitosis:
just performed a quick local hack to confirm that the `/dev/urandom` approach works as expected under Tiger:

```
cell@emac$ ./build/janet
Janet 1.10.1-local  Copyright (C) 2017-2020 Calvin Rose
janet:1:> (os/cryptorand 1)
@"<"
janet:2:> (os/cryptorand 2)
@"\xAE\x19"
janet:3:> (os/cryptorand 3)
@"s\x15>"
janet:4:> (os/cryptorand 4)
@"L%\xAAR"
```

andrewchambers:
To me either way seems ok to me if we check ( defined(JANET_APPLE) && !defined(MAC_OS_X_VERSION_10_7) ) like you said.

andrewchambers:
looks good to me.

cellularmitosis:
Ok, this is ready for a re-review now.  Thank you for obliging this PowerPC Mac rabbit hole :)

bakpakin:
Looks alright to me - I actually prefer the arc4random code, as it will still work properly when running from within a chroot, but the feature detection looks good. And it is less code. Merging, thanks for the work here.

PR #437, opened 2020-06-30
Add tests for peg/find and peg/find-all

pepe:
Please let me know if I should add some more involved tests.


bakpakin:
This probably pretty good and would have a good start for coverage (is we tracked that). I will merge as is, thanks.

PR #447, opened 2020-07-01
math/nan

cellularmitosis:
This PR adds `math/nan` to the environment.  Is this something we want?


cellularmitosis:
tested locally:

```
cell@flouride-e(nan)$ ./build/janet
Janet 1.10.2-dev-7fb8c4a  Copyright (C) 2017-2020 Calvin Rose
janet:1:> math/inf
inf
janet:2:> math/nan
nan
janet:3:> (= math/nan math/nan)
false
janet:4:> 
```

bakpakin:
Probably wouldn't hurt. Most languages seem to have a NaN literal form are binding. It is missing because the math library was originally adapted from what I remembered from Lua, which is missing such a literal.

PR #448, opened 2020-07-01
Fix janet_table_remove returning the key instead of the value

GrayJack:
I noticed that `janet_table_remove` was returning the key instead of the removed value.

It would be good to have a test for it to avoid regressions, but I have no idea where I should put such a test in the test suit or create a new test suit, so where do I put tests for it?


bakpakin:
Looks good to me, I don't think the old behavior is used anywhere. C level tests are put in the `src/boot` folder, simply add some assertions to one of the files there (there aren't many right now, as most of the testing is through testing Janet code).

GrayJack:
Done :smile:

PR #449, opened 2020-07-01
Add comment in 'not' doc about 'falsey?'

cellularmitosis:
This might be obvious to seasoned lispers, but I thought if we add a note about `falsey?` in the docstring of `not`, then if a newbie hits https://janet-lang.org/api/index.html and searches for "falsey", they will at least see the docstring.

tested locally:

```
cell@flouride-e(falsey)$ ./build/janet
Janet 1.10.2-dev-7e1e307  Copyright (C) 2017-2020 Calvin Rose
janet:1:> (doc not)


    cfunction

    (not x)
    
    Returns the boolean inverse of x.
    falsey? can be implemented as (def falsey? not).


nil
```

### DIFF
--- a/src/core/math.c
+++ b/src/core/math.c
@@ -281,7 +281,8 @@ static Janet janet_not(int32_t argc, Janet *argv) {
 static const JanetReg math_cfuns[] = {
     {
         "not", janet_not,
-        JDOC("(not x)\n\nReturns the boolean inverse of x.")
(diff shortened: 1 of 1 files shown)


bakpakin:
Not going to add, programmers should use `not` instead of `falsey?`. This is standard in pretty much any functional language, and I think more people will look for `not` than `falsey?` for boolean inversion.

cellularmitosis:
no worries, thanks for taking a look :)

PR #453, opened 2020-07-26
Fix build on NetBSD.

niacat:
The NetBSD C library's headers do not expose extensions when compiling with `-std=c99` (as opposed to `-std=gnu99` or no 'std' option), so define `_NETBSD_SOURCE` to get `timegm`, and functions that would otherwise require an `_XOPEN_SOURCE` definition, e.g. `realpath`.

Note that, as with FreeBSD, you need `gmake` to compile janet on NetBSD, and can also install it from packages.

I was previously passing `_NETBSD_SOURCE` through NetBSD's packaging system for the janet package, but this stopped working without patches when `BOOT_CFLAGS` was added. Might as well get it compiling from a fresh git clone now...

[Some more random information about _XOPEN_SOURCE as it applies on Solaris-like systems](https://gist.github.com/jperkin/b08f9108daf8d0ac695067d71f882a9d)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,11 @@ gmake test
 gmake repl
 ```
 
+### NetBSD
+
+NetBSD build instructions are the same as the FreeBSD build instuctions.
+Alternatively, install directly from packages, using `pkgin install janet`.
+
 ### Windows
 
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks good to me, thanks for the fix!

PR #454, opened 2020-07-28
Make sure JANET_NO_NANBOX is defined for ARM targets

### LINKED ISSUE #290: test failure under Windows 10 on ARM64
I've managed to compile Janet for Windows 10 on ARM64 by tweaking the `build_win.bat` a bit (so it would copy the VC redist dll). Everything appears to go fine and the resulting binaries are all `aarch64`. When running the test suite, two tests from suite 0 fail:

~~~~
✘  nan key struct ctor

✘  nan key table ctor
~~~~

All the other tests from the other suites pass. Is there any way for me to help debugging this? I have zero experience with Janet, I was just playing around.

soapdog:
This PR makes sure JANET_NO_NANBOX is defined for ARM targets and fix #290

### DIFF
--- a/src/include/janet.h
+++ b/src/include/janet.h
@@ -225,6 +225,11 @@ extern "C" {
  * To turn of nanboxing, for debugging purposes or for certain
  * architectures (Nanboxing only tested on x86 and x64), comment out
  * the JANET_NANBOX define.*/
+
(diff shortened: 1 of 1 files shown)


bakpakin:
Looks good to me, thanks for the fix.

PR #455, opened 2020-08-04
Initialize PRF with random data when it is enabled.

andrewchambers:


bakpakin:
Looks pretty good, thanks! Good to finally get that one in. Good job with jpm fix as well.

andrewchambers:
Yeah, the inversion of the PRF default made this change a lot easier to do without breaking existing platforms.

PR #456, opened 2020-08-04
Expose janet_init_random_hash_key instead of janet_cryptorand.

andrewchambers:
This seems like a more sensible thing to expose to the public.
janet_cryptorand becomes an internal util function.


andrewchambers:
Just an alternate api I thought of entirely too late.

bakpakin:
I rather like exposing janet_cryptorand, as it is a useful but non trivial function. No problem with adding janet_init_random_hash_key, though.

andrewchambers:
No problem, will close and maybe make a new PR when I have time.

PR #462, opened 2020-08-10
Added net/ready function

dabeaz:
This is an addition to the net/stream functionality that allows one to poll the I/O status of a stream to see if it allows further I/O operations or not.   This is useful in situations where writing to a stream might cause a broken pipe error (which apparently seems to make Janet terminate entirely).  I could not figure out any straightforward mechanism for detecting such errors. Hence, this change.

For context, I needed this change to implement Raft distributed consensus in Janet (see https://github.com/dabeaz/ranet). Raft must gracefully deal with server crashes/restarts.  It can't be made to work if broken writes cause the entire Janet runtime to abort.  See https://github.com/dabeaz/ranet/blob/master/transport.janet for sample usage. 

I don't have strong opinions about the name of this function or its specific implementation.  However, I didn't really want to expose the low-level internals of `poll()` in the high-level interface.

### DIFF
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -542,6 +542,7 @@ static Janet cfun_net_server(int32_t argc, Janet *argv) {
             janet_panic("setsockopt(SO_NOSIGPIPE) failed");
         }
 #endif
+	
 #ifdef SO_REUSEPORT
         if (setsockopt(sfd, SOL_SOCKET, SO_REUSEPORT, &enable, sizeof(int)) < 0) {
             JSOCKCLOSE(sfd);
@@ -610,11 +611,24 @@ static Janet cfun_stream_write(int32_t argc, Janet *argv) {
     }
 }
 
+static Janet cfun_stream_ready(int32_t argc, Janet *argv) {
+    JPollStruct pollfds[1];
+    int r;
+    janet_fixarity(argc, 1);
+    JanetStream *stream = janet_getabstract(argv, 0, &StreamAT);
+    pollfds[0].fd = stream->fd;
+    pollfds[0].events = POLLIN | POLLOUT;
+    pollfds[0].revents = 0;
+    r = JPOLL(pollfds, 1, -1);
+    return janet_wrap_boolean((pollfds[0].revents & POLLHUP) == 0);
+}
+  
 static const JanetMethod stream_methods[] = {
     {"chunk", cfun_stream_chunk},
     {"close", cfun_stream_close},
     {"read", cfun_stream_read},
     {"write", cfun_stream_write},
+    {"ready", cfun_stream_ready},
     {NULL, NULL}
 };
 
@@ -660,6 +674,12 @@ static const JanetReg net_cfuns[] = {
         JDOC("(net/connect host port)\n\n"
              "Open a connection to communicate with a server. Returns a duplex stream "
              "that can be used to communicate with the server.")
+    },
+    {
+        "net/ready", cfun_stream_ready,
+        JDOC("(net/ready stream)\n\n"
+             "Check if a stream can perform I/O. Returns boolean"
+	     )
     },
     {NULL, NULL, NULL}
 };


pepe:
Would this be solved by https://github.com/janet-lang/janet/commit/fb26c9b2c4b49af5b6a21d922b9d34a9a8888291?

dabeaz:
I think it could definitely be solved by having `select()` functionality.  That'd probably be better even.

dabeaz:
Although looking at the referenced file, it's not clear that the `ev/select` or `ev/rselect` will exactly solve the problem.  I'm looking for more of a way to detect an error/hangup condition on the channel to avoid writing more data into it (and to avoid getting a "broken pipe" error on the system).  Having an error/return code on `net/write` would also be an alternative.

bakpakin:
I wonder if just ignoring SIGPIPE would solve the issue - on platforms where it is enabled, we use SO_NOSIGPIPE, as well as MSG_NOSIGNAL when it is defined, which should prevent SIGPIPE. If neither of these work, we may just want to do `signal(SIGPIPE, SIG_IGN)` - I believe write would return nil indicating a failure in this case. This change can also probably be made to the `ev` branch, where I am working on a rewrite of the networking and event loop code.

@dabeaz , what system are you on?

dabeaz:
I'm on OS-X.  Just a clarification---I'm not actually getting a SIGPIPE signal.  In fact, I'm getting nothing at all except that the entire Janet interpreter simply stops running and exits back out to the shell when some kind of write error occurs on the channel. Normally you'd get SIGPIPE, but it looks like that's being suppressed based on the flags being used with the low-level socket operations.

bakpakin:
I can't reproduce the behavior on a macbook, do you have a reproduction? I would prefer to figure out why Janet sometimes aborts on writes before merging this fix.

bakpakin:
I was able to reproduce, and it seems I was getting a SIGPIPE on macos. The check for SO_NOSIGPIPE failed because it was hidden by the feature flags we used in features.h (_POSIX_C_SOURCE). Detecting BSDs and using _BSD_SOURCE instead seems to solve this issue. Once fixed, adding detection for failed writes is straightforward.

@dabeaz If you pull from master and try rebuilding, you can now check if `net/write` fails from its return value, which should be a string like `"write error: Broken pipe"` in the case of the client hanging up on the server or vice versa, `nil` otherwise. Let me know if this solves the problem without `net/ready`.

As for merging the `net/ready` stuff, if the above fixes the issue I don't offhand see any other uses for `net/ready`, which seems to be just a way of checking if `net/write` will succeed.

dabeaz:
I'm still seeing the same behavior on master.   Essentially Janet just "stops".  There's no error message printed or any indication of why.    

I'll try to do some more digging on it.  Sadly, where I'm seeing it also involves threads and fibers.   So... not the easiest debugging ;-).

bakpakin:
Interesting, I was confident that SIGPIPE was the problem, as it fixed exactly the symptoms you described on my macbook.

Here is my minimal server/client example for reproduction - without any sigpipe handling in net.c, I see the same behavior of Janet exiting early. Post 6f2f3fdb6830c9f09f5442e027eb7768c3595b5a, I see an error message indicating a broken pipe. https://gist.github.com/bakpakin/15bb384114f0e80b5bf89db0dc0312f3

dabeaz:
I wonder if it's getting tangled up in some kind of thread-signal mask or something.

andrewchambers:
I also wonder if sigpipe is sometimes useful, though probably not a good default. One nice thing about unix sockets is they don't disconnect arbitrarily, so sigpipe is a useful way of quitting when something has gone terribly wrong.

bakpakin:
Ok, so it looks like it was sigpipe again. We had forgotten to set SO_NOSIGPIPE on sockets created with `net/connect`. The issue never showed up on Linux (or windows), because on those platform we use a different mechanism to prevent sigpipe (on windows I don't believe there is anything like SIGPIPE).

I have verified that I can run ranet with the change without using `net/ready` on macos. It's fun to play around with, and now I can kill servers one at at a time and not have other clients randomly stop, as happens on linux without the change.

I did a quick scan and tried setting SO_NOSIGPIPE on all created sockets, rather than just on clients that connect to a server. Again, fixes are on latest master.

bakpakin:
@andrewchambers Yes, I have come to very much dislike SIGPIPE, and apparently I'm not alone (someone's cathartic rant: https://grok-mctanys.livejournal.com/156198.html). Since Janet is supposed to be embeddable, setting something like `signal(SIGPIPE, SIG_IGN)` is distasteful, even if it often easier than just being careful.

dabeaz:
That latest fix seemed to address the problem of Janet just quitting.  However, I'm never actually seeing the new `net/write` error message either.   Hmmm.

bakpakin:
@dabeaz Forgive me for being pedantic, but the net/write error message is a return value, not an error, so it would not show up unless you print the return value of `net/write`.

I made the following change to the send-message function in transport.janet to use the new behavior and try and verify that we were addressing the original problem.

```
# send-size now only returns nil on success
(defn send-message [msg sock]
  (if (not (send-size (length msg) sock))
    (do
      (if-let [e (net/write sock msg)] (error e))
      )
    (do
      (net/close sock)
      (error "Connection closed")
      )
    )
)
```

When disconnecting a node from raft, I would see an error somewhere inside the receive-message function, but never "Connection closed" as one might expect. But this is actually consistent with how sockets work I think - https://stackoverflow.com/questions/8155607/send-succeeds-even-when-the-connection-is-no-longer-available (net/write calls send) - basically sends can still succeed to a socket that is half closed, which explains things only fail when we try to read again from the socket. Anyways, I'm not sure if `net/ready` would help with this, but maybe it might as it checks POLLIN, but I haven't checked.

Anyways, most of this conversation probably should have lived in a bug report. Oh well.

dabeaz:
As I'm working on a first Janet project, I'll take all the pedantry anyone wants to throw upon it ;-).    All of this makes sense (I've just done too much Python where all errors are exceptions).     Thanks for taking a look at this!

bakpakin:
For anyone viewing this, as of Janet 1.12.1, the error behavior of of `net/*` functions like `net/write` and `net/read` do raise errors instead of returning nil or a string. This is certainly the more intuitive behavior and in line with the Janet/Erlang idiom of "let it fail".

dabeaz:
Thanks for the update!

PR #478, opened 2020-10-06
Allow tree shaking of the core environment when it is unobservable.

andrewchambers:
We can avoid loading the full core environment if the mutable
parts of core are not reachable from the main function. This
is quite a large number of programs.

This change makes 'hello world' 2-3 times faster, and uses approximately
10 times less ram.

### DIFF
--- a/jpm
+++ b/jpm
@@ -585,7 +585,7 @@
   (string (string/slice path 0 (- -1 (length modext))) statext))
 
 (defn- make-bin-source
-  [declarations lookup-into-invocations]
+  [declarations lookup-into-invocations global-state-reachable]
   (string
     declarations
     ```
@@ -609,12 +609,25 @@ int main(int argc, const char **argv) {
 #endif
 
     janet_init();
+    int handle = janet_gclock();
 
-    /* Get core env */
+    ```
+    (if global-state-reachable
+    ```
+
+    JanetTable *temptab = NULL;
     JanetTable *env = janet_core_env(NULL);
     JanetTable *lookup = janet_env_lookup(env);
-    JanetTable *temptab;
-    int handle = janet_gclock();
+    ```
+
+    ```
+
+    JanetTable *temptab = NULL;
+    JanetTable *env = janet_table(8);
+    JanetTable *lookup = janet_make_core_lookup_table();
+    ```
+    )
+    ```
 
     /* Load natives into unmarshalling dictionary */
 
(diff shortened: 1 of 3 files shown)


andrewchambers:
I feel this approach could be improved/altered with any of:

- A jpm option for this.
- Tweaks to the C code in peg, such that it does not take things directly from the root-env.

bakpakin:
I like this idea a lot, and rather than implement hacks to get around certain issues, I think it might be worth considering changing the core library to avoid any calls (directly or indirectly) `janet_core_env` in all cfunctions.

bakpakin:
BTW, this draft is not dead, just pending changes to the thread module, which seem to be the biggest hurdle in adding the functionality. , It is currently quite difficult to remove any references to the `janet_core_env` function in the thread module.

bakpakin:
The commit I have added have taken some of the code from here, but reworked. Mainly, we don't use the auto-dectection since there are cases where the auto-detection can fail, say when loading a native module that tries to use `janet_dobytes` at module load time. Other than that, it seems to be working well, and definitely cuts down start time for hello, world!

Example usage:

```
# hello.janet
(defn main [&] (print "hello, world!"))
```

```
$ jpm --no-core quickbin hello.janet hello
generating executable c source...
compiling hello.c to build/hello.o...
linking hello...
$ time ./hello
hello, world!

real	0m0.001s
user	0m0.001s
sys	0m0.000s
```

sogaiu:
Here are some local results:
```
$ jpm --no-core quickbin hello.janet hello
generating executable c source...
compiling hello.c to build/hello.o...
linking hello...

$ multitime -n10 -s1 ./hello
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
===> multitime results
1: ./hello
            Mean        Std.Dev.    Min         Median      Max
real        0.003       0.002       0.001       0.003       0.006       
user        0.001       0.001       0.000       0.001       0.003       
sys         0.002       0.002       0.000       0.002       0.006       

$ jpm quickbin hello.janet hello2
generating executable c source...
compiling hello2.c to build/hello2.o...
linking hello2...

$ multitime -n10 -s1 ./hello2
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
hello, world!
===> multitime results
1: ./hello2
            Mean        Std.Dev.    Min         Median      Max
real        0.008       0.007       0.002       0.004       0.020       
user        0.006       0.006       0.000       0.003       0.016       
sys         0.002       0.002       0.000       0.003       0.004       
```

andrewchambers:
nice, thanks, it should be useful for programs like jfmt.

PR #481, opened 2020-10-07
Use "module" instead of "namespace" in docstring

sogaiu:
This PR is a suggestion to not use the word "namespace" in the docstring of `use` to reduce / prevent confusion.


bakpakin:
Sounds good, as long as its clear that this is simply aliasing symbols.

PR #484, opened 2020-10-07
Add fastpath to table find.

andrewchambers:
I think this still needs benchmarking to prove it does anything useful.

### DIFF
--- a/src/core/table.c
+++ b/src/core/table.c
@@ -86,7 +86,26 @@ JanetTable *janet_table(int32_t capacity) {
 /* Find the bucket that contains the given key. Will also return
  * bucket where key should go if not in the table. */
 JanetKV *janet_table_find(JanetTable *t, Janet key) {
-    return (JanetKV *) janet_dict_find(t->data, t->capacity, key);
+    /* If we have deleted an item do slower search that checks tombstone values. */
+    if (t->deleted)
+        return (JanetKV *) janet_dict_find(t->data, t->capacity, key);
+
+    JanetKV *buckets = t->data;
+    int32_t i;
+    int32_t cap = t->capacity;
+    int32_t index = janet_maphash(cap, janet_hash(key));
+
+    for (i = index; i < cap; i++) {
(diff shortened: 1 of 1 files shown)


andrewchambers:
This implementation is the same as struct find, so I wonder if they could be combined.

andrewchambers:
ehh,  actually this pr is broken because deleted gets decremented.

bakpakin:
Yes, the invariant looks fine. deleted should always have the number of tombstone values in the table, so if it is zero we could fast track. However, I wonder if the micro-optimization is actually worth it. I suspect the improvement will be almost zero.

andrewchambers:
tab.janet
```
(def t @{})
(for i 0 10000
  (put t i i))
(for i 0 20000
  (get t i))
```

```
[ac@black:~/src/janet]$ valgrind --tool=callgrind ./janet-slow ./tab.janet 
==84021== Callgrind, a call-graph generating cache profiler
==84021== Copyright (C) 2002-2017, and GNU GPL'd, by Josef Weidendorfer et al.
==84021== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==84021== Command: ./janet-slow ./tab.janet
==84021== 
==84021== For interactive control, run 'callgrind_control -h'.
==84021== 
==84021== Events    : Ir
==84021== Collected : 312675345
==84021== 
==84021== I   refs:      312,675,345

[ac@black:~/src/janet]$ valgrind --tool=callgrind ./janet-fast ./tab.janet 
==84022== Callgrind, a call-graph generating cache profiler
==84022== Copyright (C) 2002-2017, and GNU GPL'd, by Josef Weidendorfer et al.
==84022== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==84022== Command: ./janet-fast ./tab.janet
==84022== 
==84022== For interactive control, run 'callgrind_control -h'.
==84022== 
==84022== Events    : Ir
==84022== Collected : 312413487
==84022== 
==84022== I   refs:      312,413,487

```

PR #498, opened 2020-11-04
Fix asm example

pepe:
As I am wandering around ev, I have found this example was not working in its old version, so I fixed it. I also reformat it with the spork/fmt.


bakpakin:
Yeah, the signature for asm changes slightly a few version back, thanks.

Issue #509, opened 2020-11-27
seg fault on `make test`


Saikyun:
On MacOS High Sierra, commit https://github.com/janet-lang/janet/commit/ca7c5b8b10d28e1351598e24567e990e3e1564f2.

I was able to build a3e812b
```
$ git clone https://github.com/janet-lang/janet test-janet && cd test-janet && make && make test
<redacted>
Running test suite 9 tests...
  
✔ made server 1: <core/stream 0x7FC6381041C0>
✔ echo hello: true
✔ echo world: true
✔ echo abcdabcdabcdabcdabcdabcdabcdab...: true
✔ ev/pipe 1: true
✔ ev/cancel 1: true
/bin/sh: line 1: 77834 Segmentation fault: 11  ./build/janet "$f"
make: *** [test] Error 139
```

Hope you don't feel bombarded. :) Wish you a nice day!

Saikyun:
If I do this, the tests pass, so I'm guessing the issue is here somewhere. :)
```
modified   src/core/ev.c
@@ -1684,8 +1684,8 @@ static Janet cfun_ev_call(int32_t argc, Janet *argv) {
     janet_arity(argc, 1, -1);
     JanetFunction *fn = janet_getfunction(argv, 0);
     JanetFiber *fiber = janet_fiber(fn, 64, argc - 1, argv + 1);
-    fiber->env = janet_table(0);
-    fiber->env->proto = janet_current_fiber()->env;
+    //    fiber->env = janet_table(0);
+    //    fiber->env->proto = janet_current_fiber()->env;
     if (NULL == fiber) janet_panicf("invalid arity to function %v", argv[0]);
     janet_schedule(fiber, janet_wrap_nil());
     return janet_wrap_fiber(fiber);
```

PR #510, opened 2020-11-27
moved error handling in ev/call

Saikyun:
fix for https://github.com/janet-lang/janet/issues/509


Saikyun:
Sadly don't know what causes the windows error, and can't test myself on windows atm. If you know what's wrong, feel free to either tell me what to do, or just make your own solution and ignore this PR. :)

pepe:
I think @bakpakin mentioned, that windows net code is not that stable yet.

pepe (approved):
LGTM. It makes a sense to check for NULL before accessing members of the fiber.

bakpakin:
Good catch, I fixed this before checking this issue when I saw CI failing, but that is indeed the fix.

Issue #524, opened 2020-12-05
`describe` used by string, symbol, keyword, buffer? Docs correct?


uvtc:
The docs for `string`, `symbol`, `keyword`, and `buffer` all mention use of `describe` when necessary for converting to bytes.

But the docs for `describe` don't mention anything about bytes (but instead say it's for a human-readable description of its arg).

What `describe` previously used for something other than it's current use?

Also, what is the difference between `(string/format "%j" s)` (which I think prints out in jdn) vs `(describe s)`? JDN seems human-readable, but I think I'm missing something here.

Thanks, and thanks @sogaiu for pointing this out to me.

sogaiu:
I had a fresh look just now and I think I may have a slightly better understanding.

The TLDR is -- perhaps the reference to `describe` is this: https://github.com/janet-lang/janet/blob/7e1d09599677c3dc5bc4ef5944e6d9e0c612e2a8/src/core/pp.c#L246-L249

The longer version is...

IIUC, `describe` is a wrapper for `janet_core_describe`:
```
static Janet janet_core_describe(int32_t argc, Janet *argv) {
    JanetBuffer *b = janet_buffer(0);
    for (int32_t i = 0; i < argc; ++i)
        janet_description_b(b, argv[i]);
    return janet_stringv(b->data, b->count);
}
```
and that calls `janet_description_b`.

`string` is a wrapper for `janet_core_string`:
```
static Janet janet_core_string(int32_t argc, Janet *argv) {
    JanetBuffer *b = janet_buffer(0);
    for (int32_t i = 0; i < argc; ++i)
        janet_to_string_b(b, argv[i]);
    return janet_stringv(b->data, b->count);
}
```

Note that it calls `janet_to_string_b` for each of the "values", where "values" is from "Values are converted to bytes if they are not byte sequences" (or the `parts` argument to `string`, etc.).

`janet_to_string_b` has a default clause: https://github.com/janet-lang/janet/blob/7e1d09599677c3dc5bc4ef5944e6d9e0c612e2a8/src/core/pp.c#L246-L249 which perhaps one might say is the "heart" of describe.  May be the defauilt clause counts as "if ... not byte sequence".

The implementations of `symbol`, `keyword`, and `buffer` are similar to that of `string`.  Thus perhaps it's a fair description of:

> Values are converted to bytes via describe if they are not byte sequences.

[...]

PR #525, opened 2020-12-06
light markup in some docs in corelib

uvtc:
For #524

### DIFF
--- a/src/core/corelib.c
+++ b/src/core/corelib.c
@@ -552,31 +552,31 @@ static const JanetReg corelib_cfuns[] = {
     },
     {
         "string", janet_core_string,
-        JDOC("(string & parts)\n\n"
-             "Creates a string by concatenating values together. Values are "
-             "converted to bytes via `describe` if they are not byte sequences. "
+        JDOC("(string & pieces)\n\n"
+             "Creates a string by concatenating the pieces together. If a piece "
+             "is not a byte sequence, it is converted to bytes via `describe`. "
              "Returns the new string.")
     },
     {
         "symbol", janet_core_symbol,
-        JDOC("(symbol & xs)\n\n"
-             "Creates a symbol by concatenating values together. Values are "
-             "converted to bytes via `describe` if they are not byte sequences. Returns "
-             "the new symbol.")
+        JDOC("(symbol & pieces)\n\n"
+             "Creates a symbol by concatenating the pieces together. If a piece "
+             "is not a byte sequence, it is converted to bytes via `describe`. "
+             "Returns the new symbol.")
     },
     {
         "keyword", janet_core_keyword,
-        JDOC("(keyword & xs)\n\n"
-             "Creates a keyword by concatenating values together. Values are "
-             "converted to bytes via `describe` if they are not byte sequences. Returns "
-             "the new keyword.")
+        JDOC("(keyword & pieces)\n\n"
+             "Creates a keyword by concatenating the pieces together. If a piece "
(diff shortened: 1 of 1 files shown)


sogaiu:
@uvtc What do you think of the idea of changing the argument in all four cases to be the same?

For `string` it's `parts`, whereas for the other three it's `xs`.

Also the docstring prose uses the word "Values".  It seems like that could be replaced with whatever the argument name ends up being (e.g. `xs`, `parts`, or `values`).

uvtc:
My guess is that "xs" (or even "args") is a sort of "catch-all" for when you don't have a more descriptive word to use for the extra args.

For these cases, I think that "parts" (as used in the doc for `string`) is good, but "pieces" would be even better. (Parts makes me think of different parts that make up an engine, but pieces makes me think of (possibly disparate) items that we're stitching together to make the final thing).

If the docs are changed, will they become out of sync with some source code elsewhere?

sogaiu:
Thanks for your latest attempt.

Regarding `parts` vs `pieces` -- to be honest I don't have similar associations.  I find `parts` slightly better because it's shorter :)  I think it may also have the advantage of being harder to mispell.

Of course, it's ultimately not up to me, but I'm also fine with using `xs` -- that seems to be used elsewhere and IIUC it's even used in code and docs in other programming languages.

Not sure about being out-of-sync with source -- actually, I don't quite follow what you mean.  Would you mind elaborating?

I did find there is at least one other use of `parts`:
```
   (array/concat arr & parts)

    Concatenates a variable number of arrays (and tuples) into the first
    argument which must be an array. If any of the parts are arrays or
    tuples, their elements will be inserted into the array. Otherwise,
    each part in parts will be appended to arr in order. Return the
    modified array arr.
```

uvtc:
> Not sure about being out-of-sync with source -- actually, I don't quite follow what you mean. Would you mind elaborating?

As in, do the function params here in the docs need to match up with (stay in sync with) implementation elsewhere?

sogaiu:
I confess I don't get how that could be an issue.

uvtc:
Sorry, my C is very rusty... I was thinking maybe there was a C-function signature somewhere (which I'm unaware of) that the doc text was supposed to match up with. But it sounds like I'm off-base and that's not the case.

sogaiu:
Thanks for explanation.

For reference, I've search across the janet repository and found the following files use `parts`:

* src/boot/boot.janet (e.g. https://github.com/janet-lang/janet/blob/master/src/boot/boot.janet#L1111 -- though it is `def`ed 5 times total)
* tools/gendoc.janet (e.g. https://github.com/janet-lang/janet/blob/master/tools/gendoc.janet#L105)
* src/core/array.c ([in the docs for `array/concat`](https://github.com/janet-lang/janet/blob/master/src/core/array.c#L346-L349))
* src/core/corelib.c ([in the doc params for `string`](https://github.com/janet-lang/janet/blob/master/src/core/corelib.c#L555))
* src/core/string.c ([in `cfun_string_join`](https://github.com/janet-lang/janet/blob/master/src/core/string.c#L432))

If you don't mind browsing how it's used in those places, perhaps it might factor into a leaning as to what to do for the current case.

My own sense is that `parts` is used less than `xs` (I did a similar search) and I think `xs` might be more familiar for some folks.

andrewchambers:
xs is the plural of x afaik. x is just a generic variable.

sogaiu:
Yes, that has been my understanding too.

uvtc:
Thanks, @sogaiu . Since "parts" doesn't really convey any extra meaning here, I changed it to `xs`.

PR #526, opened 2020-12-07
Tweak comment for janet_fiber_popframe

sogaiu:
The return type of `janet_fiber_popframe` is void, but the associated comment for the function mentions returning some value.

This PR assumes the current behavior of the function is fine and thus modifies the comment to remove mention of returning something.


pepe (approved):
LGTM

Issue #538, opened 2020-12-22
Hash for tuples (arrays) results in a lot of collisions and significant slow down for tables, distinct, etc


felixr:
> TLDR; `janet_array_calchash()` produces a lot of collisions causing slow down for anything using the hashed

 
I notice my (ugly) solution to day 22 of the Advent of Code  was quite slow, so I profiled it and noticed a lot of time spent in `janet_equal()`. Since my code (see below) does not do many direct comparisons I assumed it had to do with the table lookups. So I had a look at the hash function and replaced it with [Python's hash function for tuples](https://github.com/python/cpython/blob/63298930fb531ba2bb4f23bc3b915dbf1e17e9e1/Objects/tupleobject.c#L406). The speed up was significant.

My solution and `(distinct tuples)` with ~1.5M distinct tuples were suddenly both **~5x faster**. 

Output of running the code (below) without and with my patch:

```
janet solve.janet
Completed in 9.00 seconds
(distinct) completed in 17.00 seconds
Number of hash collisions: 115405 (7.70%)

janet_opt solve.janet
Completed in 2.00 seconds
(distinct) completed in 3.00 seconds
Number of hash collisions: 294 (0.02%)
```

```clojure
(def txt "2 31 14 45 33 18 29 36 44 47 38 6 9 5 48 17 50 41 4 21 42 23 25 28 3\n26 16 27 12 49 32 19 46 37 15 10 30 11 24 1 40 7 8 43 34 20 35 22 39 13")
 
(defn get-input []
  (->> 
    (string/split "\n" txt)
    (map (partial string/split " "))
    (map (fn [arr] (keep scan-number arr)))))
                
(def all-seen @[])
(var collect false)

(var winning-hand nil)
(defn play [[pa pb]]
  (def seen @{})
  (var ret nil)
  (var winner nil)
  (while (and (not (empty? pa)) (not (empty? pb)))
    (set winner nil)
    (set ret nil)
    (let [ca (first pa)
          cb (first pb)
          curr (tuple ;pa "|" ;pb)
          prev? (get seen curr)]
      (put seen curr true)
      (if collect
        (array/push all-seen curr))
      (if prev? 
        (set ret :p1)
        (do
          (array/remove pa 0 1)
          (array/remove pb 0 1)
          (if (and (<= ca (length pa)) (<= cb (length pb)))
            (do
              (set winner (play [(array/slice pa 0 ca) (array/slice pb 0 cb)])))
            (if (> ca cb)
[...]

PR #539, opened 2020-12-23
Improve hash for arrays and KVs 

felixr:
#538 

The current standard hash is [djb2](http://www.cse.yorku.ca/~oz/hash.html#djb2) (aka Bernstein hash). When used to combine the hash values of array or KV elements it performs poorly when the elements are numbers.

This can be easily demonstrated by created tables with tuples or structs as keys whose elements are numbers:

```sh
time janet -e '(def tbl @{}) (loop [x :in (range 1000) y :in (range 1000)] (put tbl (tuple 0 x 1 y) true))'
40.78s user 0.07s system 99% cpu 40.862 total
time janet -e '(def tbl @{}) (loop [x :in (range 1000) y :in (range 1000)] (put tbl {0 x 1 y} true))'
48.98s user 0.15s system 99% cpu 49.136 total
```

And with the changes of this PR:

```sh
time build/janet -e '(def tbl @{}) (loop [x :in (range 1000) y :in (range 1000)] (put tbl (tuple 0 x 1 y) true))'
5.89s user 0.07s system 99% cpu 5.965 total

time build/janet -e '(def tbl @{}) (loop [x :in (range 1000) y :in (range 1000)] (put tbl {0 x 1 y} true))'
3.80s user 0.09s system 99% cpu 3.885 total
```

See also:
https://www.boost.org/doc/libs/1_35_0/doc/html/boost/hash_combine_id241013.html

### DIFF
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -227,19 +227,21 @@ int32_t janet_string_calchash(const uint8_t *str, int32_t len) {
 /* Computes hash of an array of values */
 int32_t janet_array_calchash(const Janet *array, int32_t len) {
     const Janet *end = array + len;
-    uint32_t hash = 5381;
-    while (array < end)
-        hash = (hash << 5) + hash + janet_hash(*array++);
+    uint32_t hash = 0;
+    while (array < end) {
+        uint32_t elem = janet_hash(*array++);
+        hash ^= elem + 0x9e3779b9 + (hash<<6) + (hash>>2);
+    }
     return (int32_t) hash;
 }
 
 /* Computes hash of an array of values */
 int32_t janet_kv_calchash(const JanetKV *kvs, int32_t len) {
     const JanetKV *end = kvs + len;
-    uint32_t hash = 5381;
+    uint32_t hash = 0;
     while (kvs < end) {
-        hash = (hash << 5) + hash + janet_hash(kvs->key);
-        hash = (hash << 5) + hash + janet_hash(kvs->value);
+        hash ^= janet_hash(kvs->key) + 0x9e3779b9 + (hash<<6) + (hash>>2);
+        hash ^= janet_hash(kvs->value) + 0x9e3779b9 + (hash<<6) + (hash>>2);
         kvs++;
     }
     return (int32_t) hash;
--- a/src/core/value.c
+++ b/src/core/value.c
@@ -261,6 +261,16 @@ int32_t janet_hash(Janet x) {
         case JANET_STRUCT:
             hash = janet_struct_hash(janet_unwrap_struct(x));
             break;
+        case JANET_NUMBER:
+            hash = (int32_t)janet_unwrap_number(x);
+            hash = ((hash >> 16) ^ hash) * 0x45d9f3b;
+            hash = ((hash >> 16) ^ hash) * 0x45d9f3b;
+            hash = (hash >> 16) ^ hash;
+
+            uint32_t lo = (uint32_t)(janet_u64(x) & 0xFFFFFFFF);
+            hash ^= lo + 0x9e3779b9 + (hash<<6) + (hash>>2);
+
+            break;
         case JANET_ABSTRACT: {
             JanetAbstract xx = janet_unwrap_abstract(x);
             const JanetAbstractType *at = janet_abstract_type(xx);


felixr:
Another day of Advent of Code and another benchmark. This time showing we can do better than this PR:

3 runs:

1.  Usings python's tuple hashing
2.  Using this PR plus xxh3 for strings/symbols
3. Just using this PR

```
time ../janets/janet_pyhash hexagon_tuples.janet
1.92s user 0.00s system 99% cpu 1.921 total

time ../janets/janet_lto_xxh3 hexagon_tuples.janet 
4.51s user 0.00s system 99% cpu 4.515 total

time ../janets/janet_lto_opt hexagon_tuples.janet 
5.09s user 0.02s system 99% cpu 5.106 total
```

For this particular code (https://gist.github.com/felixr/ab43840e926e9899078ab9a85052db09) python's tuple hashing is much better than this PR. It also isnt slower on all my other benchmarks. So I will need to spend more time benchmarking.

But this PR is still a clear improvement over the status quo IMO.

felixr:
I ran some simple benchmarks with more variations.

1. *orig* umodified janet
2. *pyhash* using Python's tuple hash for `janet_array_calchash()` and `janet_kv_calchash()`
3. *numhash* modified the hash for numbers by rotating bits (`hi = (hi << 19) | (hi>>13);`)
4. *boost* this PR usings Boost's way of combining hash values into a new hash value
5. *boost_numhash* Changes from (3) and (4) combined.

benchmarks

I would not read much into any small differences, as my benchmark setup was not super controlled and I did not run many runs.

The interesting benchmarks are [hexagon_tuples](https://gist.github.com/felixr/ab43840e926e9899078ab9a85052db09). int_kv_hash (below) and tuple_hash (see #538).
Other benchmarks are taken from https://github.com/MikeBeller/janet-benchmarksgame


```
# int_kv_hash.janet
(def input (string/split "\n" (slurp "/usr/share/dict/words"))
(def words @{})
(defn word-to-struct [word] (struct ;(string/bytes word) ;(if (> (mod (length word) 2) 0) [1] [])))
(each word input (put words (word-to-struct word) word))
(for i 0 1000 (def word (get input (mod (* i 3491) (length input)))))
```

leahneukirchen:
As an outsider, I'd recommend adding at least a comment to each hash function what it's name is where it originates from (paper or project).

felixr:
@leahneukirchen definitely agree that it is a good idea to document why a hash function was chosen.

Since I am also an 'outsider' (I just started using Janet and dug into some performance issues I noticed) I can only guess that the original hash function (ie. djb2/bernstein) was used because it is a simple hash function that works well enough in a lot of cases and definitely doing quite well for string hashing. 
However if you look at the [SMhasher results](https://github.com/rurban/smhasher) then you will notice that bernstein is not that great. 

If you compile janet with `JANET_PRF` it will use HalfSipHash which should be a much better choice. Although, I would say that xxh3 is probably an even better choice, because a lot of times we hash very short strings (e.g symbol names or struct keys). I believe rust used to use siphash and now moved to xxh3; I might be wrong....

That said, my patches are not about string hashing. They are about two things:

1. Combining hash values of tuple/kv elements into a new hash
2. Hash values for numbers

For (1) janet also uses the bernstein hash, but from my experiments it looks like it performs very poorly when the input hashes (the ones that are combined) are relatively similar. This happens currently with numbers, because the 'hash' for whole numbers is currently the high bits of their double representation. Which leads to my change to (2) which hashes numbers in a way suitable for integers and combines it with the lower bits of the double.

For (1) my reference is Boost. From looking around it is an approach that has been copied a lot and seems to be working. And for (2) I used the suggestion from a SO answer (linked in commit).

Today I started looking into the issues a bit more methodically and started  https://github.com/felixr/janet-benchmarks Which is currently a collection of benchmarks which make this PR look good :-)   

So far I haven't found any issues with my proposed changes, but I have not yet tried hard to find benchmarks where my changes perform poorly

bakpakin:
Thanks @felixr, in general it looks pretty good and this is a useful benchmark to have done. I have changed Janet's number hashing code in 13d8d11011613e1b9d71bfe1122bab5c82b0d211 to be similar to yours but be more general to floating point numbers and have better properties when hashing non-integer values - it is based on [frexp](https://pubs.opengroup.org/onlinepubs/009695399/functions/frexp.html) with handling for NaNs, zeroes, and infinite numbers. This has multiple benefits outside this PR. 

I am still interested in how the other tuple and kv hashes will perform with this change - my hunch is that they will not much impact performance but if they are still an improvement I would be happy to merge them.

bakpakin:
@felixr I have tested your code with the changed number hashing and thankfully it still brings the same improvement so i have merged it.

felixr:
@bakpakin Thanks for merging most of my changes. I will have to add some benchmarks with floating point numbers. Tbh I had not payed much attention to them as they less likely to be used as keys as even equality testing on floats is unreliable.

In my current benchmark suite the `frexp` based hash performs significantly worse on two of the benchmarks:

image

This is comparing:
 * https://github.com/janet-lang/janet/commit/a2c837a99c4013fd8b7287e0f64453d8644742d8
 * https://github.com/felixr/janet/commit/0d46352ff447049b52e5a20152b044b04b5238e2

In the latter I reverted to my hash for numbers, but kept the inf/nan/0 handling.

Complete results are at https://github.com/felixr/janet-benchmarks/tree/main/results 
The repo also includes the code generate this chart.

bakpakin:
> Tbh I had not payed much attention to them as they less likely to be used as keys as even equality testing on floats is unreliable.

This is simply untrue - operations on floats may accumulate errors, but checking equality between two floating point numbers is not particularly mystical - if they have the same bit pattern they should have the same hash.

As for your benchmarks, I would be interested how your code handles something like `(let [x @{}] (repeat 1e6 (put x (math/random) true)))`. If all real number between 0 and 1 are hashed to 0, we have a bit of a problem.

bakpakin:
So doing some experiments with regards to number of hash collisions rather than execution time, I find that the full `frexp` based implementation does have slightly better properties than integer cast based implementation, but the difference is neglible.

The below experiment calculates the number of unique hashes generated from 1 million random numbers between 0 and 1. A perfect hash function should  approach 1,000,000 but fall short because of the Birthday Paradox, but this is quick way to compare hash functions on various inputs.
```
(length (frequencies (map (fn [_] (hash (math/random))) (range 1e6))))
```

Anyways, comparing the inputs I see about 250 collisions for the frexp based implementation on average vs 350 for the cast based implementation. According to https://math.stackexchange.com/questions/35791/birthday-problem-expected-number-of-collisions, we should expect about 34 collisions for an optimal hash function (but perhaps my reasoning is off, as we could choose the same exact number from `(math/random)` more than once), so perhaps neither of the hash functions are particularly great. In which case, the measured performance of table inserts is probably a better criteria.

@felixr Thanks for the work here. The other useful property of the frexp based hashing is that it is consistent across architectures but how useful that is, especially in the presence of pointer hashing and optional SipHash, is questionable. I will merge your hash function since it does seem to perform better on some benchmarks.

felixr:
> This is simply untrue - operations on floats may accumulate errors, but checking equality between two floating point numbers is not particularly mystical - if they have the same bit pattern they should have the same hash.

Yes, you are of course right. I was not very precise in what I said. What I meant was that equality testing on floats is usually a bad idea because you can easily accumulate errors. Eg.
`janet -e '(print  (= .3 (+ .2 .1)) " " .3 " " (+ .2 .1))'`  will output `false 0.3 0.3` Which is something unexpected to quite a few people.
And if you now use your two different `0.3`s in keys you might not get what you want....

Whether to use floats in keys or not is of course not super important here, and I did check that any changes I proposed would work with floats. But I did not focus much on performance for this use case.

> So doing some experiments with regards to number of hash collisions rather than execution time... 

Interesting, I did not look at how close we were overall to the theoretical optimum. I actually also started out by looking at the number of collisions in my benchmarks and not the run time directly. That is how I ended up looking at `sort`. I tried to sort all hashes to count duplicates (so I would not use a table and hashes for that) and then run into the issue that the hashes I generated were all in order and sort took longer than I expected ... :-)

@bakpakin Thank you for your great work on Janet! I really enjoy playing with it.

PR #540, opened 2020-12-26
Improve quicksort to avoid worst case performance on sorted input

felixr:
The current implementation will have quadratic behaviour for already
sorted arrays because it picks the last element as pivot. In an sorted
array this splits the array repeatedly into the biggest value and all
other values.

The implementation in this commit uses the *median of three* as pivot.

`janet -e "(sort (range 10000))"` to reproduce quadratic behaviour.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -790,35 +790,38 @@
 ###
 ###
 
-(defn- sort-part
-  [a lo hi by]
-  (def pivot (in a hi))
-  (var i lo)
-  (forv j lo hi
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM, thanks. We may want to improve sort more in the future (move to C), but a pure janet sort has the advantage that we can yield across the comparator. In practice, this may not be at all important though.

andrewchambers:
@bakpakin just a thought, C based sorting could perhaps yield as sorting is the kind of operation that can be retried/re-entered.

PR #543, opened 2020-12-28
Another improvement of quicksort

felixr:
Ultimately it might be a good idea to implement Timsort to get a fast and stable sorting algorithm, but it turns out that is quite a bit of work... :-)

This change makes sort switch to insertion sort when an array is smaller than 32 and it also uses an array as stack instead of recursive function call. It looks like this gives us a ~8x speed-up:

**modified**
```
time build/janet -e '(math/seedrandom 42)(sort (map (fn [x] (math/random)) (range 1000000)) >)'
time build/janet -e '(sort (range 1000000) <)'
time build/janet -e '(sort (range 1000000) >)'
time build/janet -e '(sort (array ;(range 10000 100000) ;(map - (range 100000)) ;(range 100000)) >)'

0.49s user 0.01s system 99% cpu 0.503 total
0.26s user 0.00s system 99% cpu 0.265 total
0.28s user 0.00s system 99% cpu 0.284 total
0.08s user 0.00s system 99% cpu 0.084 total
```

**unmodified**
```
time janet -e '(math/seedrandom 42)(sort (map (fn [x] (math/random)) (range 1000000)) >)'
time janet -e '(sort (range 1000000) <)'
time janet -e '(sort (range 1000000) >)'
time janet -e '(sort (array ;(range 10000 100000) ;(map - (range 100000)) ;(range 100000)) >)'

3.63s user 0.01s system 99% cpu 3.644 total
2.59s user 0.00s system 99% cpu 2.598 total
2.46s user 0.00s system 99% cpu 2.467 total
8.34s user 0.01s system 99% cpu 8.363 total
```

Further speed ups could be achieved by detecting increasing/decreasing runs (like in Timsort) and maybe by using binary insertion sort.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -795,34 +795,45 @@
     a
     (if (not= (> b a) (> b c)) b c)))
 
-(defn sort-help [a lo hi by]
-  (when (< lo hi)
-    (def pivot
-      (median-of-three (in a hi) (in a lo)
-                       (in a (math/floor (/ (+ lo hi) 2)))))
-    (var left lo)
-    (var right hi)
-    (while true
-      (while (by (in a left) pivot) (++ left))
-      (while (by pivot (in a right)) (-- right))
-      (when (<= left right)
-        (def tmp (in a left))
-        (set (a left) (in a right))
-        (set (a right) tmp)
-        (++ left)
-        (-- right))
-      (if (>= left right) (break)))
-    (sort-help a lo right by)
-    (sort-help a left hi by))
+(defn- insertion-sort [a lo hi by]
+  (for i (+ lo 1) (+ hi 1)
+    (def temp (in a i))
+    (var j (- i 1))
+    (while (and (>= j lo) (by temp (in a j)))
+      (set (a (+ j 1)) (in a j))
+      (-- j))
+
+    (set (a (+ j 1)) temp))
   a)
 
 (defn sort
   "Sort an array in-place. Uses quick-sort and is not a stable sort."
   [a &opt by]
-  (sort-help a 0 (- (length a) 1) (or by <)))
+  (default by <)
+  (def stack @[[0 (- (length a) 1)]])
+  (while (not (empty? stack))
+    (def [lo hi] (array/pop stack))
+    (when (< lo hi)
+      (when (< (- hi lo) 32) (insertion-sort a lo hi by) (break))
+      (def pivot (median-of-three (in a hi) (in a lo) (in a (math/floor (/ (+ lo hi) 2)))))
+      (var left lo)
+      (var right hi)
+      (while true
+        (while (by (in a left) pivot) (++ left))
+        (while (by pivot (in a right)) (-- right))
+        (when (<= left right)
+          (def tmp (in a left))
+          (set (a left) (in a right))
+          (set (a right) tmp)
+          (++ left)
+          (-- right))
+        (if (>= left right) (break)))
+      (array/push stack [lo right])
+      (array/push stack [left hi])))
+  a)
 
 (undef median-of-three)
-(undef sort-help)
+(undef insertion-sort)
 
 (defn sort-by
   `Returns a new sorted array that compares elements by invoking


bakpakin:
LGTM. I think it would be worth moving the sort implementation to corelib.c and working from there - I have a feeling that would be an easy 10x speed up in the common case, and the main caveat would just be that one could not yield inside a custom comparator function. I think that is an acceptable trade off, though.

andrewchambers:
I would think the libc qsort function would be appropriate, just because it is called qsort, does not necessarily mean the system requires the quick sort algorithm.

andrewchambers:
I did an experiment with c qsort:

t.janet:
```
(math/seedrandom 42)
(sort (map (fn [x] (math/random)) (range 1000000)))
```

results:

```
[ac@black:~/src/janet]$ hyperfine './janet-c-qsort ./t.janet' './janet-janet-qsort ./t.janet' --export-markdown results.md
Benchmark #1: ./janet-c-qsort ./t.janet

  Time (mean ± σ):     259.8 ms ±   1.2 ms    [User: 253.5 ms, System: 6.1 ms]
 
  Range (min … max):   258.5 ms … 261.9 ms
 
Benchmark #2: ./janet-janet-qsort ./t.janet

  Time (mean ± σ):      4.684 s ±  0.055 s    [User: 4.674 s, System: 0.007 s]
 
  Range (min … max):    4.597 s …  4.769 s
 
Summary

  './janet-c-qsort ./t.janet' ran
   18.03x faster than './janet-janet-qsort ./t.janet'
```

Here is the patch if you want to play with it:

```
index 1dc035a8..fddb2651 100644
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -790,37 +790,6 @@
 ###
 ###
 
-(defn- sort-part
-  [a lo hi by]
-  (def pivot (in a hi))
-  (var i lo)
-  (forv j lo hi
-    (def aj (in a j))
-    (when (by aj pivot)
-      (def ai (in a i))
-      (set (a i) aj)
-      (set (a j) ai)
-      (++ i)))
-  (set (a hi) (in a i))
-  (set (a i) pivot)
-  i)
-
-(defn- sort-help
-  [a lo hi by]
-  (when (> hi lo)
-    (def piv (sort-part a lo hi by))
-    (sort-help a lo (- piv 1) by)
-    (sort-help a (+ piv 1) hi by))
-  a)
-
-(defn sort
-  "Sort an array in-place. Uses quick-sort and is not a stable sort."
-  [a &opt by]
-  (sort-help a 0 (- (length a) 1) (or by <)))
-
-(undef sort-part)
-(undef sort-help)
-
 (defn sort-by
   `Returns a new sorted array that compares elements by invoking
   a function on each element and comparing the result with <.`
diff --git a/src/core/corelib.c b/src/core/corelib.c
index 5e35bc18..e4549cd6 100644
--- a/src/core/corelib.c
+++ b/src/core/corelib.c
@@ -406,6 +406,26 @@ static Janet janet_core_struct(int32_t argc, Janet *argv) {
     return janet_wrap_struct(janet_struct_end(st));
 }
 
+static int sort_compare_dflt(const void *pl, const void *pr) {
+    return janet_compare(*(Janet*)pl, *(Janet*)pr);
+}
+
+static Janet janet_core_sort(int32_t argc, Janet *argv) {
+    JanetArray *a;
+    Janet by;
+
+    janet_arity(argc, 1, 2);
+    a = janet_getarray(argv, 0);
+    by = argv[1];
+
+    if (argc == 1) {
+        qsort(a->data, a->count, sizeof(Janet), sort_compare_dflt);
+    } else {
+        janet_panic("UNIMPLEMENTED");
+    }
+    return janet_wrap_array(a);
+}
+
 static Janet janet_core_gensym(int32_t argc, Janet *argv) {
     (void) argv;
     janet_fixarity(argc, 0);
@@ -617,6 +637,11 @@ static const JanetReg corelib_cfuns[] = {
              "an odd number of elements, an error will be thrown. Returns the "
              "new struct.")
     },
+    {
+        "sort", janet_core_sort,
+        JDOC("(sort a &opt by)\n\n"
+             "Sort an array in-place. Is not a stable sort.")
+    },
     {
         "gensym", janet_core_gensym,
         JDOC("(gensym)\n\n"
```

felixr:
@andrewchambers  You should rerun this with my latest version of `(sort)`. It is much (10x?) faster than the original version.

For your benchmark I get ~0.5s with C qsort() and ~0.65s with the current sort.

andrewchambers:
That is quite impressive, you may be right - The C version has lots of complications that have not been addressed in my patch either.

bakpakin:
Looks good to me, I'm honestly surprised we do so well without C. Merging.

felixr:
I believe you (accidentally?) merged this change already https://github.com/janet-lang/janet/blob/master/src/boot/boot.janet#L809

felixr:
@bakpakin I just looked at it again and I think it is buggy :scream:  So it might be to good to be true! I would rollback the change

felixr:
Confirmed. I messed something up !! 
`/bin/time build/janet -e '(math/seedrandom 42) (map print (sort (map (fn [x] (math/floor (* 100 (math/random)))) (range 20))))' `
will be nice and sorted (because it is just the insertion sort)
But `/bin/time build/janet -e '(math/seedrandom 42) (map print (sort (map (fn [x] (math/floor (* 100 (math/random)))) (range 40))))' ` is not

Sorry about this! I meant to look at it in more detail, but had not done it.

PR #545, opened 2021-01-03
Revert my buggy hybrid sort

felixr:
This reverts commit 0ea77cabfb30afc15433581f5888171c1f65aafd.

The change was buggy. A `(break)` ended sorting early and left the array mostly unsorted. Without this bug, the code is slower than the pure quicksort. 

See als https://github.com/janet-lang/janet/pull/543

TODO: add test to test suite that tests sorting on longer arrays?

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -795,45 +795,34 @@
     a
     (if (not= (> b a) (> b c)) b c)))
 
-(defn- insertion-sort [a lo hi by]
-  (for i (+ lo 1) (+ hi 1)
-    (def temp (in a i))
-    (var j (- i 1))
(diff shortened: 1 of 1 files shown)


andrewchambers:
Haha, too good to be true :P , it seems we must expand the test suite.

felixr:
Like the test suite I only checked short examples :shamecube: Regarding test suite, maybe something like: 

```
+(assert (deep=
+          (range 99)
+          (sort (flatten ((juxt |(map first $) |(map last $) |(map |(get $ 1) $)) (partition 3 (range 99)))))) "sort 5")
```

bakpakin:
Added sorting tests in 2ca9300bf73f390c73718db175667ef000d57798

PR #546, opened 2021-01-04
Update doc for thread/new and remove ws in loop's

pepe:
I am not sure, why doubling the \n in .c is needed, but without it `(doc thread/new)` prints:

```
repl:1:> (doc thread/new)


    cfunction

    (thread/new func &opt capacity flags)

    Start a new thread that will start immediately. If capacity is
    provided, that is how many messages can be stored in the thread's
    mailbox before blocking senders. The capacity must be between 1 and
    65535 inclusive, and defaults to 10. Can optionally provide flags to
    the new thread - supported flags are: 	:h - Start a heavyweight
    thread. This loads the core environment by default, so may use more
    memory initially. Messages may compress better, though. 	:a - Allow
    sending over registered abstract types to the new thread 	:c - Send
    over cfunction information to the new thread. Returns a handle to the
    new thread.
```

I also changed '\t' to '*' as in `loop` doc. In `loop` doc I shaved some whitespace.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -533,23 +533,16 @@
 
   * :iterate -- repeatedly evaluate and bind to the expression while it is
     truthy.
-
   * :range -- loop over a range. The object should be a two-element tuple with
     a start and end value, and an optional positive step. The range is half
     open, [start, end).
-
   * :range-to -- same as :range, but the range is inclusive [start, end].
-
   * :down -- loop over a range, stepping downwards. The object should be a
     two-element tuple with a start and (exclusive) end value, and an optional
     (positive!) step size.
-
   * :down-to -- same as :down, but the range is inclusive [start, end].
-
   * :keys -- iterate over the keys in a data structure.
-
   * :pairs -- iterate over the key-value pairs as tuples in a data structure.
-
   * :in -- iterate over the values in a data structure or fiber.
 
   `loop` also accepts conditionals to refine the looping further. Conditionals are of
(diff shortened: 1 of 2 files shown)


sogaiu:
Looks nicer for sure.

LGTM!

uvtc:
Note, the 2 spaces of whitespace before the bullet marker in markdown is intentional. While not strictly *required*, it's quite common and gives a nice consistent look to the document, as subsequent lines of the item content get indented 4 spaces to line up vertically. IMO it also helps list items stand out from surrounding text. This also works with numbered lists, and is consistent with 4 space indents for code blocks as well. For example:

~~~
Hi.

  * list content starts at
    four spaces in.

  * 2nd unordered list item

Other content.

 1. this content starts
    at four spaces in as well

        code block inside
        numbered list content

    Here's the 2nd paragraph of
    this numbered list item.

 2. Second numbered list item.

Done.
~~~

pepe:
It is not markdown per se. I see the docs mostly in the terminal, many times 80 chars wide, so less eastward is actually good. 

Finally, every byte ends up in the final image, so I save some precious bytes :-).

uvtc:
Yes, it's not *exactly* markdown, but it's markdown-ish (markdown-ey?) enough so that it's nice and pleasant to read in the terminal (and also helps clarify between prose and code symbols).

The width shouldn't matter much, as the nicely-indented list items should wrap regardless, even if there's lest width than 80 chars, for example:

~~~
...supported flags are:

  * :h - Start a heavyweight thread. This
    loads the core environment by default,
    so may use more memory initially.
    Messages may compress better, though.
  * :a - Allow sending over registered
    abstract types to the new thread
  * :c - Send over cfunction information
    to the new thread. Returns a handle
    to the new thread.
~~~

As for saving every precious byte... personally I don't mind spending a little if it makes the docs easier on the eyes. :)

pepe:
@bakpakin's call I would say :)

uvtc:
Yup, of course. (updated my comment)

sogaiu:
@uvtc I'm not quite following this comment: https://github.com/janet-lang/janet/pull/546#issuecomment-754730196 -- would you mind elaborating on it a bit?

In the original, I see spaces before bullet markers in 2 sections of the docstring.  An example from one section is line 551 and an example from the other section is line 563.  It looked to me like these bullet markers should have the same horizontal positions, but do not.  Is it related to this point?  Or may be it's something else?

uvtc:
@sogaiu , I saw that @pepe made a few different changes in his commit:

  * the first change was removing blank lines in that itemized list
  * the 2nd change was lstripping that whitespace before the list markers
  * and the third was changing the tabs to list markers

I was commenting on the middle one --- left stripping (shaving) off some whitespace. Which I think, really, is just a side-note here.

FWICT, the main issue here that Pepe points out is that the blank lines between list items appears to be necessary, otherwise the doc rendered in the terminal comes out all flowed together, as he's shown. I don't understand why that would be: markdown should support both styles: [compact and loose](https://pandoc.org/MANUAL.html#lists). So I think the doc-formatting machinery in janet should certainly handle both compact and loose lists.

My point of commenting here was: although not central to the main issue, I'm suggesting it's better to use the two spaces before the bullet marker (i.e., always indent list item content a multiple of 4 places), as

  * it makes the plain text docs easier-to-read, and also
  * makes the doc formatting more consistent when writing
      * nested lists,
      * multi-paragraph list items, or
      * list items containing code blocks.

In my experience, following this "4 space rule" avoids lots of markdown formatting errors.

sogaiu:
@uvtc Thanks for the detailed explanation.  I feel less lost now :)

PR #589, opened 2021-01-15
Add support for a profile.janet

yumaikas:
Add support for a profile.janet, along with a flag for disabling it's use in the presence of a JANET_PROFILE env var.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3378,8 +3378,10 @@
           (file/read stdin :line buf))
         (def env (make-env))
         (when-let [profile.janet (dyn :profilepath)] 
-            (def new-env (dofile profile.janet :exit true :env env))
-            (merge-module env new-env "" false))
+            (def src (env :source))
+            (def new-env (dofile profile.janet :exit true :source "profile.janet"))
+            (merge-module env new-env "" false)
+            (put env :source src))
         (if *debug* (put env :debug true))
         (def getter (if *raw-stdin* getstdin getline))
         (defn getchunk [buf p]
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3283,9 +3283,9 @@
   (var *debug* false)
   (var *compile-only* false)
 
-  (when-let [jp (getenv-alias "JANET_PATH")] (setdyn :syspath jp))
-  (when-let [jp (getenv-alias "JANET_HEADERPATH")] (setdyn :headerpath jp))
-  (when-let [jprofile (getenv-alias "JANET_PROFILE")] (setdyn :profilepath jprofile))
+  (if-let [jp (getenv-alias "JANET_PATH")] (setdyn :syspath jp))
+  (if-let [jp (getenv-alias "JANET_HEADERPATH")] (setdyn :headerpath jp))
+  (if-let [jprofile (getenv-alias "JANET_PROFILE")] (setdyn :profilepath jprofile))
   
   # Flag handlers
   (def handlers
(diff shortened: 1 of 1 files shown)


src/boot/boot.janet
```diff
@@ -3283,9 +3283,10 @@
   (var *debug* false)
   (var *compile-only* false)
 
-  (if-let [jp (getenv-alias "JANET_PATH")] (setdyn :syspath jp))
```
pepe, line 3286: May I ask why you changed if-let to when-let? I consider it an antipattern to use when with only one form. Also, I wouldn't say I like changes outside of the PR main domain.

pepe (changes requested):

src/boot/boot.janet
yumaikas: I had chosen to do it that way for newly added form, as a means of communicating I don't think the else branch should be considered. I can change it back if that's considered bad Janet style.

pepe: I actually learned this from @bakpakin's code so that I would say his call. Anyway, I do not find it proper to change things to look the same as the new thing you have added, at least without the note about it.

yumaikas:
So, one problem at play is that this approach doesn't seem to handle errors correctly?

For instance, when I put (error "HEY!") at the end of my profile.janet, that error isn't reported to the REPL. Not sure what the best thing to do there is. I'll scan for error handling once.

src/boot/boot.janet
```diff
@@ -3374,6 +3377,11 @@
           (file/flush stdout)
           (file/read stdin :line buf))
         (def env (make-env))
+        (when-let [profile.janet (dyn :profilepath)] 
+            (def src (env :source))
```
bakpakin, line 3381: No need for this dance, I don't know why this is here. It doesn't do anything since :source should not be set by loading the profile.

src/boot/boot.janet
bakpakin: Doesn't matter IMO, I would accept the change.

src/boot/boot.janet
bakpakin: Also, why are you using `(dyn :profilepath)` instead of just using `jprofile` directly?

yumaikas: Source seems to get set by merge-module, that problem I had with importing things went away when I did this fix.

yumaikas: I'll change things for `jprofile` when I get back to my personal machine.

yumaikas: So, revisiting this, I'm using `(dyn :profilepath)` here because the handler for the `-R` flag sets `(dyn :profilepath)` to null, but doesn't do the same to jprofile. I could change it to change jprofile as well, or get rid of jprofile as a def, but that's why I went down that road.

src/boot/boot.janet
pepe: I was thinking about this a lot, even go through the boot.janet to check `when` vs `if` for one branch stuff and it seems it is not so clear as I saw it originally. So sorry about a bikeshed, definitely should not be the main argument.

Still, I think, you should not change the code, around your change to match your style.

bakpakin:
Ok, so looking at the code as is, this is really not what I would expect - profile is run for each script, when it really should probably only run for the repl. The few languages that I looked at with REPLs do this with the exception of shells, for example TCL and Python.

The code for loading the rc file should probably occur right before the repl is run, not in the loop for handling command line flags.

src/boot/boot.janet
bakpakin: I don't think :source is working how you think it is - `merge-module` does not touch `:source`

yumaikas:
So, it should be in `repl`, not in `cli-main`?

bakpakin:
No, I meant in the block of code at the end of `cli-main`, where we make the call to `repl`. This way, it will only be executed once.

yumaikas:
Isn't it in that block? Doesn't that block start with ` (if (or *should-repl* *no-file*)` on line 3365? And the new call to dofile is on line 3380, guarded by a when-let. I tested it locally with a top-level print statement, and it only printed once when I passed two files to janet.

bakpakin:
Ah, right, nvm. Using the github diff tool confused me into thinking it was in the dohandler.

bakpakin:
Alright, looks good to me then. Using a dyn for :profilepath here should work fine for now.

PR #606, opened 2021-01-21
Replace malloc + memset with calloc.

snimmagadda:
Fixes an overflow warning from gcc with '-Wstringop-overflow' on
NetBSD-current.


bakpakin:
LGTM.

PR #609, opened 2021-01-22
Change tracev to upscope instead of using let

yumaikas:
So, tracev can't be used to wrap defs right now, which feels off (but could have valid reasons).

This added the ability to do that, along with a test of it working as expected.


yumaikas:
@andrewchambers helped me noodle out the code for this yesterday.

andrewchambers:
its maybe an unusual use, but i can't think of a reason why 'do is better in this case.

PR #628, opened 2021-02-05
Switch out `by` to `before?` in sort functions.

yumaikas:
Makes docstrings easier to read, and reduces confusion with sorted-by and sort-by.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -756,7 +756,7 @@
     a
     (if (not= (> b a) (> b c)) b c)))
 
-(defn- sort-help [a lo hi by]
+(defn- sort-help [a lo hi before?]
   (when (< lo hi)
     (def pivot
       (median-of-three (in a hi) (in a lo)
@@ -764,8 +764,8 @@
     (var left lo)
     (var right hi)
     (while true
-      (while (by (in a left) pivot) (++ left))
-      (while (by pivot (in a right)) (-- right))
+      (while (before? (in a left) pivot) (++ left))
+      (while (before? pivot (in a right)) (-- right))
       (when (<= left right)
         (def tmp (in a left))
         (set (a left) (in a right))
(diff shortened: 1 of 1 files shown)


yumaikas:
I'm not sure why that build failed? It built fine for me on my machine, but that is the x64 configuration that works.

yumaikas:
I can reproduce this build error on my laptop, so I'll see if I can figure out what's going on. Seems strange that it's only happening on x86, but it is a marshalling bug. It does happen after a `git clean -fdx`, so it seems unlikely that it's makefile related.

yumaikas:
It seems to pass when running tests as part of `build_win test`, but fails with `build_win all` on my machine?

yumaikas:
So, when I put a trace on test-marsh

```
(defn testmarsh [x msg]
  (def marshx (marshal x m-lookup))
  (def out (marshal (unmarshal marshx um-lookup) m-lookup))

  (when (= msg "marshal table")
    (tracev [(string marshx) (string out) msg]))
  (assert (= (string marshx) (string out)) msg))
```

I get this:

```
trace [test\suite0001.janet] on line 151, column 5: [(string marshx) (string out) msg] is ("\xD3\x02\x01\x02\x03\x04" "\xD3\x02\x03\x04\x01\x02" "marshal table")
```

yumaikas:
When I run this tests on x86 (via `build_win all` or `build_win test`). `d374e90033c2f3506a2f1b50419962b717203f64`, the current janet-lang master branch as of 2021-2-4. I get the same errors. Bisecting now.

yumaikas:
Working with git bisect, I've traced this down to commit [b168b075](https://github.com/janet-lang/janet/commit/b168b075).

Seems like something should be done to fix the test. In any event. I don't think, on it's own, it should be cause for this PR to be blocked. There might obviously be other feedback for it.

subsetpark:
Might be worth adding a line to the doc strings explaining what `before?` is?

bakpakin:
The build error was caused by a regression in how hashing is done and has been fixed. Merging.

PR #639, opened 2021-02-16
Fix or with zero arguments

leahneukirchen:
The value is nil to be consistent for and/or and all/some.
Also add some tests for and/or.


bakpakin:
LGTM

pepe (approved):
LGTM

PR #664, opened 2021-03-12
Fix include path when using meson

### LINKED ISSUE #661: Wrong includedir in .pc file when using meson build
With the Makefile, a pkgconfig is generated with

```
includedir=/usr/include/janet
```

However, the meson build generates

```
includedir=${prefix}/include
```

which breaks `#include <janet.h>`.

leahneukirchen:
Found out how to fix it...

Closes #661.


bakpakin:
LGTM, thanks for the work.

PR #672, opened 2021-03-25
Allow .cc file extension in jpm declare-native

Luewd:
Currently jpm's `declare-native` helper handles compilation of C++ source, but only if it has the `.cpp` extension. However, `.cc` is a common extension for C++ source that is not currently supported by jpm, so it felt prudent to implement this support.

This pull request implements support for the `.cc` extension in `declare-native` with a small bit of refactoring to avoid duplicating the code used to handle the `.cpp` file extension. Additionally, a new case has been added to `test/install` with the `.cc` extension to confirm that jpm correctly builds and installs native modules that have `.cc` source file. The existing test cases were executed successfully, so I believe this change does not introduce any regressions to jpm.

### DIFF
--- a/jpm
+++ b/jpm
@@ -968,17 +968,18 @@ int main(int argc, const char **argv) {
   (var has-cpp false)
   (def objects
     (seq [src :in sources]
-      (cond
-        (string/has-suffix? ".cpp" src)
-        (let [op (out-path src ".cpp" objext)]
-          (compile-cpp opts src op)
-          (set has-cpp true)
-          op)
-        (string/has-suffix? ".c" src)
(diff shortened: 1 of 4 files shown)


sogaiu:
I tried this locally on a project and it seemed to work.

Nice :)

sogaiu:
Hmm, I may have messed up somewhere, but...

I checked out the master branch (83f7de33c0b9c71eb6c0c8c51a9480e51565203a) and copied over the `testmod5.cc` bits and tested there and it seems to work.

May be in my original testing I did something wrong and it just looked like `.cc` wasn't supported.

PR #702, opened 2021-05-21
Add butlast function

skuzzymiglet:
butlast returns an array of all elements except the last. It works on empty arrays. Similar to Clojure's

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -692,7 +692,8 @@
 (defn butlast
   "Return an array containing all but the last element in ind"
   [ind]
-  (if (= (length ind) 0) []
+  (if (= (length ind) 0)
+    @[]
     (array/slice ind 0 -2)))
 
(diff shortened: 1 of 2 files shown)


skuzzymiglet:
Should this function preserve the tuple/array distinction?

subsetpark:
@skuzzymiglet i think as long as it relies on arrays to work, it shouldn’t automatically convert back to tuple for you. If there were some alternative implementations specific to tuples that might be a different story.

andrewchambers:
I think we should make a semi official 'extended/supplemental core' library designed to be imported with 'use. The idea being that most things that want to be in core must first go through there.

bakpakin:
Nothing wrong with his function in general, but not sure it should go into the core (just because Clojure has it in the core doesn't always mean we should too). I agree with Andrew that this would be better as a function in a library.

PR #712, opened 2021-07-11
Tweak function name

sogaiu:
There seems to be a missing `p` in a function name.


pepe (approved):
LGTM

uvtc:
In https://github.com/janet-lang/janet/commit/72c1d1c48461998e478a41c95481627bf2abda69 it looks like all the `panicf`'s were being changed to `panic`'s. It seems like the initial typo was made in changing to `panic`, not `panicf` (since it was *already* initially panicf), no?

PR #713, opened 2021-07-11
Add TerminateProcess to janet_proc_gc and os_proc_kill on Windows

yumaikas:
On windows, nothing was called that would actually terminate a process. I diagnosed this by writing a file-watcher on windows that ran osprey servers.

This should also fix a long-running issue with Janitor


yumaikas:
I can rebase this once master is fixed

bakpakin:
I was on vacation trying to make a quick fix and forgot to double check my work, so I broke master. This should be fixed now, and it looks like this is is a good fix too. I think I was under the impression that closing the handle will terminate the process, which obviously doesn't make sense.

LGTM, thanks for finding this!

Issue #716, opened 2021-07-22
Windows: os/open creates a stream that only accepts the first write?


yumaikas:
I'm on a local build of 1.16.1 on Windows 10, and I'm seeing this:

```janet
(def my-file (os/open "file.txt" :wc))
(:write my-file "First Line\r\n")
(:write my-file "Second Line\r\n")
(:write my-file "Third Line\r\n")
(:close my-file)
```

This results in a `file.txt` that only contains `Third Line\r\n`.

andrewchambers:
First line of the example seems to be missing the os/open call.

yumaikas:
Fixed, and I still see the problem.

sogaiu:
I get a similar result with 54d73f67220e9940702b7d99956b6ee7144a10e7 (for a Windows 10 VM).

yumaikas:
Can we get a `bug` tag on this?

yumaikas:
So, digging into things more, it seems that os/open gives us a core/stream, whose `:write` is bound to `ev/write`. Digging into that, it looks like the differences are that ev_machine_write is called when we're on Windows but not on any other os? Why is this?

PR #717, opened 2021-07-24
Fix os open write windows, and add TerminateProcess calls

yumaikas:
Fixes https://github.com/janet-lang/janet/issues/716.


bakpakin:
Looks good to me. Perhaps we need a flag when creating process that indicates if you want to kill the process on GC or not (say, if you were trying to create an old style daemon).

PR #719, opened 2021-07-24
Clarify docs on take and drop functions

uvtc:
Pass in indexed and bytes, return tuples and strings, respectively.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1027,7 +1027,7 @@
   ret)
 
 (defn take
-  "Take first n elements in an indexed type. Returns new indexed instance."
+  "Take the first n elements of an indexed or bytes type. Returns a new tuple or string, respectively."
   [n ind]
   (def use-str (bytes? ind))
   (def f (if use-str string/slice tuple/slice))
@@ -1038,7 +1038,7 @@
   (f ind 0 end))
 
 (defn take-until
-  "Same as (take-while (complement pred) ind)."
+  "Same as `(take-while (complement pred) ind)`."
   [pred ind]
   (def use-str (bytes? ind))
   (def f (if use-str string/slice tuple/slice))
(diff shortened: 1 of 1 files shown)


sogaiu:
Off the top of my head, I don't know what `indexed type` covers in janet land.

If it includes strings, keywords, and other byte types(?), then it seems like:
```
of an indexed or bytes type
```
is redundant and potentially confusing.

uvtc:
Afaik, indexed types are only tuple and array.

Bytes types are string, buffer, keyword, and symbol.

That said, all of those (above) can be sliced. And the `take` and `drop` (and friends) make use of slicing.

If you use `take`, `drop`, and friends on a string, buffer, keyword, or symbol (that is, something for which `bytes?` returns true), then you get a string back.

If you use `take`, `drop`, and friends on a tuple or array (that is, something for which `indexed?` returns true), then you get a tuple back.

sogaiu:
Along the lines of what is and is not an indexed type, check out `find-index`:
```
(find-index identity "hello")
# => 0
```
Part of its current docstring is:
```
Find the index of indexed type for which pred is true.
```

It's true that the docs for `reduce` say:
```
an indexed type (array, tuple)
```

I'm kind of asking for whether "indexed type" is a pretty well-defined thing and if so what's its definition > @bakpakin 

If it isn't all that well-defined, may be it can either be made that way or perhaps we can avoid using the term altogether.

Or something else :)

uvtc:
I see what you're saying. `find-index` is also supposed to only work on indexed objects, but it works on strings too:

~~~python
(find-index |(= 97 $) "---abc")  #=> 3
~~~

Maybe a new predicate:

~~~clojure
(defn sliceable?
  [x]
  (or (bytes? x)
      (indexed? x)))
~~~

PR #721, opened 2021-07-25
docs for doc, searching for strings

uvtc:


uvtc:
String passed to `doc` is not necessarily a prefix --- `doc` wonderfully shows me all binding names containing that string anywhere in their name.

sogaiu:
Convenient!

Looks like [`string/find` was used](https://github.com/janet-lang/janet/blob/d37c43716aeeea4df108d1e26eccdd5177d75eae/src/boot/boot.janet#L1390) even before [the docstring started to mention a prefix](https://github.com/janet-lang/janet/commit/c335bf5dc57250d8a9c4069c4e4d4afb7fcc7852).

LGTM

Issue #720, opened 2021-07-25
Add Source Mapping information to native Janet Modules


yumaikas:
We're adding support for adding source-map information to Native Janet modules written in C. Soon, @bakpakin will be adding a couple of macros to make that easy.

Here we'll be tracking the effort to port the various parts of the standard library.

- [x] src\core\array.c
- [x] src\core\asm.c
- [x] src\core\buffer.c
- [x] src\core\compile.c
- [x] src\core\corelib.c
- [x] src\core\debug.c
- [x] src\core\ev.c
- [x] src\core\fiber.c
- [x] src\core\inttypes.c
- [x] src\core\io.c
- [x] src\core\marsh.c
- [x] src\core\math.c
- [x] src\core\net.c
- [x] src\core\os.c
- [x] src\core\parse.c
- [x] src\core\peg.c
- [x] src\core\string.c
- [x] src\core\table.c
- [x] src\core\thread.c
- [x] src\core\tuple.c
- [x] src\core\typedarray.c

bakpakin:
Updated `asm` to use the new macros:

Works as follows:

```c
JANET_CORE_FN(cfun_myfun,
   "(my-mod/my-fn x y z)",
   "Does a thing.") {
    janet_fixarity(argc, 3);
    double x = janet_getnumber(argv, 0);
    JanetTable *y = janet_gettable(argv, 1);
    JanetArray *z = janet_getarray(argv, 2);
    ...
    return janet_wrap_array(z);
}

janet_init_some_core_functions(JanetTable *env) {
   JanetRegExt cfuns[] = {
     JANET_CORE_REG("my-mod/my-fn", cfun_myfn),
     JANET_REG_END
   };
   janet_core_cfuns_ext(env, NULL, cfuns);
}
```

bakpakin:
There are also added macros for use outside of the core library that work in a similar way, `JANET_FN` and `JANET_REG`. These can be used in native modules for source mapping.

andrewchambers:
@bakpakin It might be a good time to add arity information to C functions as an option to this macro.

yumaikas:
Just an fyi, @ogdenstudios is planning on helping out, as part of an effort to get familiar with both Janet and C

ogdenstudios:
Excited to help out and learn a bit! Should I hold off until the arity information is added to this macro?

yumaikas:
[...]

PR #726, opened 2021-07-26
Update ev.c to use the new binding style that provides source information

yumaikas:
Update ev.c based on #720

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -735,7 +735,7 @@ JANET_CORE_FN(cfun_channel_push,
 
 JANET_CORE_FN(cfun_channel_pop, 
         "(ev/take channel)",
-        "Write a value to a channel, suspending the current fiber if the channel is full."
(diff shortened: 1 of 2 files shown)


yumaikas:
https://lgtm.com/projects/g/janet-lang/janet/logs/rev/pr-f2aec906aedc1c53032e71c67f37a39b37ca55d3/lang:cpp/stage:Build%20merge_b08a16e375cf0148e24e3d5e1bad87d4116d765c

Looks like the new macro declares things differently than they are in `janet.h`?

pepe:
It looks like `ev/take` doc was copied from `ev/give` in your version

bakpakin:
Missing ev/rselect, which I think is what is causing the LGTM error.

yumaikas:
So, ev/reselect was added, and the errors are still happening.

yumaikas:
LGTM checks are passing now, and this is a closer match to what was being done before

PR #730, opened 2021-07-26
Update math.c for the new style

yumaikas:
Part of the #720 effort

### DIFF
--- a/src/core/math.c
+++ b/src/core/math.c
@@ -117,7 +117,12 @@ double janet_rng_double(JanetRNG *rng) {
     return ldexp((double)(big >> (64 - 52)), -52);
 }
 
-static Janet cfun_rng_make(int32_t argc, Janet *argv) {
+JANET_CORE_FN(cfun_rng_make,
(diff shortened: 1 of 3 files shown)


yumaikas:
I did update some of the docs as I went, making "Returns" used consistently, switching to "the natural logarithm of x", as well as adding small macro for when math constants were defined.

bakpakin:
Looks good. I think we can remove all reference  to the JDOC macro now though, which I will delete after we finish this porting process. The bootstrapping special handling has been moved into the JANET_CORE_* macros.

yumaikas:
So, my C-macro-fu isn't up to the task right now. I'll try to come back to this in a couple days if someone else doesn't get something written up for it, but I want to make sure that we have a macro for defining constants, rather than just defining a macro in math.c that might not follow the bootstrapping process.

bakpakin:
Thanks, the JANET_DEF macros look good as well.

yumaikas:
Fixed the merge conflicts with the ev.c changes

PR #733, opened 2021-07-26
Update io.c with new style core function declarations.

sogaiu:
Part of the https://github.com/janet-lang/janet/issues/720 effort.


sogaiu:
Forgot to mention that there were a couple of docstrings that I tweaked:

* https://github.com/janet-lang/janet/blob/master/src/core/io.c#L714
* https://github.com/janet-lang/janet/blob/master/src/core/io.c#L719

Just added the letter "f" to each name.

PR #740, opened 2021-07-28
Fix macro stringification for one-arg math functions

yumaikas:


sogaiu:
Tried it locally and it seems to work:
```
Janet 1.17.0-dev-b0f97393 linux/x64 - '(doc)' for help
repl:1:> (doc math/cos)


    cfunction
    src/core/math.c on line 268, column 1

    (math/cos x)

    Returns the cosine of x.


nil
```

Issue #751, opened 2021-08-04
`get-in` bug


soulawaker:
```clojure
$ (def t2 {:side false})
{:side false}
$ (get-in t2 [:side :note])
false #; I expected nil
$ (get-in t2 [:side :note] "dflt")
false #; not "dflt"
```

pepe:
It indeed is unexpected behavior caused by `false`. @sogaiu's version is right.

sogaiu:
Not sure I follow what you mean, but IIRC, janet's tables and structs are like Lua's in the sense that:
```
repl:1:> (def t @{:a 1})
@{:a 1}
repl:2:> (put t :a nil)
@{}
repl:3:> t
@{}
```
Or as Lua's [Tables Tutorial](http://lua-users.org/wiki/TablesTutorial) says:
> You can erase a key/value pair from a table by assigning nil to the value.

bakpakin:
Fixed, closing.

PR #752, opened 2021-08-05
Tweak get-in behavior

sogaiu:
An attempt to address https://github.com/janet-lang/janet/issues/751

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1351,7 +1351,7 @@
   a sequence of keys.`
   [ds ks &opt dflt]
   (var d ds)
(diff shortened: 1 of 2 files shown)


pepe (commented):
Just nitpicking but we can use `:until` modifier and condition without `not`.

sogaiu:
I have no attachment to the current implementation in this PR :)

pepe:
It actually does not matter, as the macro expansion of both implementations would probably be very similar.

pepe (approved):
LGTM

bakpakin:
Lgtm

PR #754, opened 2021-08-06
Improve docstring for the invert function.

harryvederci:


uvtc:
You wrote:

> "If multiple keys in the original mapped to the same value, only one of those values will become a key in the returned table.".

I think that sentence is missing an "are". Also, for brevity + clarity you can directly refer to the arg to `invert` by name. What do you think of:

> "if multiple keys in `ds` **are** mapped to the same value ..."

?

sogaiu:
It's a minor thing, but I think:

> if multiple keys in `ds` mapped to the same value

also reads fine to me.  That is, I don't think the lack of "are" is a problem (having it is ok too).

Here are some usages from elsewhere:

> Returns the value mapped to key, not-found or nil if key not present.

via: https://clojuredocs.org/clojure.core/get

> Returns the map with the vals mapped to the keys.

via: https://clojuredocs.org/clojure.set/map-invert

harryvederci:
I actually meant it in the past tense, so `mapped to` implicitly meant `were mapped to`.
Not sure if that makes sense?

Then it would become:
> If multiple keys in `ds` **were** mapped to the same value ...

harryvederci:
Actually, I think the present tense makes more sense, as it doesn't seem to change the input at all.
(Janet newbie here, had to verify it first 😄 )

In that case I agree with @uvtc :
```
If multiple keys in `ds` are mapped to the same value ...
```

PR #756, opened 2021-08-06
Make take/take-while/take-until fiber-aware

subsetpark:
This includes the ability to `take` with either a predicate or number on a fiber, and return an array with the taken values.

It doesn't extend the same behaviour to `drop` because, unlike `take`, dropping values on an infinite stream would return another infinite stream, and collecting those into an array would fail.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1031,30 +1031,60 @@
     (set k (next ind k)))
   ret)
 
-(defn take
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks good to me. I suppose there are work around for drop but the reasoning makes sense.

pepe (approved):
LGTM

PR #761, opened 2021-08-09
Fix init race for environ lock.

andrewchambers:


andrewchambers:
This bug may appear in more places in the code.

andrewchambers:
Actually this

andrewchambers:
Closing because this is insufficient too, this causes a race where an uninitialized critical section may be used.

PR #768, opened 2021-08-21
Remove threads example

sogaiu:
The tail end of `make test` output was being obscured by deprecation messages from the [`threads.janet` example](https://github.com/janet-lang/janet/blob/49142fa385e4bd52327fb53d349e35e8432610b8/examples/threads.janet).  Since there are plans to remove the `thread` module in the  release after next, perhaps there is little harm in removing the example now.


pepe (approved):
I am all for it.

Issue #638, opened 2021-02-16
jpm silently ignores build errors


andrewchambers:
In a shell without go installed `jpm install https://github.com/andrewchambers/poolparty` fails to install poolparty (the failure is expected)... But it seems like janet is ignoring and hiding build errors, the build and also the install has failed, the jpm exit code is still 0.
```
[nix-shell:~]$ jpm install https://github.com/andrewchambers/poolparty
cloning repository https://github.com/andrewchambers/poolparty to /tmp/derp/.cache/https___github.com_andrewchambers_poolparty
Cloning into '/tmp/derp/.cache/https___github.com_andrewchambers_poolparty'...
remote: Enumerating objects: 1142, done.
remote: Counting objects: 100% (1142/1142), done.
remote: Compressing objects: 100% (707/707), done.
remote: Total 1142 (delta 441), reused 1058 (delta 358), pack-reused 0
Receiving objects: 100% (1142/1142), 2.20 MiB | 1.73 MiB/s, done.
Resolving deltas: 100% (441/441), done.
no dependencies found

[nix-shell:~]$ echo $?
0
```

If I manually clone poolparty and run jpm install, then I correctly see the error. os/execute failed to build poolparty
because go was not installed.

```
[nix-shell:~/src/poolparty]$ jpm install
("go" "build" "-mod=vendor" "-o" "./build/poolparty" "./cmd/poolparty/main.go"): No such file or directory
  in os/execute
  in shell [/nix/store/vg73iml2g4ihmcczsc5gljhqbm8jh1kh-janet-1.15.2/bin/jpm] (tailcall) on line 141, column 12
  in do-rule [/nix/store/vg73iml2g4ihmcczsc5gljhqbm8jh1kh-janet-1.15.2/bin/jpm] on line 262, column 26
  in do-rule [/nix/store/vg73iml2g4ihmcczsc5gljhqbm8jh1kh-janet-1.15.2/bin/jpm] on line 258, column 44
  in do-rule [/nix/store/vg73iml2g4ihmcczsc5gljhqbm8jh1kh-janet-1.15.2/bin/jpm] (tailcall) on line 258, column 44
```

pepe:
I can confirm the same behavior here.

sogaiu:
Confirmed here too.

Possibly obvious, but FWIW, this line appears to be executed: 
https://github.com/janet-lang/janet/blob/master/jpm#L170-L171

but this one doesn't appear to be reached:
[...]

PR #785, opened 2021-09-03
Add kqueue support to Janet

### LINKED ISSUE #784: Support kqueue in ev
Was checking in `ev.c` and noticed that EPOLL is supported, but kqueue isn't (POLL is seemingly used for everything but linux). It would be nice if Janet included support for kqueue for the BSD's and possibly Mac OS X (not sure about Mac). I can "read" C code but writing it might be another thing (as in, I'm extremely certain someone could hash this together either quicker, or better, or both, vs me attempting).

llmII:
~~Note that this is in no way ready for merge. I'm sort of seeking guidance/comments as to this whilst I make it ready for merging.~~

~~This seeks to add kqueue support to Janet for *BSD.~~ ~~I already know there are probably logic errors and at least one case of probably should not compile (latter with regards to NetBSD).~~

~~I'm hoping I can have a little help getting this ready for merging especially as there are parts of the Janet code base where I'm pretty sure I've come to wrong assumptions or simply don't know how it works.~~

~~Welcoming all comments, suggestions, and aid.~~

**EDIT**:
Ready for merge (passes all tests).

This adds kqueue support to Janet for *BSD (see notes below). Would fix #784.

*NOTE*:
~~Might need this tested on other BSD's, otherwise if no one can I may want to change things so it only gets compiled in under FreeBSD.~~ ~~In passing I saw a note about NetBSD using a different pointer type for the `udata` field in `struct kevent` but I actually referred to NetBSD's man page here and there and did not note that as documented!~~ ~~This should work under NetBSD which is the only system with a differently defined kevent struct.~~ We borrowed a macro from [wahern/cqueues](https://github.com/wahern/cqueues/blob/master/src/lib/kpoll.c) that casts to the underlying type for the `.udata` field in kevent structs for this purpose. ~~Should work under OpenBSD since the kevent struct is the same as FreeBSD.~~ ~~There may be subtle issues for Mac OS X and I don't believe this code should be active for it unless someone works on it to make sure it works. I've read in passing that Mac OS X has a buggy kqueue implementation.~~ No known issues on MacOS.

*NOTE*:
~~Broke on OpenBSD - NOTE_MSECONDS, NOTE_ABSTIME undefined!~~

Now works on FreeBSD, NetBSD, OpenBSD, and MacOS. Is DragonFly a supported BSD target?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -73,6 +73,7 @@ conf.set('JANET_NO_REALPATH', not get_option('realpath'))
 conf.set('JANET_NO_PROCESSES', not get_option('processes'))
 conf.set('JANET_SIMPLE_GETLINE', get_option('simple_getline'))
 conf.set('JANET_EV_NO_EPOLL', not get_option('epoll'))
+conf.set('JANET_EV_NO_KQUEUE', not get_option('kqueue'))
 conf.set('JANET_NO_THREADS', get_option('threads'))
 conf.set('JANET_NO_INTERPRETER_INTERRUPT', not get_option('interpreter_interrupt'))
 if get_option('os_name') != ''
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -18,6 +18,7 @@ option('umask', type : 'boolean', value : true)
 option('realpath', type : 'boolean', value : true)
 option('simple_getline', type : 'boolean', value : false)
 option('epoll', type : 'boolean', value : false)
+option('kqueue', type : 'boolen', value : false)
 option('interpreter_interrupt', type : 'boolean', value : false)
 
 option('recursion_guard', type : 'integer', min : 10, max : 8000, value : 1024)
--- a/src/conf/janetconf.h
+++ b/src/conf/janetconf.h
@@ -48,6 +48,7 @@
 /* #define JANET_OS_NAME my-custom-os */
 /* #define JANET_ARCH_NAME pdp-8 */
 /* #define JANET_EV_NO_EPOLL */
+/* #define JANET_EV_NO_KQUEUE */
 /* #define JANET_NO_INTERPRETER_INTERRUPT */
 
 /* Custom vm allocator support */
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -1571,13 +1571,9 @@ void janet_ev_deinit(void) {
  */
 
 #elif defined(JANET_EV_KQUEUE)
-/*
- * NOTE: need to touch up janet.h & state.h to have specifics for kqueue.
- */
 
 /* TODO: make this available be we using kqueue or epoll, instead of
- * redefinining it for kqueue and epoll separately.
- */
+ * redefinining it for kqueue and epoll separately? */
 static JanetTimestamp ts_now(void) {
     struct timespec now;
     janet_assert(-1 != clock_gettime(CLOCK_MONOTONIC, &now), "failed to get time");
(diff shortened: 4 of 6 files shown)


llmII:
It now compiles! Fails at test suite 9 though, will work on that.

llmII:
This exits in [suite 9](https://github.com/janet-lang/janet/blob/703753294347f51a0bf3f0f01569d03088122680/test/suite0009.janet#L32) with no message at all, not at all sure why.

llmII:
So... skipped pipes for the moment!

Went straight to the "Net testing" section in Suite 9 - the test-echo thing works!

Copied the code into a `gmake debug` repl and modified it to print the res , no issues.

Pipes are now what's stumping me.

llmII:
I think writes to pipes half work, but it might be a matter of *when* it is written, vs it being writeable - for instance, this is a run with pipe being written (under gdb):

```
repl:4:> (def janet (dyn :executable))
"/usr/home/local/dev/janet/build/janet"
repl:5:>   (let [p (os/spawn [janet "-e" `(print (file/read stdin :line))`] :px {:in :pipe})]
repl:6:(>     (:write (p :in) "hello!\n")
repl:7:(>     (os/proc-wait p)))
[Detaching after vfork from child process 6380]
[New LWP 106492 of process 6374]
hello!

[LWP 106492 of process 6374 exited]
0
repl:7:22: parse error: unexpected delimiter
```

Now the similar code below (again under gdb) fails?:
```
repl:1:> (def janet (dyn :executable))
"/usr/home/local/dev/janet/build/janet"
repl:2:> (def p (os/spawn [janet "-e" `(file/read stdin :line)`] :px {:in :pipe}))
[Detaching after vfork from child process 6349]
<core/process 0x000800A4BB40>
repl:3:> (:write (p :in) "hello!\n")
error: Broken pipe
  in ev/write [src/core/ev.c] on line 2912
  in _thunk [repl] (tailcall) on line 3, column 1
```

And am still not sure why :/

andrewchambers:
Just a note, I think github has a feature where you can mark your pull request as draft.

llmII:
@andrewchambers 
> Just a note, I think github has a feature where you can mark your pull request as draft.

I thought I had marked it that way... seems I missed it thought, thanks!

llmII:
Reading a pipe (pre-close) now works... post close segfaults, maybe progress.

llmII:
All of suite 9 passes now!
Sadly stuck at "os/execute with env" (whatever comes after it) in suite 10.

llmII:
@andrewchambers @bakpakin 
Could you fill me in on how the Regression #638 test in suite 10 works? That's where this hangs and I may just have hit a hurdle where I have no idea what I'm trying to fix. Even looking at the commit that addressed this I'm not sure what I'm looking for...

Going to drop that out of the suite temporarily to try to see if that's the last thing I need to fix to at least pass all tests.

llmII:
All is fixed now, a run of `gmake test 2>&1 | grep "tests passed."` yields the below:
```
%> gmake test 2>&1 | grep "tests passed."                                                            
186 of 186 tests passed.
130 of 130 tests passed.
132 of 132 tests passed.
187 of 187 tests passed.
26 of 26 tests passed.
41 of 41 tests passed.
83 of 83 tests passed.
209 of 209 tests passed.
114 of 114 tests passed.
89 of 89 tests passed.
57 of 57 tests passed.
```

Will mark as ready once I remove some superfluous commentary and notes from the code since I used it to keep track of possibly bad assumptions I made and things that may be wanted or not.

bakpakin:
This is great, I had hoped we would get this at some point! I don't see why this shouldn't work for MacOS, but looking over it right now, it looks ok. In general with porting to different IO apis, there are almost always going to be some edge cases that pop up.

I have an old macbook air I can test on for MacOS, I suspect that the apple implementation of kqueue isn't _that_ bad, and we could work around any issue that arise.

llmII:
OpenBSD lacks NOTE_MSECONDS and NOTE_ABSTIME, will have to manually calculate an interval, vs a specific time. Will see about adding conditional code to detect "I am OpenBSD" and do the right thing.

llmII:
Pushed fixes for OpenBSD support and NetBSD support.

This now works properly on FreeBSD and OpenBSD, need to test NetBSD.

OpenBSD passes all tests `gmake test 2>&1 | grep pass`:
```
openbsd$ gmake test 2>&1 | grep pass
186 of 186 tests passed.
130 of 130 tests passed.
132 of 132 tests passed.
187 of 187 tests passed.
26 of 26 tests passed.
41 of 41 tests passed.
83 of 83 tests passed.
209 of 209 tests passed.
114 of 114 tests passed.
89 of 89 tests passed.
57 of 57 tests passed.
```

llmII:
All passes under OpenBSD and FreeBSD now. NetBSD is passing up to a certain point in suite 9 (haven't figured it out yet) and then dying. I'm beginning to think it can't handle negative timeout intervals (shouldn't really see these I'd think in practice), or 0 valued timeout intervals (maybe need to make JANET_KQUEUE_TS always return a value greater than 0).

llmII:
All fixed under NetBSD and OpenBSD, output remains the same as shown prior for FreeBSD and OpenBSD with regards to tests.

NetBSD test output `gmake test 2>&1 | grep passed` below:
```
localhost: {37} sh
$ gmake test 2>&1 | grep passed
186 of 186 tests passed.
130 of 130 tests passed.
132 of 132 tests passed.
187 of 187 tests passed.
26 of 26 tests passed.
41 of 41 tests passed.
83 of 83 tests passed.
209 of 209 tests passed.
114 of 114 tests passed.
89 of 89 tests passed.
57 of 57 tests passed.
```

bakpakin:
Seems to work when enabled on macos as well - at least passes all tests.

llmII:
Do I need to make sure JANET_EV_KQUEUE is defined under MacOS? Or is it already without further modifications?

bakpakin:
I just added a check for `JANET_APPLE` as well as `JANET_BSD` in the ifdef for  JANET_KQUEUE - (no, apple is not considered a BSD by the JANET_BSD macro).

llmII:
Should be all set now.

PR #789, opened 2021-09-06
Add bind option to net/connect

### LINKED ISSUE #787: Suggestion: support binding address in `net/connect`
Sometimes, a machine has multiple addresses, and being able to select what address is used for outbound traffic programmatically is useful in such scenarios.

An example scenario is like with ZNC where you might want to bind a different host per ZNC user so as to keep from overruning your allowance on an I:Line or to give each a unique vhost.

llmII:
This will allow us to set the address we use for outgoing connections.

**All current tests pass**, works as intended as well.

Fixes #787

### DIFF
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -376,10 +376,12 @@ JANET_CORE_FN(cfun_net_sockaddr,
 }
 
 JANET_CORE_FN(cfun_net_connect,
-              "(net/connect host port &opt type)",
+              "(net/connect host port &opt type bindhost)",
               "Open a connection to communicate with a server. Returns a duplex stream "
               "that can be used to communicate with the server. Type is an optional keyword "
-              "to specify a connection type, either :stream or :datagram. The default is :stream. ") {
+              "to specify a connection type, either :stream or :datagram. The default is :stream. "
+              "Bindhost is an optional string to select from what address to make the outgoing "
+              "connection, with the default being the same as using the OS's preferred address. ") {
     janet_arity(argc, 2, 4);
 
     int socktype = janet_get_sockettype(argv, argc, 2);
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -302,8 +302,7 @@ static struct addrinfo *janet_get_addrinfo(Janet *argv, int32_t offset, int sock
          * specify from where we connect, and in general don't care about a
          * port */
         int32_t current_offset = 3;
-        if (janet_keyeq(argv[current_offset], "stream") ||
-            janet_keyeq(argv[current_offset], "datagram")) {
+        if (janet_checktype(argv[3], JANET_KEYWORD)) {
            current_offset = 4;
         }
         host = (char *)janet_getcstring(argv, current_offset);
@@ -419,12 +418,9 @@ JANET_CORE_FN(cfun_net_connect,
         }
     }
 
-    /* TODO: check if need bind and bind! */
     ai = NULL;
     if (argc >= 3 && is_unix == 0) {
-        if (argc == 4 ||
-            (!janet_keyeq(argv[3], "stream") &&
-             !janet_keyeq(argv[3], "datagram"))) {
+        if (argc == 4 || !janet_checktype(argv[3], JANET_KEYWORD)) {
             ai = janet_get_addrinfo(argv, 0, socktype, 0, &is_unix, 1);
         }
     }
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -304,7 +304,7 @@ static struct addrinfo *janet_get_addrinfo(Janet *argv, int32_t offset, int sock
          * specify from where we connect, and in general don't care about a
          * port */
         int32_t current_offset = 3;
-        if (janet_checktype(argv[3], JANET_KEYWORD)) {
+        if (janet_checktype(argv[2], JANET_KEYWORD)) {
            current_offset = 4;
         }
         host = (char *)janet_getcstring(argv, current_offset);
@@ -422,7 +422,7 @@ JANET_CORE_FN(cfun_net_connect,
 
     ai = NULL;
     if (argc >= 3 && is_unix == 0) {
-        if (argc == 4 || !janet_checktype(argv[3], JANET_KEYWORD)) {
+        if (argc == 4 || !janet_checktype(argv[2], JANET_KEYWORD)) {
             ai = janet_get_addrinfo(argv, 0, socktype, 0, &is_unix, 1);
         }
     }
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -429,23 +429,25 @@ JANET_CORE_FN(cfun_net_connect,
         }
     }
 
-    /* Check all addrinfos in a loop for the first that we can bind to. */
-    struct addrinfo *rp = NULL;
-    for (rp = ai; rp != NULL; rp = rp->ai_next) {
+    if (ai != NULL) {
+        /* Check all addrinfos in a loop for the first that we can bind to. */
+        struct addrinfo *rp = NULL;
+        for (rp = ai; rp != NULL; rp = rp->ai_next) {
 #ifdef JANET_WINDOWS
-        sock = WSASocketW(rp->ai_family, rp->ai_socktype | JSOCKFLAGS, rp->ai_protocol, NULL, 0, WSA_FLAG_OVERLAPPED);
+            sock = WSASocketW(rp->ai_family, rp->ai_socktype | JSOCKFLAGS, rp->ai_protocol, NULL, 0, WSA_FLAG_OVERLAPPED);
 #else
-        sock = socket(rp->ai_family, rp->ai_socktype | JSOCKFLAGS, rp->ai_protocol);
+            sock = socket(rp->ai_family, rp->ai_socktype | JSOCKFLAGS, rp->ai_protocol);
 #endif
-        if (!JSOCKVALID(sock)) continue;
+            if (!JSOCKVALID(sock)) continue;
 
-        /* Bind */
-        if (bind(sock, rp->ai_addr, (int) rp->ai_addrlen) == 0) break;
-        JSOCKCLOSE(sock);
-    }
-    freeaddrinfo(ai);
-    if (NULL == rp) {
-        janet_panic("could not bind outgoing address");
+            /* Bind */
+            if (bind(sock, rp->ai_addr, (int) rp->ai_addrlen) == 0) break;
+            JSOCKCLOSE(sock);
+        }
+        freeaddrinfo(ai);
+        if (NULL == rp) {
+            janet_panic("could not bind outgoing address");
+        }
     }
 
     /* Connect to socket */
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -445,7 +445,7 @@ JANET_CORE_FN(cfun_net_connect,
     }
     freeaddrinfo(ai);
     if (NULL == rp) {
-        janet_panic("could not bind to any sockets");
+        janet_panic("could not bind outgoing address");
     }
 
     /* Connect to socket */
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -420,17 +420,18 @@ JANET_CORE_FN(cfun_net_connect,
         }
     }
 
-    ai = NULL;
+    /* Check if we're binding address */
+    struct addrinfo *binding = NULL;
     if (argc >= 3 && is_unix == 0) {
         if (argc == 4 || !janet_checktype(argv[2], JANET_KEYWORD)) {
-            ai = janet_get_addrinfo(argv, 0, socktype, 0, &is_unix, 1);
+            binding = janet_get_addrinfo(argv, 0, socktype, 0, &is_unix, 1);
         }
     }
 
-    if (ai != NULL) {
+    if (binding != NULL) {
         /* Check all addrinfos in a loop for the first that we can bind to. */
         struct addrinfo *rp = NULL;
-        for (rp = ai; rp != NULL; rp = rp->ai_next) {
+        for (rp = binding; rp != NULL; rp = rp->ai_next) {
 #ifdef JANET_WINDOWS
             sock = WSASocketW(rp->ai_family, rp->ai_socktype | JSOCKFLAGS, rp->ai_protocol, NULL, 0, WSA_FLAG_OVERLAPPED);
 #else
@@ -442,7 +443,7 @@ JANET_CORE_FN(cfun_net_connect,
             if (bind(sock, rp->ai_addr, (int) rp->ai_addrlen) == 0) break;
             JSOCKCLOSE(sock);
         }
-        freeaddrinfo(ai);
+        freeaddrinfo(binding);
         if (NULL == rp) {
             janet_panic("could not bind outgoing address");
         }
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -301,6 +301,7 @@ static struct addrinfo *janet_get_addrinfo(Janet *argv, int32_t offset, int sock
         /* when is_bind is set, we're performing a connect, but wanting to
          * specify from where we connect, and in general don't care about a
          * port */
+        /* TODO: remove check, argv[3] is where our bindhost is at! */
         int32_t current_offset = 2;
         if (janet_checktype(argv[current_offset], JANET_KEYWORD)) {
            current_offset++;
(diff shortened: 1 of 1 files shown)


llmII:
Marking as ready for merge.

To test, configure a secondary loopback address (127.0.0.2 for instance) and run the following (hard to know this for sure works without peername, have to watch netstat/sockstat/whatnot somehow to see who connects from what, maybe I should also add peername support...)

Test code as follows (with assumption of possessing a secondary loopback address):
```
(defn handler
    "Simple handler for connections."
    [stream]
    (defer (:close stream)
      (print "got connection")
      (def id (gensym))
      (def b @"")
      (net/read stream 1024 b)
      (print "read: " b)
      (net/write stream b)
      (buffer/clear b)))

  (def s (net/server "127.0.0.1" "8000" handler))
  (assert s "made server 1")

  (defn test-echo [msg]
    (with [conn (net/connect "127.0.0.1" "8000")]
      (net/write conn msg)
      (def res (net/read conn 1024))
      (assert (= (string res) msg) (string "echo " msg))))

  (defn test-echo-arity-3 [msg address]
    (with [conn (net/connect "127.0.0.1" "8000" address)]
      (net/write conn msg)
      (def res (net/read conn 1024))
      (print res)
      (assert (= (string res) msg) (string "echo " msg))))

  (defn test-echo-arity-4 [msg address]
    (with [conn (net/connect "127.0.0.1" "8000" :stream address)]
      (net/write conn msg)
      (def res (net/read conn 1024))
      (print res)
      (assert (= (string res) msg) (string "echo " msg))))

  (test-echo "hello")
  (test-echo "world")
  (test-echo-arity-3 "hello" "127.0.0.1")
  (test-echo-arity-3 "world" "127.0.0.2")
  (test-echo-arity-4 "hello" "127.0.0.1")
  (test-echo-arity-4 "world" "127.0.0.2")
```

src/core/net.c
```diff
@@ -248,7 +248,9 @@ JANET_NO_RETURN static void janet_sched_accept(JanetStream *stream, JanetFunctio
 /* Adress info */
 
 static int janet_get_sockettype(Janet *argv, int32_t argc, int32_t n) {
-    JanetKeyword stype = janet_optkeyword(argv, argc, n, NULL);
+    JanetKeyword stype = NULL;
+    if(janet_checktype(argv[n], JANET_KEYWORD))
+        stype = janet_optkeyword(argv, argc, n, NULL);
```
bakpakin, line 253: This change doesn't seem to do anything afaik, since the logic is what optkeyword already does. What is the reason for it?

llmII: The issue was that when I use `janet_optkeyword` against a non-keyword type, it gives an error at runtime. My understanding is that `janet_optkeyword` errors if what it tries to get is not either nil or a keyword. The way `(net/connect)` is now written it either `:stream` or a `bindhost` (IP) can be in position 2, and a bind host (string) does not count as a keyword, so `janet_optkeyword` errors out.

bakpakin:
LGTM. Yeah, a wrapper around getpeername could be useful as well.

src/core/net.c
bakpakin: Ok, makes sense. Still, seems like we would have not as good error detection in the non-bind case (say a user passed in a string by accident to `net/listen`.

llmII:
> LGTM. Yeah, a wrapper around getpeername could be useful as well.
See PR #790 ... peername and sockname support.

src/core/net.c
llmII: If they pass in a string say "google.com" or "gobledegook", for the former they'll get the same as the latter (unless they are running at an address that resolves to google.com) which would supposedly be a `janet_panic("could not bind outgoing address)`

For listeners, the arity is still the same as it was before and the call never sets "is_bind" so they supposedly could never see it.

Am I missing an error case for when args > 3  && args <= 4?

llmII: Actually I see what you mean, the error they get would be later along the line as a "cannot bind" error for `(net/connect)` and for listen it should never be anything but an "expected socket type" error. I'm not sure how to fix this and maintain bind support exactly, except maybe to pass down "is_bind" and if so check with `janet_checktype` but otherwise assume and eventually get to the error?

llmII: Thinking further, with how `janet_get_sockettype` is done, depending on argv, I'm unsure it's possible to get it to check and get a socket type in the case where one is binding. It needs to be pointed at a blank argv to assume "not here, do default" and there's no good way to give it indication "not here, go default" without basically moving it up into `(net/connect)` for the things that `(net/connect)` does. Which ultimately gets you the same error, the one in `(net/connect)` instead of the one in `janet_get_sockettype`.  The only thing it can do is check that argv[2] is correctly set when params length is 4, when params length is 3 there's just no way for it to know if socket type or bindhost was meant to be used...

I'll give it more thought, but I'm not sure I have any great ideas on how to keep socket type errors only in `janet_get_sockettype`, so if you or someone else has a good idea I'm all ears.

bakpakin: I think there is some misunderstanding here - get_sockettype is just a simple function to read `:stream`, `:datagram`, or `nil` from the arglist. By removing the check, you are allowing errors through in other functions that use it.

bakpakin: For example, `(net/listen "127.0.0.1" "8000" @"blahblahblahblah")` should throw an error - with this change, it will continue as normal.

llmII: You're correct, and it sucks, but I'm gonna fix that here in a bit. Basically, gonna move that check out of the socket type getter and into net/connect (gonna try to collect all address infos at the head of the function in net/connect with a strict socket type check when argc == 4 and a lax (caller checks) check when argc == 3). The result will be that net/connect could error in socket type, or it can error in a bind failure.

pepe: I guess we are used to using `nil` for the unused argument. For example in `fiber/new` when you are not providing the resume value, but you provide a supervisor channel

llmII: Pushed fix, keeps other `net/*` from being effected by changes to `net/connect`. The issue here is that `net/connect` can still have an issue like how you mentioned with `net/listen` (except that unless the user managed to pick something hostname or ip address like, they'd get a bind error rather than an error about calling the function with the wrong types).

The only way to fix that would be to break the `net/connect` API call and have it take keyword arguments for all optional arguments instead of variadic limited arity arguments. If you have any ideas on how to detect nil in position 3 or 4 when the argc can be 3 or 4 let me know but without a language change (sparse argv) or API change I'm not really sure how to make `net/connect` for sure give an argument error for not using keyword when it should be a keyword instead of a bind error.

I could leave a note on the function to "break API and use keyword arguments in $version" where I would guess version would be 2.0.0 (I don't think breaking api changes are wanted for the second number (like 0.X.0).

The final idea I have is perhaps to create a strict 4 argument `net/connect_impl` and then create functions in the boot.janet that wrap around it that support the old way (like how the code is now) and a new way later on down the road (I don't think janet allows a function to have multiple definitions dependent upon argument types where it'd need to at runtime pick a call to one that takes optionals and another that takes keywords excepting if we use a function intermediate to dispatch to the correctly defined underlying function).

sogaiu: > if one can create a function with arguments like [& xs] where it can be called like it's using keyword arguments...

Note sure I follow, but do you mean like this?
```
(defn make-recipe-3
  "Create some kind of recipe using &keys..."
  [&keys {:flour flour
          :sugar sugar
          :baking-soda soda
          :water water
          :eggs eggs
          :vanilla-extract vanilla}]
  (def dry [flour sugar soda])
  (def wet [water eggs vanilla])
  {:name "underspecified-cake" :wet wet :dry dry})

# Calling this function is a bit different now - no struct
(make-recipe-3
  :flour 1
  :sugar 1
  :baking-soda 0.5
  :water 2
  :eggs 2
  :vanilla-extract 0.5)
```
via: https://janet-lang.org/docs/functions.html#Keyword-style-arguments

llmII: @sogaiu 

Say we have a cfunction `net/connect_impl (host port sockettype bind)` (no opt args)

Now say we create in boot.janet a dispatcher that will use that:

`(net/connect [host port [& xs]])`

Can I use xs like keyword arguments? The idea is to allow a call like:
`(net/connect [host port &opt sockettype bind]` 
and at the same time allow 
`(net/connect [host port [&keys {:type type :bind bind}]])`
which would dispatch to `net/connect_impl`

Which the first one defined (`(net/connect [host port [& xs]])`) would dispatch to depending on what sorts of args it got.

sogaiu: I don't think so, but am not sure.

May be @bakpakin or @andrewchambers know better.

llmII:
Do CI services let you create loopback addresses? If so... could do something like

psuedocode:
```
(match [get-os]
  :linux   (setup-loopback-ips-linux)
  :bsd     (setup-loopback-ips-bsd)
  :mac     (setup-loopback-ips-mac)
  :windows (setup-loopback-ips-windows))
(test-bind)
```

In the test suites... if not, no idea of a good way to automate testing for this honestly.

src/core/net.c
llmII: Just realized, but are keyword arguments even optional? I mean

```
(net/connect host port :type :stream :bind "127.0.0.1")
# should be the same as below
(net/connect host port :bind "127.0.0.1")
# right?
```

Which brings me to thinking it'd be nice to have keys that have a default set in the function itself... such that you could do something like:

```
(net/connect [host port [&keys {:type &default stream :bind &default ""}]])
```

Tho I guess in all reality... a really well done macro could possibly do all the hard work there. Though I don't think a macro could tell the difference between "they meant to set socket type with :stream but gave a string" and "they actually meant it as an IP so it's ok" either...

bakpakin:
The argument handling in connect still looks very wrong to me, I don't understand why we need a switch statement on the argument count, and the strange handling as is.

> The final idea I have is perhaps to create a strict 4 argument net/connect_impl and then create functions in the boot.janet that wrap around it that support the old way (like how the code is now) and a new way later on down the road (I don't think janet allows a function to have multiple definitions dependent upon argument types where it'd need to at runtime pick a call to one that takes optionals and another that takes keywords excepting if we use a function intermediate to dispatch to the correctly defined underlying function)

Yes, that sounds much better. We don't support the latter by design, don't try and implement it in C. Really, narrowing a function call like `net/connect` from 4 arguments to 3 in some cases just doesn't make sense, don't bother.

llmII:
> The argument handling in connect still looks very wrong to me, I don't understand why we need a switch statement on the argument count, and the strange handling as is.

The argument handling was basically borne from trying to keep `(net/connect [host port &opt type])` interface the same as before (prevent breaking code that uses the 2 arg call or 3 arg call) while adding another argument to determine the host address to bind the socket to. Basically, having to handle things with the switch (which I did because it looked better than a few if statements)  is more a maintain backwards compatibility thing and I thought all arguments after `&opt` were optional and shifted left on absence of a value.

> > The final idea I have is perhaps to create a strict 4 argument net/connect_impl and then create functions in the boot.janet that wrap around it that support the old way (like how the code is now) and a new way later on down the road (I don't think janet allows a function to have multiple definitions dependent upon argument types where it'd need to at runtime pick a call to one that takes optionals and another that takes keywords excepting if we use a function intermediate to dispatch to the correctly defined underlying function)
> 
> Yes, that sounds much better, although just use `net/connect` as the 4 argument function. We don't support the latter by design, don't try and implement it in C. Really, narrowing a function call like `net/connect` from 4 arguments to 3 in some cases just doesn't make sense, don't bother.

You mean change `net/connect` to be `net/connect_impl` and change the options handling over to wrapper functions in `boot.janet`? Or change it to a 4 arg call (breaking compatibility with all code)? or?

I do think it's hard for a user to try `"stream"` instead of `:stream` and expect things to work for instance since a string literal isn't treated like a keyword at all (at least by the `janet_optkeyword` call.

As to not supported by design (dispatch based on argument types), that's fine, and I'm a long way away from trying to implement a new language feature, vs library features. The language has enough features that if one is willing to box types into an abstract type they can achieve a similar effect (possibly at performance overhead *shrugs*).

If you let me know what you want it to be, exactly, I can probably do it. I've sort of been around it a few times trying to maintain the old interface, but ultimately it's up to you all if the old interface is discarded or retained, and I just need a function prototype definition and I'll make it fit (probably).

bakpakin:
There is no "left shift" going on - I don't know what you are talking about, simply adding handling for `argv[3]` as bindhost as an optional string will work fine and not have any side effects nor break existing code.

`(defn x [a b &opt c d] ...)` means that _both_ c and d are optional arguments. If you wanted pass in bindhost but not type you would call `(net/connect "127.0.0.1" "8000" nil "127.0.0.2")` and that's it. The 2 argument version and 3 argument version will still work fine.

llmII:
> There is no "left shift" going on - I don't know what you are talking about, simply adding handling for `argv[3]` as bindhost as an optional string will work fine and not have any side effects nor break existing code.

What I meant is, if a function is defined to have args `[a b c &opt d e f]` it has a minimum of 3 arguments, and you can place what values you want but if you don't wish to define d, but want to define e and f, then d must be nil (or e and f get shifted left in the argument list to d and e).

> `(defn x [a b &opt c d] ...)` means that _both_ c and d are optional arguments. If you wanted pass in bindhost but not type you would call `(net/connect "127.0.0.1" "8000" nil "127.0.0.2")` and that's it. The 2 argument version and 3 argument version will still work fine.

I see, I've been trying to make `&opt` act way differently than is intended to keep from having to write `nil` for a 3rd argument. I was busy making things work where it was like it was 3 different functions: 
`(net/connect host port type)`, `(net/connect host port bind)`, `(net/connect host port type bind)`.

I'll make it work like that, with a nil required and only arg 4 being possible for binds.

Sorry bout my nearsightedness there, couldn't see the forest for the trees!

bakpakin:
To elaborate more, I would just try something like this:


```
/* Inside cfun_net_connect */
if (argc > 3 && !janet_checktype(argv[3], JANET_NIL)) {
  int is_bindhost_unix = 0;
  struct addrinfo *ainfo = janet_get_addrinfo(argv, 3, stype, 0, &is_bindhost_unix, 1);
  /* Do binding here */
}

```

llmII:
Passes under FreeBSD, NetBSD, ~~OpenBSD~~, ~~Linux~~ with the following output (note, crossed out items don't mean it doesn't work, just means not yet tested):
```
<--
        remote view: @["127.0.0.1" "65533" "127.0.0.1" "8000"]
        local view: ("127.0.0.1" 8000 "127.0.0.1" 65533)
-->
        remote view: @["127.0.0.1" "8000" "127.0.0.1" "65533"]
        local view: ("127.0.0.1" 65533 "127.0.0.1" 8000)
<--
        remote view: @["127.0.0.1" "65532" "127.0.0.1" "8000"]
        local view: ("127.0.0.1" 8000 "127.0.0.1" 65532)
-->
        remote view: @["127.0.0.1" "8000" "127.0.0.1" "65532"]
        local view: ("127.0.0.1" 65532 "127.0.0.1" 8000)
<--
        remote view: @["127.0.0.2" "65531" "127.0.0.1" "8000"]
        local view: ("127.0.0.1" 8000 "127.0.0.2" 65531)
-->
        remote view: @["127.0.0.1" "8000" "127.0.0.2" "65531"]
        local view: ("127.0.0.2" 65531 "127.0.0.1" 8000)
<--
        remote view: @["127.0.0.2" "65530" "127.0.0.1" "8000"]
        local view: ("127.0.0.1" 8000 "127.0.0.2" 65530)
-->
        remote view: @["127.0.0.1" "8000" "127.0.0.2" "65530"]
        local view: ("127.0.0.2" 65530 "127.0.0.1" 8000)
```

Using code that merges the [llmII:feature-bind-connect](https://github.com/llmII/janet/tree/feature-bind-connect) and [llmII:feature-getsockname-getpeername](https://github.com/llmII/janet/tree/feature-getsockname-getpeername) branches so as to illustrate it.

Code used for testing is:
```
(defn check-matching-names [stream &opt direction]
    "Checks that the remote agrees with the local about ip/port"
    (let [[my-ip my-port]         (net/localname stream)
          [remote-ip remote-port] (net/peername stream)
          to-write                  (string/join 
                                       @[my-ip (string my-port)
                                         remote-ip (string remote-port)]
                                                 " ")
          buffer @""]
      (if (= direction :write) 
        (do (net/write stream to-write) (net/read stream 1024 buffer))
        (do (net/read stream 1024 buffer) (net/write stream to-write)))
      (def comparison (string/split " " buffer))
      (printf "%s\n\t%s%q\n\t%s%q" (if (= direction :write) "-->" "<--")
                                   "remote view: "
                                   comparison
                                   "local view: "
                                   [my-ip my-port remote-ip remote-port])
      (assert (and (= my-ip (get comparison 2))
                   (= (string my-port) (get comparison 3))
                   (= remote-ip (get comparison 0))
                   (= (string remote-port) (get comparison 1))))))
(defn handler
    "Simple handler for connections." 
    [stream]
    (defer (:close stream)
      (check-matching-names stream)))
    
(def s (net/server "127.0.0.1" "8000" handler))
(assert s "made server 1")

(defn test-echo [&opt bind]
  (with [conn (net/connect "127.0.0.1" "8000" :stream bind)]
    (check-matching-names conn :write)))
```

src/core/net.c
llmII: Fixed to work as @bakpakin specified in [#789](https://github.com/janet-lang/janet/pull/789#issuecomment-914571833)

bakpakin:
LGTM

PR #800, opened 2021-09-12
add NULL check in gc.c to avoid UB

marler8997:
After the UB was fixed in value.c, I tried running the build again and encoutered another instance of UB in gc.c.  With this fixed I'm now able to build janet with ubsan enabled, meaning there's no more UB encountered in janet_boot during the build.


bakpakin:
Thanks, this looks pretty cut and dry.

PR #821, opened 2021-09-28
Fix #819 timeouts under kqueue on FreeBSD broken

### LINKED ISSUE #819: kqueue ABSTIME under FreeBSD (ev/sleep) dislikes timers that expire before being mapped into the kqueue.
What I believe is happening, is that ABSTIME is in milliseconds... and 0 milliseconds (or slightly more) being mapped in as $now+$interval in absolute time has already occured by time the kqueue has the event mapped into it, and it gives an invalid argument error to ev.

Basically, every `(ev/sleep 0)` fails.

The solution... only ever use intervals with kqueue. It means a subtraction operation occurs every timeout that is mapped into the kqueue. The only place where this would remain slightly special is NetBSD where it's invalid for there to be an interval below 1 millisecond.

I'll probably work on this some tomorrow, I have a fix, just it is really a simple hack at the moment and causes compiler warnings. The fix would need to be tested against NetBSD, FreeBSD, and OpenBSD (and depend on CI for the MacOS test).

llmII:
FreeBSD is the only BSD supporting ABSTIME timers, whereas the rest demand intervals. Janet operates on timestamps,  which are absolute times, as per ABSTIME. The idea was to use that under FreeBSD but not the other BSDs. This commit changes that since ABSTIME breaks when the timeout supplied is for a time prior to whatever the time is now (invalid argument). We now utilize the same logic we use on the other BSDs with FreeBSD to effect interval timeouts since intervals are absolutely sometime beyond now, be it now and less than a millisecond, or more than a millisecond. This will hopefully unbreak BSD builds when running the test suite.

Tested under FreeBSD, NetBSD, and OpenBSD - relying on CI for MacOS.

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -1576,15 +1576,7 @@ void janet_ev_deinit(void) {
  * NetBSD uses intptr_t while others use void * for .udata */
 #define EV_SETx(ev, a, b, c, d, e, f) EV_SET((ev), (a), (b), (c), (d), (e), ((__typeof__((ev)->udata))(f)))
 #define JANET_KQUEUE_TF (EV_ADD | EV_ENABLE | EV_CLEAR | EV_ONESHOT)
-
-/* NOTE:
- * NetBSD doesn't like intervals less than 1 millisecond so lets make that the
- * default anywhere JANET_KQUEUE_TS will be used. */
-#ifdef __NetBSD__
-#define JANET_KQUEUE_MIN_INTERVAL 1
-#else
 #define JANET_KQUEUE_MIN_INTERVAL 0
-#endif
 
 /* NOTE:
  * NetBSD and OpenBSD expect things are always intervals, and FreeBSD doesn't
@@ -1593,10 +1585,10 @@ void janet_ev_deinit(void) {
  * kqueue. Also note that NetBSD doesn't accept timeout intervals less than 1
  * millisecond, so correct all intervals on that platform to be at least 1
  * millisecond.*/
-JanetTimestamp fix_interval(const JanetTimestamp ts) {
+JanetTimestamp to_interval(const JanetTimestamp ts) {
     return ts >= JANET_KQUEUE_MIN_INTERVAL ? ts : JANET_KQUEUE_MIN_INTERVAL;
 }
-#define JANET_KQUEUE_TS(timestamp) (fix_interval((timestamp - ts_now())))
+#define JANET_KQUEUE_INTERVAL(timestamp) (to_interval((timestamp - ts_now())))
 
 
 /* TODO: make this available be we using kqueue or epoll, instead of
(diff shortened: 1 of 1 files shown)


bakpakin:
Should we ad some `(ev/sleep 0)` to the test suite just to be sure?

llmII:
We could, there are a few `(ev/sleep 0)`s already in the test suite actually (previous breakage points). Under all the BSD's I just ran `(for i 0 100 (print i) (ev/sleep 0))` successfully. Executes quick (near instant) on FreeBSD and within a second on NetBSD (enforced 1 millisecond timeout) and OpenBSD (might be internally forcing a 1 millisecond minimum at the OS level perhaps).

I'm unsure of how early on in the test suites it should run, or which test suite it should run in though.

bakpakin:
Hmm, `(ev/sleep 0)` should always execute immediately, i.e. not actually wait. I would consider anything else to be incorrect.

llmII:
> Hmm, `(ev/sleep 0)` should always execute immediately, i.e. not actually wait. I would consider anything else to be incorrect.

That's one where I don't know for sure we can get any more correct with regards to NetBSD insisting timeouts be 1 millisecond or more, at least with the timer filter.

> Looking a the kqueue code some more, I think we can avoid this whole issue by skipping the timer altogether and using the timeout argument to kevent. Since you are resetting the timer with EV_SETx every call to kevent, I'm not sure it's even more efficient.
> 
> EDIT: I'm fairly sure this would also solve the waiting issue, since I would expect kevent to return immediately if given a 0 timeout and no events are present.

This might work. The reason for timers being done that way was modeling after the epoll code and making sure I didn't skip any steps in case there was some non-obvious necessity to them. It really should work, and I know OpenBSD and FreeBSD support 0 on their timers so they should support 0 in timeout. The only issue will be if NetBSD treats timeouts like it does timers with a minimum wait of 1ms or more. Returning to "execute immediately", if the OS's underlying implementation is either slow enough for perceptible time passage or internally treats 0 as 1ms, I'm unsure of what we could do there. Same with if the minimum (to keep from having invalid argument error) is 1ms. I'll try this out the way you describe later today and test across the BSDs.

The only thing I can think of that might need some thought is if the timers are used to map fibers to sleep sequences. Say you have 10 fibers and each sleeps 2 more seconds than the other, first sleeping 0 seconds, where does Janet figure out the fiber to run based on interval of time passed? I'd have to look a bit more deeply but if a timing wheel or something of the sorts is being used I guess this is worry over nothing most likely.

bakpakin:
We should be good for using timeout, that is how the original poll code worked anyway. As for epoll using a timer fd, that is because on Linux we can update the timer efficiently without messing with the epoll descriptor. I don't know that it is more efficient, but perhaps slightly and I think should be slightly more accurate.

An I would be very surprised if netBSD didn't allow for a zero timeout to kevent, but maybe that is naive.

I wouldn't worry about extend timing accuracy here, since 1 millisecond accuracy is already so far from accurate that I can only see this improving. Using timestamps internally and then converting to intervals only at the last moment also would let us more easily implement a call like `(ev/sleep-to timestamp)` in the future.

llmII:
Latest commit changes it to how you described @bakpakin - still need it to run in CI to check against MacOS and probably will need to run it through the formatter as well as some minor cleanup after this, but it worked (and faster) under all the BSDs.

Hopefully this will enable the idea of `(ev/sleep-to timestamp)`. Let me know if there's anything else you'd like to see in the kqueue fix before it's considered for merge.

I did also test the for 100 sleeps thing again as well.

bakpakin:
LGTM, although the macos timing out issue seems related to a small change I made for peername/localname testing, will investigate.

PR #823, opened 2021-09-30
Fix #822 - kqueue hang in suite 9.

### LINKED ISSUE #822: BSD + MacOS hang
The a slight change in suite 9 for peername localname seems to have exposed a hang using kqueue on BSDs in CI, both on GitHub and in Sourcehut - see 504411eade2860d3a51482ea4eb2dda1e7999317

I expect that something is happening here with handling concurrent reads and writes on the same file descriptor - not sure if this is an issue with the kqueue implementation, or with BSD in general.

To summarize:
- We need to verify if this happens when janet is built without kqueue support
- If kqueue is the issue, identify and fix our implementation. If BSD is the issue, figure out what events we are missing or skipping causing a concurrent read and write on the same stream to hang.

llmII:
This fixes #822 

Priorly we only checked exactly one state when an event was received. This was incorrect. A state may have a next state, an action to take after the first in the list of states has been taken. This change acknowledges that and makes the code work with the state list vs just the head of the list.

Tested under FreeBSD, NetBSD, and OpenBSD with success. MacOS is tested by CI and if that flakes we'll need someone to try it on a Mac device.

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -1709,7 +1709,8 @@ void janet_loop1_impl(int has_timeout, JanetTimestamp timeout) {
         } else {
             JanetStream *stream = p;
             JanetListenerState *state = stream->state;
-            if (NULL != state) {
+            while (NULL != state) {
+                JanetListenerState *next_state = state->_next;
                 state->event = events + i;
                 JanetAsyncStatus statuses[4];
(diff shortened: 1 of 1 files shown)


bakpakin:
Lgtm, glad the fix was something clean like this.

PR #827, opened 2021-10-06
Tweak format strings

sogaiu:
A couple of format strings were using `%s` where may be `%V` would be better.

With fbe06782, I get:
```
$ janet
Janet 1.18.0-dev-fbe06782 linux/x64 - '(doc)' for help
repl:1:> (def s (net/server "127.0.0.1" "0"))
<core/stream 0x563B07CEB270>
repl:2:> (net/peername s)
Segmentation fault (core dumped)
```


pepe (approved):
LGTM

PR #840, opened 2021-10-15
Support Android (termux)

svercl:
This change allows compiling and running on Android in termux.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,6 +30,7 @@ header_path = join_paths(get_option('prefix'), get_option('includedir'), 'janet'
 cc = meson.get_compiler('c')
 m_dep = cc.find_library('m', required : false)
 dl_dep = cc.find_library('dl', required : false)
+android_spawn_dep = cc.find_library('android-spawn', required : false)
 thread_dep = dependency('threads')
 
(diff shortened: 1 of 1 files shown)


sogaiu:
It built here and all tests passed.

bradms:
I added it to the meson build as well. Tested on my device. One thing that needs to be changed is that the dep should only be added when `JANET_REDUCED_OS` is not defined.

sogaiu:
After installing `ninja` and `meson` (using `pip3`) I tried out the modifications in 559fd70 with some success :)

sogaiu:
I also had success via the `Makefile` route by installing the clang, binutils, make, git, and libandroid-spawn packages first.

svercl:
Yes! I assumed those packages are installed already, but I think this should be mentioned for running in Termux.

sogaiu:
Not sure where that info ought to go (may be in the [janet-lang.org repository](https://github.com/janet-lang/janet-lang.org)?).

For future folks, a copy-paste one-liner to install aforementioned packages might be:

```pkg install binutils clang git libandroid-spawn make```

PR #868, opened 2021-11-12
Allow C code to block SIGINT.

andrewchambers:
Previously the repl always exits on SIGINT, this change
means the repl will only exit on SIGINT if the SIGINT handler
causes it to exit.


bakpakin:
Ah thanks, there were some commits a while ago that tried to turn the Janet repl into something more "shell-like" including having Ctrl-C clear the console.

PR #880, opened 2021-11-24
Fix math/gamma and add math/log-gamma

pepe:
I am working on the statistical library for Janet, and I found a typo for the gamma function in the `math/`. This PR corrects the typo and adds `log-gamma` function pointing to the correct function.

This PR also adds tests for these two functions and could be the place where we can test the `math` functions.


bakpakin:
Looks good to me.

PR #902, opened 2021-12-25
Tweak match docstring

sogaiu:
Adds a bit of text to mention that bracket tuples are handled like arrays in `match`.


pepe (approved):
LGTM

PR #905, opened 2022-01-07
Fix typo in janet_epoll_sync_callback, add test for "async" read from normal fd

ishehadeh:
Fixed a typo in janet_epoll_sync_callback  where `JANET_ASYNC_LISTEN_WRITE` was checked instead of `JANET_ASYNC_LISTEN_READ`.
It caused the runtime to hang when reading from a normal file descriptor, like one from `os/open`.

This PR also introduces a test for `ev/read`, its almost identical to the test for writing to a normal fd.
The only difference is it reads the content back with `:read` instead of `slurp`.


bakpakin:
Looks good, I think there still needs to be some cleanup on macos and windows for this functionality here based on the CI. But epoll is now working correctly.

PR #913, opened 2022-01-20
Support for `& rest` pattern in destructure and match

ishehadeh:
This PR adds support for the `& rest` pattern in destructure and the `match` macro. The pattern can be used inside an array or tuple. It binds any remaining input to `rest`, if everything preceding it matched.

Examples:
```
(match [:a :b :c]
  [:a & rest] rest)
# => (:b :c)
```

```
(def [x y & r] [1 2])
# x = 1, y = 2, r = ()
```

#912

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1678,6 +1678,7 @@
 
   * array or bracket tuple -- an array or bracket tuple will match only if
     all of its elements match the corresponding elements in `x`.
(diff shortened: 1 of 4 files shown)


pepe:
This PR already looks excellent to me. I hope this will get into lang.

bakpakin:
Looks pretty good to me. Thanks for the PR, @ishehadeh!

PR #915, opened 2022-01-21
Add ppc64 to os/arch

AlbertoGP:
Same as #431 (Add ppc to os/arch) but for the 64-bit version.
This is tested on a Power9 CPU in Little-Endian mode, on Linux.


bakpakin:
LGTM

PR #923, opened 2022-02-20
add `(int/to-number)`: convert int/s64 and int/u64 to a number

ishehadeh:
Currently, there's no way to get the core number back from an `int/` 64-bit integer.

This PR adds a function `(int/to-number)` which converts either an `int/s64` or `int/u64` to a number.
It fails if the input will not fit in a 32-bit int, and uses `janet_wrap_integer` internally.

### DIFF
--- a/src/core/inttypes.c
+++ b/src/core/inttypes.c
@@ -216,23 +216,22 @@ JANET_CORE_FN(cfun_to_number,
 
         if (janet_abstract_type(abst) == &janet_s64_type) {
             int64_t value = *((int64_t*)abst);
-            if (value > INT32_MAX) {
-                janet_panicf("cannot convert %q to a number, exceededs int32 max", argv[0]);
+            if (value > JANET_INTMAX_INT64) {
+                janet_panicf("cannot convert %q to a number, must be in the range [%q, %q]", argv[0], janet_wrap_number(JANET_INTMIN_DOUBLE), janet_wrap_number(JANET_INTMAX_DOUBLE));
             }
-            if (value < INT32_MIN) {
-                janet_panicf("cannot convert %q to a number, exceededs int32 min", argv[0]);
+            if (value < -JANET_INTMAX_INT64) {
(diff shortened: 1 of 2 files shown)


andrewchambers:
Technically you can support much larger ranges than INT_MAX, iirc the largest number is something in the range of 52 bits, but you will need to check this.

paulsnar:
> Technically you can support much larger ranges than INT32_MAX, iirc the largest and smallest number is something in the range of 53 bits, but you will need to check this.

Base precedent for this is briefly documented, i.e., in JavaScript, which also shares the implementation detail of all numbers being IEEE 754 doubles. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

ishehadeh:
I chose to limit it to the int32 range because it seems like that's what Janet considers an "integer" (i.e. `int?` returns true for it).
I mostly did that just because it seems like *really* odd behavior if you convert a `int/u64` -> `number` then assert `int?` and it fails.

bakpakin:
Agree with Andrew here - upping the integer limit to 53 bits will be much more useful.

The `int?` function is a bit of a relic but it is better to think of `int` referirng to the C data type, not `integer`. The docstring makes this explicit.

ishehadeh:
Makes sense, I've updated to restrict the range to JANET_INTMIN/JANET_INTMAX instead.

ishehadeh:
By the way - `janet_wrap_integer` casts the  input to an `int32_t`. Instead, it might be better if it cast the input to an `int64_t`, checked it against `JANET_INTMIN/JANET_INTMAX` then called `janet_wrap_number`. Is there any reason for the current implementation?

bakpakin:
LGTM

> By the way - janet_wrap_integer casts the input to an int32_t. Instead, it might be better if it cast the input to an int64_t, checked it against JANET_INTMIN/JANET_INTMAX then called janet_wrap_number. Is there any reason for the current implementation?

We cast to int32_t since that is the most commonly used type internally. Using 32 bit integers internally also ensures that whatever integer value we have can be represented by a double - the same is not true for 64 bit integers.

PR #924, opened 2022-02-23
Update docs for `describe`

### LINKED ISSUE #524: `describe` used by string, symbol, keyword, buffer? Docs correct?

uvtc:
should close #524

### DIFF
--- a/src/core/corelib.c
+++ b/src/core/corelib.c
@@ -339,7 +339,10 @@ JANET_CORE_FN(janet_core_native,
 
 JANET_CORE_FN(janet_core_describe,
               "(describe x)",
-              "Returns a string that is a human-readable description of a value x.") {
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM. We could omit the phrase "more or less unique", though.

PR #929, opened 2022-03-02
made profile.janet override default dyns such as :pretty-format

saikyun:
I want to be able to change `:pretty-format` when using the repl.

I realize there are multiple ways that this can be done, so this is just a suggestion. Main problem with this implementation is the question "should command line args or profile.janet have precedence?".

In this case I chose to let profile.janet win, but if it should be the other way around, let me know.


pepe:
I would definitely vote for command line args precedence over profile.janet.

bakpakin:
Command line arguments are generally expected to override configs. Doing otherwise seem very strange to me

PR #930, opened 2022-03-05
Normalize number

saikyun:
Sorry if this is completely unnecessary, but I found a way to normalize the negative float without branching. I don't know a good way to profile this sort of change, so feel free to just disregard it.

If you have tips on how to benchmark this sort of change, I'd be happy to try it.


saikyun:
Whoops, this PR was messed up, it included other changes too.

PR #931, opened 2022-03-05
normalize zero without branching

saikyun:
Sorry if this is completely unnecessary, but I found a way to normalize the negative float without branching (https://stackoverflow.com/questions/9657993/how-to-convert-negative-zero-to-positive-zero-in-c/44707354#44707354).
I don't know a good way to profile this sort of change, so feel free to just disregard the PR.

If you have tips on how to benchmark this sort of change, I'd be happy to try it.

### DIFF
--- a/src/core/value.c
+++ b/src/core/value.c
@@ -322,7 +322,7 @@ int32_t janet_hash(Janet x) {
                 uint64_t u;
             } as;
             as.d = janet_unwrap_number(x);
-            as.d = as.d == 0 ? 0 : as.d; /* normalize negative 0 */
+            as.d += 0.0; /* normalize negative 0 */
             uint32_t lo = (uint32_t)(as.u & 0xFFFFFFFF);
             uint32_t hi = (uint32_t)(as.u >> 32);
             uint32_t hilo = (hi ^ lo) * 2654435769u;


bakpakin:
Not sure that this is that useful, or even that this is guaranteed to convert -0.0 to 0.0 by any C standard. For this kind of thing I would wonder if the assembly was any different.

saikyun:
Ah, okay. Yeah, I don't really have any clue, and not enough knowledge to
answer your questions. ^^; So we can close the PR.

Den sön 6 mars 2022 20:15Calvin Rose ***@***.***> skrev:

GrayJack:
I did try to turn that single line into functions in [godbolt](https://godbolt.org/z/enM5M7qYr) to see if the operation itself made any difference in the instructions produced. It seems it did, but I had to say the double number is volatile to avoid the compiler enabling optimization to no-op because the functions were not used and I'm not sure if that changes more things than I wanted

bakpakin:
godbolt results look convincing, will merge.

pepe:
OT, but thank you @GrayJack for mentioning godbolt. It is very nice tool, I have not been aware of.

herrgahr:
FWIW @GrayJack I think this variation (having the fcts return the normalized value) might be more informative since volatile does add some noise to the codegen: https://godbolt.org/z/Ebfeejcq7

PR #932, opened 2022-03-05
Add int/to-bytes: Serialize int/[su]64 to a buffer

ishehadeh:
Add `int/to-bytes` which writes the bytes of an `int/` type to a buffer. 
Optionally takes a byte order and existing buffer to write to.


pepe (approved):
LGTM

PR #936, opened 2022-03-17
Refactor __MACH__ to JANET_APPLE

cellularmitosis:
# Summary:

This PR does two things:
- Change some uses of `__MACH__` to `JANET_APPLE`
- Change `JANET_APPLE` to check for both `__APPLE__` as well as `__MACH__`

# Details

Heyo Janet folk,

I came across an implementation of `clock_gettime` for older OS X machines which uses `mach_absolute_time`, but before I get into that...

I had noticed that sometimes `__MACH__` is used and sometimes `JANET_APPLE` is used.  This PR consolidates to using `JANET_APPLE` across the board (I hope that's what was intended by `JANET_APPLE`!).

Also, I realized I didn't actually understand the difference between `__MACH__` vs `__APPLE__`, so I dug a bit and it turns out `__APPLE__` is also defined on System 9, etc, so it looks like the most correct thing is actually to use both.

Apple seems to indicate as much here: https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html#//apple_ref/doc/uid/TP40002850-SW13

> To define a section of code to be compiled on OS X system, you should define a section using \_\_APPLE\_\_ with \_\_MACH\_\_ macros.

also this xcode-users mailing list post: https://lists.apple.com/archives/xcode-users/2006/Nov/msg00356.html

> many people seem to use: #if defined(\_\_APPLE\_\_) && defined(\_\_MACH\_\_)

> \_\_APPLE\_\_ is defined for builds on classic too, so may not limit the conditional enough.

as well as various stackoverflow posts, e.g. https://stackoverflow.com/a/26225829

> #if defined(\_\_APPLE\_\_) && defined(\_\_MACH\_\_)

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -68,7 +68,7 @@ extern char **environ;
 #endif
 
 /* For macos */
-#ifdef __MACH__
+#ifdef JANET_APPLE
 #include <mach/clock.h>
 #include <mach/mach.h>
 #endif
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -790,7 +790,7 @@ int32_t janet_sorted_keys(const JanetKV *dict, int32_t cap, int32_t *index_buffe
 /* Clock shims for various platforms */
 #ifdef JANET_GETTIME
 /* For macos */
-#ifdef __MACH__
+#ifdef JANET_APPLE
 #include <mach/clock.h>
 #include <mach/mach.h>
 #endif
@@ -806,7 +806,7 @@ int janet_gettime(struct timespec *spec) {
     spec->tv_nsec = wintime % 10000000LL * 100;
     return 0;
 }
-#elif defined(__MACH__)
+#elif defined(JANET_APPLE)
 int janet_gettime(struct timespec *spec) {
     clock_serv_t cclock;
     mach_timespec_t mts;
--- a/src/include/janet.h
+++ b/src/include/janet.h
@@ -57,8 +57,8 @@ extern "C" {
 #define JANET_BSD 1
 #endif
 
-/* Check for Mac */
-#ifdef __APPLE__
+/* Check for macOS or OS X */
+#if defined(__APPLE__) && defined(__MACH__)
 #define JANET_APPLE 1
 #endif
 


cellularmitosis:
Oh, also, I built and ran the tests, but then I realized that `janet.c` is amalgamated, and I don't know how to re-amalgamate it, so actually I don't think I was testing my changes...

sogaiu:
I think the amalgamated `janet.c` (what comes into existence under `build/c`) will [disappear via `make clean`](https://github.com/janet-lang/janet/blob/c815185574f88f8b0c43a334ccdda84669bb11e7/Makefile#L326-L327).  IIUC, a subsequent `make` should ["amalgamate" / create another one](https://github.com/janet-lang/janet/blob/c815185574f88f8b0c43a334ccdda84669bb11e7/Makefile#L160-L161) (again under `build/c`).

cellularmitosis:
ah thanks @sogaiu , looks like that worked:

```
$ grep -r __MACH__ *
build/janet.h:#if defined(__APPLE__) && defined(__MACH__)
src/include/janet.h:#if defined(__APPLE__) && defined(__MACH__)
```

and the tests pass:

```
$ make test
for f in test/suite*.janet; do ./build/janet "$f" || exit; done
Starting suite 0...
Finished suite 0 in 0.003 seconds - 190 of 190 tests passed.
Starting suite 1...
Finished suite 1 in 0.005 seconds - 153 of 153 tests passed.
Starting suite 2...
Finished suite 2 in 0.001 seconds - 132 of 132 tests passed.
Starting suite 3...
Finished suite 3 in 0.004 seconds - 187 of 187 tests passed.
Starting suite 4...
Finished suite 4 in 0.000 seconds - 26 of 26 tests passed.
Starting suite 5...
Finished suite 5 in 0.000 seconds - 41 of 41 tests passed.
Starting suite 6...
Finished suite 6 in 0.001 seconds - 104 of 104 tests passed.
Starting suite 7...
Finished suite 7 in 0.013 seconds - 210 of 210 tests passed.
Starting suite 8...
Finished suite 8 in 0.002 seconds - 122 of 122 tests passed.
Starting suite 9...
Finished suite 9 in 0.253 seconds - 890 of 890 tests passed.
Starting suite 10...
Finished suite 10 in 0.002 seconds - 77 of 77 tests passed.
Starting suite 11...
Finished suite 11 in 0.000 seconds - 6 of 6 tests passed.
for f in examples/*.janet; do ./build/janet -k "$f"; done
```

PR #937, opened 2022-03-22
Mac clock shim not needed after 10.12

cellularmitosis:
# Summary

This PR:
- changes the Mac clock shim to only be used for Macs older than 10.12.
- moves the `AvailabilityMacros.h` include from `os.c` to `util.c`
- removes the unused `mach/clock.h` and `mach/mach.h` includes from `os.c`

# Details

In `util.c`, the Mac clock shim was included for all Macs:

```
if defined(JANET_APPLE)
```

However, `clock_gettime()` was added in 10.12, so the clock shim isn't needed from that point forward: 
- https://opensource.apple.com/releases/
- https://opensource.apple.com/source/Libc/Libc-1158.1.2/gen/clock_gettime.c.auto.html

This PR changes the preprocessor logic to only use the clock shim for Macs running OS X older than 10.12:

```
/* clock_gettime() wasn't available on Mac until 10.12. */
#elif defined(JANET_APPLE) && !defined(MAC_OS_X_VERSION_10_12)
```

Update: I also noticed `AvailabilityMacros.h` was included in `os.c`, but is only used in `util.c`, so I moved the include.

Update2: Also noticed that `mach/clock.h` and `mach/mach.h` were included in `os.c`, but are only needed for the shim.  I'd guess that the clock shim stuff used to live in `os.c` before being pulled out into `util.c`.

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -39,10 +39,6 @@
 #include <sys/stat.h>
 #include <signal.h>
 
-#ifdef JANET_APPLE
-#include <AvailabilityMacros.h>
-#endif
-
 #ifdef JANET_WINDOWS
 #include <windows.h>
 #include <direct.h>
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -63,12 +63,6 @@ extern char **environ;
 #endif
 #endif
 
-/* For macos */
-#ifdef JANET_APPLE
-#include <mach/clock.h>
-#include <mach/mach.h>
-#endif
-
 /* Not POSIX, but all Unixes but Solaris have this function. */
 #if defined(JANET_POSIX) && !defined(__sun)
 time_t timegm(struct tm *tm);
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -36,6 +36,10 @@
 #endif
 #endif
 
+#ifdef JANET_APPLE
+#include <AvailabilityMacros.h>
+#endif
+
 #include <inttypes.h>
 
 /* Base 64 lookup table for digits */
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -789,11 +789,6 @@ int32_t janet_sorted_keys(const JanetKV *dict, int32_t cap, int32_t *index_buffe
 
 /* Clock shims for various platforms */
 #ifdef JANET_GETTIME
-/* For macos */
-#ifdef JANET_APPLE
-#include <mach/clock.h>
-#include <mach/mach.h>
-#endif
 #ifdef JANET_WINDOWS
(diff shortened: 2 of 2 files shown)


cellularmitosis:
It looks like `clock_gettime` is a lot faster than the shim on my M1 running BigSur 11.6.

`bench1.c`:

```c
#include <time.h>
#include <stdio.h>

int janet_gettime(struct timespec *spec) {
    return clock_gettime(CLOCK_REALTIME, spec);
}

int main(int argc, char** argv) {
    struct timespec spec;
    int i;
    for (i = 0; i < 100000000; i += 1) {
        janet_gettime(&spec);
    }
    printf("%ld\n", spec.tv_sec);
    return 0;
}
```

`bench2.c`:

```c
#include <time.h>
#include <stdio.h>

#include <mach/clock.h>
#include <mach/mach.h>
int janet_gettime(struct timespec *spec) {
    clock_serv_t cclock;
    mach_timespec_t mts;
    host_get_clock_service(mach_host_self(), CALENDAR_CLOCK, &cclock);
    clock_get_time(cclock, &mts);
    mach_port_deallocate(mach_task_self(), cclock);
    spec->tv_sec = mts.tv_sec;
    spec->tv_nsec = mts.tv_nsec;
    return 0;
}

int main(int argc, char** argv) {
    struct timespec spec;
    int i;
    for (i = 0; i < 1000000; i += 1) {
        janet_gettime(&spec);
    }
    printf("%ld\n", spec.tv_sec);
    return 0;
}
```

```
cell@indium$ gcc -Wall -o bench1 bench1.c && time ./bench1 
1647914319

real	0m1.385s
user	0m1.328s
sys	0m0.002s

cell@indium$ gcc -Wall -o bench2 bench2.c && time ./bench2 
1647914324

real	0m1.536s
user	0m0.279s
sys	0m1.206s
```

So I can call `clock_gettime` 100 million times in about the same runtime as calling the shim 1 million times.

(I also tried `-O2` which didn't make any difference).

cellularmitosis:
Looks like the tests failed on windows:

```
Starting suite 9...
✘ "localname should match peername: msg=\"127.0.0.1 57922 127.0.0.1 8000\", buf=@\"57922\"": false
✘ "localname should match peername: msg=\"8000\", buf=@\"57923\"": false
✘ "localname should match peername: msg=\"57923\", buf=@\"8000\"": false
✘ "localname should match peername: msg=\"8000\", buf=@\"57924\"": false
✘ "localname should match peername: msg=\"57924\", buf=@\"8000\"": false
```

Hmm, that doesn't feel related to the PR.  Is this just a CI burp?

PR #942, opened 2022-03-30
Correct version 1.21.1 in meson.build

Techcable:
This causes incorrect version in meson compiles (including homebrew)

In the future you may want to consider using the `version: file` argument.


pepe (approved):
LGTM

PR #943, opened 2022-03-31
Produce dylibs on macOS

cellularmitosis:
This PR changes the Makefile to produce dylibs on macOS.

It appears to work:

Screen Shot 2022-03-31 at 12 13 20 AM

And it appears that Linux is unaffected:

Screen Shot 2022-03-31 at 12 17 11 AM

(heh, I just noticed the SONAME in the makefile needs to be bumped from 1.20 to 1.21)

This is the first time I've needed a conditional inside of a _target_ of a makefile.  I based my solution on the technique presented here https://stackoverflow.com/questions/15977796/if-conditions-in-a-makefile-inside-a-target but if there are better ways, I'm open!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ UNAME:=$(shell uname -s)
 ifeq ($(UNAME), Darwin)
 	CLIBS:=$(CLIBS) -ldl
 	SONAME_SETTER:=-Wl,-install_name,
+	JANET_LIBRARY=build/libjanet.dylib
 	LDCONFIG:=true
 else ifeq ($(UNAME), Linux)
 	CLIBS:=$(CLIBS) -lrt -ldl
@@ -165,7 +166,11 @@ build/c/janet.c: build/janet_boot src/boot/boot.janet
 ##### Amalgamation #####
 ########################
 
+ifeq ($(UNAME), Darwin)
+SONAME=libjanet.1.20.dylib
+else
 SONAME=libjanet.so.1.20
+endif
 
 build/c/shell.c: src/mainclient/shell.c
 	cp $< $@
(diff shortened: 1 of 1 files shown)


cellularmitosis:
Hmm, actually it looks like I should also include `-compatibility_version` and `-current_version` flags.

LuaJIT is similar to janet in that it does no use autotools to generate a Makefile.  See https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/Makefile#L324-L326

This is also a useful guide: https://begriffs.com/posts/2021-07-04-shared-libraries.html

pepe:
> (heh, I just noticed the SONAME in the makefile needs to be bumped from 1.20 to 1.21)

I guess we could PR this separately?

cellularmitosis:
I noticed the trick of using `$(shell $(JANET_TARGET) -e '(print janet/version)')` in the Makefile to pull in the version information after the janet binary had been built.

I went down the rabbit hole of trying to use this to get rid of all hard-coded version info in the Makefile:

Screen Shot 2022-04-01 at 9 04 55 PM

but I couldn't get it to work.  It turns out certain things are hard to do in make:
- delay top-level expressions until after a certain rule is run (e.g. until after the janet binary has been built)
- use conditionals and define top-level make vars inside of a target

oh well.

bakpakin:
Is there a good reason for this? As far as I can tell for Janet's use case they should have very similar usage

cellularmitosis:
> Is there a good reason for this?

Are you asking if there is a good reason for producing `.dylib`'s rather than `.so`'s on macOS?  My understanding is that you can't link against a `.so` on macOS (though I have actually never tried it).

Oh, I just noticed that homebrew already installs dylibs:

```
$ ls -l /opt/homebrew/Cellar/janet/1.21.1/lib/
total 1284
drwxr-xr-x 3 cell admin     96 Mar 27 11:21 janet
-rw-r--r-- 1 cell admin 592096 Mar 31 22:02 libjanet.1.21.dylib
-r--r--r-- 1 cell admin 717976 Mar 27 11:21 libjanet.a
lrwxr-xr-x 1 cell admin     19 Mar 27 11:21 libjanet.dylib -> libjanet.1.21.dylib
drwxr-xr-x 3 cell admin     96 Mar 31 22:02 pkgconfig
```

I guess it isn't using the Makefile, or is patching it?

Hmm, it looks like they are using ninja: https://github.com/Homebrew/homebrew-core/blob/master/Formula/janet.rb

If building dylibs from the Makefile isn't desired I can just close this PR.

bakpakin:
On further reading this seems like a good idea so I'll merge it.

PR #949, opened 2022-04-13
Use defdyn for syspath in boot.janet

sogaiu:
Part of an effort to address https://github.com/janet-lang/janet-lang.org/issues/83.


pepe (approved):
LGTM

PR #963, opened 2022-04-18
table.c docstrings

uvtc:
Added some backticks around code in docstrings to distinguish them from prose.

Light editing to `table/raw-get`. Is my change there correct? (t --> the key)


sogaiu:
LGTM

PR #966, opened 2022-04-19
boot.janet docstrings (part 2)

uvtc:
Added some backticks around code in docstrings to distinguish them from prose.

Also some light editing.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -206,7 +206,7 @@
 
 (defmacro case
   ``Select the body that equals the dispatch value. When `pairs`
-  is an odd number of arguments, the last is the default expression.
+  has an odd number of elements, the last is the default expression.
   If no match is found, returns nil.``
   [dispatch & pairs]
   (def atm (idempotent? dispatch))
@@ -1147,7 +1147,7 @@
     (tuple/slice ret 0)))
 
 (defmacro juxt
-  "Macro form of juxt*. Same behavior but more efficient."
+  "Macro form of `juxt*`. Same behavior but more efficient."
   [& funs]
   (def parts @['tuple])
   (def $args (gensym))
@@ -1193,8 +1193,8 @@
   "Bound to the name of the currently compiling file.")
 
 (defmacro tracev
-  `Print a value and a description of the form that produced that value to
-  stderr. Evaluates to x.`
+  `Print to stderr a value and a description of the form that produced that value.
+  Evaluates to x.`
   [x]
   (def [l c] (tuple/sourcemap (dyn *macro-form* ())))
   (def cf (dyn *current-file*))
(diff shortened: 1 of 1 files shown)


src/boot/boot.janet
```diff
@@ -1193,8 +1193,8 @@
   "Bound to the name of the currently compiling file.")
 
 (defmacro tracev
-  `Print a value and a description of the form that produced that value to
-  stderr. Evaluates to x.`
+  `Print to stderr a value and a description of the form that produced that value.
+  Evaluates to x.`
```
pepe, line 1197: I guess x could be in `` too?

src/boot/boot.janet
```diff
@@ -1235,11 +1235,11 @@
   (reduce fop x forms))
 
 (defmacro -?>
-  `Short circuit threading macro. Inserts x as the second value in the first form
-  in forms, and inserts the modified first form into the second form
+  ``Short circuit threading macro. Inserts x as the second value in the first form
```
pepe, line 1238: Again, x could be in ``?

src/boot/boot.janet
```diff
@@ -1251,11 +1251,11 @@
   (reduce fop x forms))
 
 (defmacro -?>>
-  `Short circuit threading macro. Inserts x as the last value in the first form
-  in forms, and inserts the modified first form into the second form
+  ``Short circuit threading macro. Inserts x as the last value in the first form
```
pepe, line 1254: x in ``?

src/boot/boot.janet
```diff
@@ -1300,14 +1300,14 @@
...
+  "Similar to `postwalk`, but do pre-order traversal."
   [f form]
   (walk (fn [x] (prewalk f x)) (f form)))
 
 (defmacro as->
-  `Thread forms together, replacing as in forms with the value
-  of the previous form. The first for is the value x. Returns the
-  last value.`
+  ``Thread forms together, replacing `as` in `forms` with the value
+  of the previous form. The first form is the value x. Returns the
```
pepe, line 1309: x in ``?

src/boot/boot.janet
```diff
@@ -1317,10 +1317,10 @@
   prev)
 
 (defmacro as?->
-  `Thread forms together, replacing as in forms with the value
-  of the previous form. The first for is the value x. If any
+  ``Thread forms together, replacing `as` in `forms` with the value
+  of the previous form. The first form is the value x. If any
```
pepe, line 1321: x in ``?

src/boot/boot.janet
```diff
@@ -1365,17 +1365,16 @@
...
   [ind]
   (var res true)
   (loop [x :in ind :while res]
     (if x nil (set res x)))
   res)
 
 (defn any?
-  `Returns the first truthy value in ind, otherwise nil.
-  falsey value.`
+  ``Returns the first truthy value in `ind`, otherwise nil.``
```
pepe, line 1377: Shouldn't nil be too in ``?

pepe (changes requested):
I am not sure, if `x`s make sense?

uvtc:
"x" shows up so often, and is so self-explanatory, that I usually don't put it into backticks (unless it's right next to some other parameter that was put into backticks, and it would look weird if the x were *not* in backticks).

I try to have a light touch here and not be slavishly consistent in cases where the docstring is already clear as-is.

My primary goal is to clarify the docstring prose, especially in cases where a parameter name or other function name (code) can be mistaken for prose.

pepe (approved):
LGTM

pepe:
Thank you for the explanation, it makes sense.

PR #991, opened 2022-07-01
Trace function to the stderr

pepe:
After the last debugging debate, I started using the `trace` function. It works fine, yet it prints to the stdout, which is not ideal when doing webdev.

This commit changes it and traces the function call to stderr, similarly to `tracev`.


pepe:
For posterity, I have started using trace by default when debugging.

PR #993, opened 2022-07-02
Add basic test for tabseq

pepe:
This commit adds a new suite with the `tabseq` tests. For the second one, I thought it would be cool to lint empty `value-body`, but that would require a significant reorganization of the `boot.janet` code. I still think it will be worth doing, so we can start to use this functionality (macro linting) in the core more (yep, a rhyme).


bakpakin:
LGTM - I guess empty value body would be syntactically correct but just return an empty table, so linting it would be fine.

pepe:
That will need to move `maclintf` way up in the `boot.janet`. If you think it is worth pursuing, I will volunteer.

PR #994, opened 2022-07-02
Use relative path for include/janet.h symlink

shassard:
When using make to build an rpm, the current symlink is
created with an absolute path to the buildroot as causes
the make install target to fail with:

```bash
error: Symlink points to BuildRoot: /usr/include/janet.h -> /home/stephen/rpmbuild/BUILDROOT/janet-1.23.0-3.x86_64/usr/include/janet/janet.h
```

We can create the link relatively which makes this more
portable, where:

```bash
ln -sf -t '/home/stephen/rpmbuild/BUILDROOT/janet-1.23.0-3.x86_64/usr/include' janet.h janet/janet.h
```

Resulting in the following symlink:

```bash
ls -la BUILDROOT/usr/include/janet.h
lrwxrwxrwx. 1 stephen stephen 13 Jul  2 08:17 BUILDROOT/usr/include/janet.h -> janet/janet.h
```

This symlink can then be properly packaged without path
issues.


bakpakin:
Interesting, good catch.

PR #1003, opened 2022-07-21
Add string/contains? method

jgoday:
Implements string/contains? method.

It returns true if a substring is included in a particular string.
Similar to [clojure.string/includes?](https://clojuredocs.org/clojure.string/includes_q).

Internally it uses c **strstr** function ([reference](https://en.cppreference.com/w/c/string/byte/strstr))

Added some basic test to suite0001.janet.


bakpakin:
This will fail to work as it should on Janet strings with embedded 0s. Also, "string/find" will do the same thing.

PR #1006, opened 2022-08-04
Change all occurrences of (= nil x) to (nil? x)

pepe:
This commit is a tentative change to use `nil?` instead of equality with nil. This change may have some consequences I am not aware of, even if I am aware that this change is just cosmetic.


bakpakin:
(= nil x) is optimized by the VM and is slightly faster, so I won't merge this.

pepe:
That was my hunch. And good to know! Yet it leads me to another thought what if nil? would be a macro then?

Thanks

PR #1012, opened 2022-08-18
Clarify the documentation of janet -E flag

Techcable:
This confused me, despite having a fair deal of janet experience.

Also I find the man-page roff format fairly counterintuitive. Maybe consider switching from raw man pages to something like [scdoc](https://git.sr.ht/~sircmpwn/scdoc)? Here is an [example file](https://git.sr.ht/~emersion/hut/tree/master/item/doc/hut.1.scd) and the [format docs](https://man.archlinux.org/man/scdoc.5.en).

### DIFF
--- a/janet.1
+++ b/janet.1
@@ -164,9 +164,14 @@ Execute a string of Janet source. Source code is executed in the order it is enc
 arguments are executed before later ones.
 
 .TP
-.BR \-E\ code arguments
+.BR \-E\ code\ arguments...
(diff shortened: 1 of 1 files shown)


bakpakin:
Thanks for the example, that does clear it up somewhat. That said, we are sticking to man pages. The man pages aren't meant to be the the main form of documentation, just something I think any command line tool should have.

As far as manpage doc generators, I think we could hack together a ~200 line Janet one that was a lot more expressive and easy to use than scdoc, and bundle it with Janet source, but not really on my list of things to do.

Issue #1018, opened 2022-08-30
UndefinedBehaviorSanitizer doesn't like several bugs (including "applying zero offset to null pointer")


Techcable:
UndefinedBehaviorSanitizer is unhappy about "applying zero offset to null pointer".

Presumably this comes up when empty strings are involved.

This seems to be an issue in two files:  [parse.c:1197](https://github.com/janet-lang/janet/blob/03ba1f7021e8beefb62afb29e3c7a17cf373596f/src/core/parse.c#L1197) in `parser_state_delimiters ` and [utils.c:119](https://github.com/janet-lang/janet/blob/03ba1f7021e8beefb62afb29e3c7a17cf373596f/src/core/util.c#L119) in `check_str_const`

````
Janet 1.25.0-dev-8102eb2d macos/aarch64 - '(doc)' for help
src/core/parse.c:1197:31: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/core/parse.c:1197:31 in 
src/core/util.c:119:30: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/core/util.c:119:30 in 
````

I found this trying to debug my own code.

**Comment by Techcable** [comment]
Caught this in a couple other places running the test suite:

```
src/core/specials.c:207:88: runtime error: left shift of negative value -5
src/core/pp.c:639:63: runtime error: applying zero offset to null pointer
src/core/inttypes.c:491:1: runtime error: signed integer overflow: 9223372036854775807 + 1 cannot be represented in type 'long long'
src/core/parse.c:1208:27: runtime error: applying zero offset to null pointer
src/core/parse.c:1212:18: runtime error: applying zero offset to null pointer
src/core/util.c:119:30: runtime error: applying zero offset to null pointer
````

**Comment by Techcable** [comment]
I don't believe the "applying zero offset to null pointer" bug is very serious.

The one that does look interesting (besides integer overflow) is:
```
src/core/specials.c:207:88: runtime error: left shift of negative value -5
```

The code is [here](https://github.com/janet-lang/janet/blob/03ba1f7021e8beefb62afb29e3c7a17cf373596f/src/core/specials.c#L207).

According to Python, `-5 << 8` is `-1280` (same answer as `-5 * 256`)

Not sure why this would give you an issue except that you're using `|=` between a `uint32_t` and a `int32_t` (so it's probably just being pedantic)

[...]

PR #1019, opened 2022-08-30
Signed integer overflow is undefined behavior in C, avoid it in inttypes.c

Techcable:
In C, signed arithmetic overflow is undefined behvior
but unsigned arithmetic overflow is twos complement

Unconditionally switch to unsigned arithmetic internally for +, -, *
This will not affect the result thanks to twos complement awesomeness.

I don't think this will be an issue in these functions,
but it has a history of causing bugs.....

I think this is probably the most important part of #1018 (`NULL + 0` is probably not going to miscompile....)

### DIFF
--- a/src/core/inttypes.c
+++ b/src/core/inttypes.c
@@ -407,13 +407,26 @@ static Janet cfun_it_u64_compare(int32_t argc, Janet *argv) {
     return janet_wrap_nil();
 }
 
+/*
+ * In C, signed arithmetic overflow is undefined behvior
+ * but unsigned arithmetic overflow is twos complement
+ *
+ * Reference:
+ * https://en.cppreference.com/w/cpp/language/ub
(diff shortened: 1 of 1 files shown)


andrewchambers:
I noticed a flag JANET_NANBOX_32 and it made me wonder if this change works when 32 bit nan boxing is enabled?

Techcable:
> I noticed a flag JANET_NANBOX_32 and it made me wonder if this change works when 32 bit nan boxing is enabled?

Right now `int/s64` is implemented as a regular abstract type provided by an extension module. It just happens to be a core module but doesn't really get special status. It shouldn't be affected by NaN boxing.

Looking t `rg 'janet_u64|\.u64' -t c`, it appears that when `JANET_NANBOX_32` is set, the `Janet.u64` field of the union is only used for hashing & booleans.

andrewchambers:
Thanks, I totally misunderstood the change, for some reason in my mind you were updating the unwrap macros.

PR #1023, opened 2022-09-09
Set JANET_MANPATH environment variable while installing.

ScriptDevil:
JPM on windows currently installs its manpage to C:\ directly because this isn't set when installing Janet through the MSI installer


ScriptDevil:
When doing either `jpm install jpm` or `janet boostrap.jpm`, I noticed that the manpage gets installed directly to C:\ on windows. Further inspection made it clear that JPM expects janet to have set this environment variable beforehand.

```
copying jpm/cgen.janet to C:\Users\ajadatharan\AppData\Local\Apps\Janet\Library\/jpm...
jpm\cgen.janet
1 File(s) copied
copying jpm.1 to /usr/local/share/man/man1...
C:jpm.1
```

PR #1028, opened 2022-09-12
Made peg 'not' and 'if-not' drop their captures on success

### LINKED ISSUE #1026: Peg pattern `if-not` does not drop condition's captures properly
I would expect the following code to return just an empty array:

```janet
(peg/match '(if-not (* (constant 7) "a") "hello") "hello")
```

However it returns `@[7]`, despite the spec saying that the condition will not produce any captures.

autumnull:
fixes #1026

### DIFF
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -261,24 +261,46 @@ tail:
             goto tail;
         }
 
-        case RULE_IF:
-        case RULE_IFNOT: {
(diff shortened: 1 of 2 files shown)


sogaiu:
The changes (and a preliminary port of them to [margaret](https://github.com/sogaiu/margaret)) appear to work here.

PR #1033, opened 2022-09-15
Remove `toggle` from boot.janet

locriacyber:
It doesn't have much reason to be here.


bakpakin:
Any reason why you want to just delete functions from the core library?

pepe:
Also part of the commit is more changes.

locriacyber:
> Any reason why you want to just delete functions from the core library?

I think the core library should be small.

PR #1035, opened 2022-09-16
Add: Allow :user0 - :user9 as argument to `signal`

locriacyber:
Allow :user0 - :user9 as argument to `signal`, and deprecate integer argument to `signal`

Encourage use of `(signal :user0)` instead of `(signal 0)`, to align with return value of `fiber/status`.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -375,7 +375,7 @@
 (defn return
   "Return to a prompt point."
   [to &opt value]
-  (signal 0 [to value]))
+  (signal :user0 [to value]))
 
 (defmacro with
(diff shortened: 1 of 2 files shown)


bakpakin:
So I would prefer to remove the print statement at runtime for deprecated usage, since there is no reason not to use the integer arguments, but otherwise looks good to me.

locriacyber:
That's because I don't know how to add lint message properly. Is it possible to produce `:deprecated` messages?

bakpakin:
Closing as WNF. There are other reasons to use the integer signal, it is slightly faster so makes sense to use in core library functions. We can allow using proper signal names for user signals as a preferred alternative, but without deprecating user signals.

PR #1041, opened 2022-11-06
#include <stdatomic.h> in `core/abstract.c`

autumnull:
Added include for `stdatomic.h`, for compatibility with [`tcc`](https://bellard.org/tcc/), which otherwise complains about `__ATOMIC_RELAXED` being undeclared.

### DIFF
--- a/src/core/abstract.c
+++ b/src/core/abstract.c
@@ -34,6 +34,8 @@
 #endif
 #endif
 
+#include <stdatomic.h>
+
 /* Create new userdata */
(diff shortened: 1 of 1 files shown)


autumnull:
I don't have a windows machine to test this on but it would probably be fixed just by moving the include statement into the `#else` branch of the preproc statement above about windows

bakpakin:
Fixed in bfcfd58259911b92ff516bab9216831b34653805, closing

PR #1052, opened 2022-12-16
(os/cpu-count) should not be defined at all with JANET_REDUCED_OS

dressupgeekout:
`(os/cpu-count)` is not registered in the core if JANET_REDUCED_OS is enabled, so we can compile it out entirely in that case.


pepe (approved):
LGTM

PR #1059, opened 2023-01-11
Make PEGs treat bracketed tuples as sequences

ianthehenry:
This is a backwards-incompatible change, and it's also completely subjective, so no hard feelings if you just close it :). Think of this more like an idea that I'm throwing out that happens to have some (trivial) code attached.

The `*` operator is probably my most frequently used PEG thing, and I think `[]` feels like a natural notation for sequencing that saves a couple characters. For example:

```janet
(def ip-address
 '{:dig (range "09")
   :0-4 (range "04")
   :0-5 (range "05")
   :byte (+
           ["25" :0-5]
           ["2" :0-4 :dig]
           ["1" :dig :dig]
           (between 1 2 :dig))
   :main [:byte "." :byte "." :byte "." :byte]})
```

Mildly easier to type? Mildly easier to read? Loses the nice symmetry with `+` and `*` though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## ??? - Unreleased
+- Bracketed tuples can now be used in place of `sequence`, such that `'["a" "b"]` is the same as `'(* "a" "b")`.
+
 ## 1.26.0 - 2023-01-07
 - Add `ffi/malloc` and `ffi/free`. Useful as tools of last resort.
 - Add `ffi/jitfn` to allow calling function pointers generated at runtime from machine code.
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -1311,6 +1311,10 @@ static uint32_t peg_compile1(Builder *b, Janet peg) {
             const Janet *tup = janet_unwrap_tuple(peg);
             int32_t len = janet_tuple_length(tup);
             if (len == 0) peg_panic(b, "tuple in grammar must have non-zero length");
+            if (janet_tuple_flag(tup) & JANET_TUPLE_FLAG_BRACKETCTOR) {
+                spec_sequence(b, len, tup);
+                break;
+            }
             if (janet_checkint(tup[0])) {
                 int32_t n = janet_unwrap_integer(tup[0]);
                 if (n < 0) {
--- a/test/suite0003.janet
+++ b/test/suite0003.janet
@@ -409,6 +409,11 @@
 (check-match '(* (? "hi") -1) "hi" true)
 (check-match '(* (? "hi") -1) "no" false)
 
+# Bracket sequence
+
+(check-match '["a" "b"] "ab" true)
+(check-match '["a" ["b" "c"]] "ab" false)
+
 # Drop
 
 (check-deep '(drop '"hello") "hello" @[])


pepe:
I do not want to crash any ideas, and I am not the one who will decide in the end. But I would leave brackets for some new uses. :-) I do not have any problem with `(* )`. 

And technically, it only saves half of a couple of characters.

sogaiu:
Though I wonder if anyone has done something like the following:

```
$ janet
Janet 1.26.0-75179de8 linux/x64 - '(doc)' for help
repl:1:> (peg/match '[sequence "a" "b"] "ab")
@[]
```

Wouldn't that sort of thing break with these specific changes?

pepe:
@sogaiu I guess that it was meant by backward-incompatible :-). Anyway, this change does not make sense after another round of thought. :-)

sogaiu:
Ah, thanks for the explanation.

bakpakin:
Neat idea but discussion here indicates it's probably best not to merge. Also, peg short-hands can be confusing enough as is.

sogaiu:
@ianthehenry I found a use case for keeping the interpretation of square brackets for pegs the way they are.

I've been experimenting with doing code search (like [comby](https://comby.dev/)) where one uses pegs instead of regular expressions.

Supposing one wants to match code like:

```c
JANET_DEFINE_MATHOP(exp2, "Returns 2 to the power of x.")
```

Currently, one could write a template like this:

```janet
  (def query-str
    ``
    JANET_DEFINE_MATHOP(<name '(to ",")>, <...>)
    ``)
```

The angle brackets delimit the match / capture constraints.

Suppose one also wanted `JANET_DEFINE_NAMED_MATHOP` and `JANET_DEFINE_MATH2OP`, one might write a template to specify what to search for ike this:
```janet
  (def query-str
    ``
    <(sequence (choice "JANET_DEFINE_NAMED_MATHOP"
                       "JANET_DEFINE_MATH2OP"
                       "JANET_DEFINE_MATHOP"))
    >(<(choice (sequence :string "," :ws)
               "")
    ><name (group (sequence (line)
                            (position)
                            (capture (to ","))))>, 
    <:string>)
    ``)
```
Apart from confusing whitespace, the parens don't help with readability in this case.  However, if I use square brackets, it can look like this:
```janet
  (def query-str
    ``
    <[sequence [choice "JANET_DEFINE_NAMED_MATHOP"
                       "JANET_DEFINE_MATH2OP"
                       "JANET_DEFINE_MATHOP"]]
    >(<[choice [sequence :string "," :ws]
               ""]
    ><name [group [sequence [line]
                            [position]
                            [capture [to ","]]]]>, 
    <:string>)
    ``)
```

This seems a bit more readable.

ianthehenry:
Makes sense! This idea turned into a small exploration in alternative PEG notations, none of which really make sense as a core part of Janet, but that I might publish as a library one day.

sogaiu:
Looking forward to what you come up with :)

BTW, the Janet website has this bit at the syntax and parser page:

```
    :hex (range "09" "af" "AF")
    :escape (* "\\" (+ (set "ntrzfev0\"\\")
                       (* "x" :hex :hex)
                       (* "u" [4 :hex])
                       (* "U" [6 :hex])
                       (error (constant "bad escape"))))
```

So it looks like there was an example hidden in plain sight!

On a side note, it looks like `:h` might be usable instead of `:hex`...even more compact (^^)

PR #1061, opened 2023-01-21
exclude windows specific clib functions when no dynamic modules flag defined

AlecTroemel:
really found by @sogaiu while working on getting https://github.com/nesbox/TIC-80/pull/2079 building on windows.

Without this additional check the function definitions in `util.h` and `util.c` don't match up. but if I'm understanding everything correctly the intent of `JANET_NO_DYNAMIC_MODULES` is to really cut out all these functions

### DIFF
--- a/src/core/util.c
+++ b/src/core/util.c
@@ -927,7 +927,7 @@ const char *error_clib(void) {
 }
 
 #else
-#if defined(JANET_WINDOWS)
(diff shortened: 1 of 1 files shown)


bakpakin:
Yes, this change is good. However, coincidentally I have begun adding mingw support this weekend, that among other changes, has a fix for this. Check if the latest master fixes this, and if so you can close this PR.

AlecTroemel:
yep, confirmed latest master compiles just fine on windows with no dynamic modules. 

closing!

Issue #1067, opened 2023-02-03
Add compiler flag to disable File IO


AlecTroemel:
flag will most likely be named  `JANET_REDUCED_IO`

needed for this issue -> https://github.com/nesbox/TIC-80/issues/2108

The tricky part here is to make sure bootstrapping works correctly. Since print and eprint can be redirected to buffers and functions, the default IO could simply be to do nothing instead of writing to standard handles in that case.

import/use/require functions will still be included, but the well make some changes to the default module loaders to remove actual file IO. Basically just read from module cache.

(Janet would still include stdio.h, but would not expose most of the functionality to programmers)

((these notes are a mostly a summary of comments by bakpakin in gitter)

AlecTroemel:
here's a sketch of relevant functions & calls by sogaiu https://gist.github.com/sogaiu/d50ae376ccff9948bd75c0a773b56fdf

bakpakin:
I've done some preliminary work on adding a `janet_sandbox` function to Janet to disable this sort of thing. The idea is similar to BSD `pledge`, or to Deno's permission API (albeit newer and frankly less granular at this point). This is is easier than compiler defines for a number of reasons:

* Compiler defines are intended for portability - they are frankly a pain to change and cannot be very granular - this add inflexibility to a  project that uses them for security.
* Janet relies on IO for it's bootstrapping process and it's main executable.
* Compiler defines cannot be changed at runtime - this makes it impossible to disable file system access programmatically only after certain setup code has been run.

Adding compiler defines to disable IO did not pan out into something that was either clean or useful, so hopefully this will address the issues in tic-80 (and more generally embedding Janet in applications).

1. https://man.openbsd.org/pledge.2
2. https://deno.land/manual@v1.29.1/basics/permissions

PR #1068, opened 2023-02-03
1067 Reduced File IO flag

AlecTroemel:
Adds `JANET_REDUCED_IO` flag which removes these functions from `src/io.c` 
    ```
    cfun_io_temp
    cfun_io_fopen
    cfun_io_fread
    cfun_io_fwrite
    cfun_io_fflush
    janet_file_close
    cfun_io_gc
    cfun_io_fclose
    cfun_io_fseek
    ```

In `boot.janet`,  file related functions like `slurp` and `spit` throw errors when required file functions not available 

--- 

The big unsolved problem still here is that bootstrapping the amalgamation requires file IO, specifically in  `do-one-file`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ build/janet_boot: $(JANET_BOOT_OBJECTS)
 
 # Now the reason we bootstrap in the first place
 build/c/janet.c: build/janet_boot src/boot/boot.janet
-	build/janet_boot . JANET_PATH '$(JANET_PATH)' > $@
+	build/janet_boot . JANET_PATH '$(JANET_PATH)' 'src/boot/' 'boot.janet' > $@
 	cksum $@
 
 ########################
@@ -184,8 +184,8 @@ endif
 build/c/shell.c: src/mainclient/shell.c
 	cp $< $@
 
-build/janet.h: $(JANET_TARGET) src/include/janet.h $(JANETCONF_HEADER)
-	./$(JANET_TARGET) tools/patch-header.janet src/include/janet.h $(JANETCONF_HEADER) $@
+build/janet.h: build/janet_boot tools/patch-header.janet src/include/janet.h $(JANETCONF_HEADER)
+	build/janet_boot . JANET_PATH '$(JANET_PATH)' 'src/include/janet.h' '$(JANETCONF_HEADER)' '$@' 'tools/' 'patch-header.janet'
 
 build/janetconf.h: $(JANETCONF_HEADER)
 	cp $< $@
--- a/src/boot/boot.c
+++ b/src/boot/boot.c
@@ -33,8 +33,15 @@
 extern const unsigned char *janet_gen_boot;
 extern int32_t janet_gen_boot_size;
 
-int main(int argc, const char **argv) {
+char* concat(const char *s1, const char *s2) {
+    char *result = malloc(strlen(s1) + strlen(s2) + 1); // +1 for the null-terminator
+    // in real code you would check for errors in malloc here
+    strcpy(result, s1);
+    strcat(result, s2);
+    return result;
+}
 
+int main(int argc, const char **argv) {
     /* Init janet */
     janet_init();
 
@@ -55,7 +62,7 @@ int main(int argc, const char **argv) {
 
     /* Create args tuple */
     JanetArray *args = janet_array(argc);
-    for (int i = 0; i < argc; i++)
+    for (int i = 0; i < (argc - 2); i++)
         janet_array_push(args, janet_cstringv(argv[i]));
     janet_def(env, "boot/args", janet_wrap_array(args), "Command line arguments.");
 
@@ -74,8 +81,11 @@ int main(int argc, const char **argv) {
 #ifdef JANET_NO_SOURCEMAPS
     boot_filename = NULL;
 #else
-    boot_filename = "boot.janet";
+    boot_filename = argv[argc - 1];
 #endif
+
+    const char *boot_file_path = argv[argc - 2];
+    char *boot_full_path = concat(boot_file_path, boot_filename);
 
     int chdir_status = chdir(argv[1]);
     if (chdir_status) {
@@ -83,9 +93,9 @@ int main(int argc, const char **argv) {
         exit(1);
     }
 
-    FILE *boot_file = fopen("src/boot/boot.janet", "rb");
+    FILE *boot_file = fopen(boot_full_path, "rb");
     if (NULL == boot_file) {
-        fprintf(stderr, "Could not open src/boot/boot.janet\n");
+        fprintf(stderr, "Could not open %s%s\n", argv[argc - 2], boot_filename);
         exit(1);
     }
 
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1699,9 +1699,9 @@
 (defn slurp
   ``Read all data from a file with name `path` and then close the file.``
   [path]
-  (if-let [fopen (in root-env 'file/open)
-           fread (in root-env 'file/read)
-           fclose (in root-env 'file/close)]
+  (if-let [fopen (get-in root-env 'file/open)
+           fread (get-in root-env 'file/read)
+           fclose (get-in root-env 'file/close)]
     (do (def f (fopen path :rb))
         (if-not f (error (string "could not open file " path)))
         (def contents (fread f :all))
@@ -1712,9 +1712,9 @@
 (defn spit
   ``Write `contents` to a file at `path`. Can optionally append to the file.``
   [path contents &opt mode]
-  (if-let [fopen (in root-env 'file/open)
-           fwrite (in root-env 'file/write)
-           fclose (in root-env 'file/close)]
+  (if-let [fopen (get-in root-env 'file/open)
+           fwrite (get-in root-env 'file/write)
+           fclose (get-in root-env 'file/close)]
     (do (default mode :wb)
         (def f (fopen path mode))
         (if-not f (error (string "could not open file " path " with mode " mode)))
@@ -2301,8 +2301,8 @@
   [where line col]
   (if-not line (break))
   (unless (string? where) (break))
-  (if-let [fopen (in root-env 'file/open)
-           fread (in root-env 'file/read)]
+  (if-let [fopen (get-in root-env 'file/open)
+           fread (get-in root-env 'file/read)]
     (do (when-with [f (fopen where :r)]
                    (def source-code (fread f :all))
                    (var index 0)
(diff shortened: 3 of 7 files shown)


sogaiu:
@AlecTroemel Below is a diff that got me a bit further in the build process -- though I'm not quite sure if it was appropriate.

Also, this doesn't make it to the end, as header patching fails:
```
./build/janet tools/patch-header.janet src/include/janet.h src/conf/janetconf.h build/janet.h
error: file IO disabled
  in dofile [boot.janet] on line 2891, column 5
  in cli-main [boot.janet] on line 3960, column 17
make: *** [Makefile:188: build/janet.h] Error 1
```

Perhaps at least the `boot.janet` tweaks are useful though.

---

```diff
diff --git a/src/boot/boot.janet b/src/boot/boot.janet
index 62192274..46d209f9 100644
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1699,9 +1699,9 @@
 (defn slurp
   ``Read all data from a file with name `path` and then close the file.``
   [path]
-  (if-let [fopen (in root-env 'file/open)
-           fread (in root-env 'file/read)
-           fclose (in root-env 'file/close)]
+  (if-let [fopen (get-in root-env ['file/open :value])
+           fread (get-in root-env ['file/read :value])
+           fclose (get-in root-env ['file/close :value])]
     (do (def f (fopen path :rb))
         (if-not f (error (string "could not open file " path)))
         (def contents (fread f :all))
@@ -1712,9 +1712,9 @@
 (defn spit
   ``Write `contents` to a file at `path`. Can optionally append to the file.``
   [path contents &opt mode]
-  (if-let [fopen (in root-env 'file/open)
-           fwrite (in root-env 'file/write)
-           fclose (in root-env 'file/close)]
+  (if-let [fopen (get-in root-env ['file/open :value])
+           fwrite (get-in root-env ['file/write :value])
+           fclose (get-in root-env ['file/close :value])]
     (do (default mode :wb)
         (def f (fopen path mode))
         (if-not f (error (string "could not open file " path " with mode " mode)))
@@ -2301,8 +2301,8 @@
   [where line col]
   (if-not line (break))
   (unless (string? where) (break))
-  (if-let [fopen (in root-env 'file/open)
-           fread (in root-env 'file/read)]
+  (if-let [fopen (get-in root-env ['file/open :value])
+           fread (get-in root-env ['file/read :value])]
     (do (when-with [f (fopen where :r)]
                    (def source-code (fread f :all))
                    (var index 0)
@@ -2700,9 +2700,9 @@
 # If reduced IO, this function always returns nil
 (defn- fexists
   [path]
-  (if-let [fopen (in root-env 'file/open)
-           fread (in root-env 'file/read)
-           fclose (in root-env 'file/close)]
+  (if-let [fopen (get-in root-env ['file/open :value])
+           fread (get-in root-env ['file/read :value])
+           fclose (get-in root-env ['file/close :value])]
     (compif (dyn 'os/stat)
             (= :file (os/stat path :mode))
             (when-let [f (fopen path :rb)]
@@ -2827,7 +2827,7 @@
   `run-context` call. If `exit` is true, any top level errors will trigger a
   call to `(os/exit 1)` after printing the error.``
   [path &named exit env source expander evaluator read parser]
-  (if-let [fopen (in root-env 'file/open)]
+  (if-let [fopen (get-in root-env ['file/open :value])]
     (do (def f (case (type path)
                  :core/file path
                  :core/stream path
@@ -2912,7 +2912,7 @@
   ``A table of loading method names to loading functions.
   This table lets `require` and `import` load many different kinds
   of files as modules.``
-  (if-let [fopen (in root-env 'file/open)]
+  (if-let [fopen (get-in root-env ['file/open :value])]
     @{:native native-loader
       :source source-loader
       :preload preload-loader
diff --git a/src/core/io.c b/src/core/io.c
index c70fc869..2cd16e02 100644
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -56,7 +56,7 @@ const JanetAbstractType janet_file_type = {
     JANET_ATEND_NEXT
 };
 
-#ifndef JANET_REDUCED_IO
+#if !defined(JANET_REDUCED_IO) || defined(JANET_BOOTSTRAP)
 /* Check arguments to fopen */
 static int32_t checkflags(const uint8_t *str) {
     int32_t flags = 0;
@@ -114,7 +114,7 @@ static void *makef(FILE *f, int32_t flags) {
     return iof;
 }
 
-#ifndef JANET_REDUCED_IO
+#if !defined(JANET_REDUCED_IO) || defined(JANET_BOOTSTRAP)
 JANET_CORE_FN(cfun_io_temp,
               "(file/temp)",
               "Open an anonymous temporary file that is removed on close. "
@@ -286,7 +286,7 @@ static int cfun_io_gc(void *p, size_t len) {
     return 0;
 }
 
-#ifndef JANET_REDUCED_IO
+#if !defined(JANET_REDUCED_IO) || defined(JANET_BOOTSTRAP)
 /* Close a file */
 JANET_CORE_FN(cfun_io_fclose,
```

AlecTroemel:
@sogaiu I've applied your changes, should definitely be using `get-in` in boot and the `|| defined(JANET_BOOTSTRAP)` checks make sense to me. I also added that extra or check down in the io's module entry point.

AlecTroemel:
Ah, our get-in form was still a little off. using this got my build further `(get-in root-env ['file/open :value])`. new error is this

```
➜   make
./build/janet tools/patch-header.janet src/include/janet.h src/conf/janetconf.h build/janet.h
error: unknown method :read invoked on <core/file 0x55B5AB2C5F30>
  in chunks [boot.janet] on line 2841, column 30
  in run-context [boot.janet] on line 2508, column 12
  in dofile [boot.janet] on line 2864, column 11
  in cli-main [boot.janet] on line 3960, column 17
make: *** [Makefile:188: build/janet.h] Error 1
```

sogaiu:
https://github.com/janet-lang/janet/pull/1068/commits/2a20d4c28f20cdc6191df062fdb7378da8534fa1 looks better.

The error I'm getting from that now matches (I think) what I got earlier as stated at: https://github.com/janet-lang/janet/pull/1068#issuecomment-1416687299

sogaiu:
IIUC, at this point `build/janet` doesn't have file IO capabilities so the file header patching it tries to do fails (or something like that).

sogaiu:
I wonder if there is some way to get `build/janet_boot` (or a slight tweak?) to do the job.

sogaiu:
Right now `src/boot/boot.c` (what becomes `build/janet_boot`?) is hard-wired to process `boot.janet` (when `JANET_NO_SOURCEMAPS` is not defined?).

If we could customize which file it processed as Janet source I wonder if we could get it to do the file patching...

AlecTroemel:
@sogaiu you talking about [this filename](https://github.com/janet-lang/janet/blob/master/src/boot/boot.c#L77) being customizable? I'm still trying to understand how this part of Janet works.. assuming we did swap `boot.janet` out with something else, what would it do?

sogaiu:
@AlecTroemel I'm going to try to spell things out a bit (may be if the description is off we can improve it and its corresponding understanding).

IIUC, we are encountering an error on is this line of the `Makefile`:

https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/Makefile#L188

I think what that does is to run a [short `.janet` script](https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/tools/patch-header.janet):

```janet
# Patch janet.h 
(def [_ janeth janetconf output] (dyn :args))
(spit output (string/replace `#include "janetconf.h"` (slurp janetconf) (slurp janeth)))
```

It looks like that needs file IO.

If we stick with trying to use some kind of `janet` binary to accomplish the task, a full `JANET_REDUCED_IO`-version of `janet` isn't going to work.

Ok, so where can we get something that can execute `.janet` content?

AFAIU, [`boot.c`](https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/src/boot/boot.c) is essentially a program that embeds Janet to build `janet.c` and further, it can do file IO (even with `JANET_REDUCED_IO`, because it's built with `JANET_BOOTSTRAP` and recent changes in this PR ensure file IO capabilities remain intact for it).

It does the building of `janet.c` by executing `boot.janet`.  The tail end of `boot.janet` is all about [printing out `janet.c`](https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/src/boot/boot.janet#L3962-L4104).

IIUC, in its current state, `boot.c` seems to only know how to execute one specific `.janet` file -- `boot.janet`.  This seems to be hard-wired in a few places:

* https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/src/boot/boot.c#L77
* https://github.com/janet-lang/janet/blob/55af6ce83443360a2d9115373d613bf240ca0fa7/src/boot/boot.c#L86-L88

If we change `boot.c` so that it can read some file other than `boot.janet` (say, via the content of some environment variables), then we can get it to execute a different `.janet` file.  Specifically, one that can perform the step that is causing a failure -- the header patchings step.

Does that make sense?

AlecTroemel:
@sogaiu thank you for that amazingly detailed write up, its really helps! 

One possible concern; If we open up `boot.c` to execute arbitrary Janet (through a new command line argument), are there any "security" implications? In one sense what we would have done is create a second Janet executable that has elevated privileges (for example file IO even though we're building Janet explicitly without it). 

I'm not sure how someone could find or use `boot.c` maliciously.. so this could be a non-issue?

Unless someone swoops in and says this is a horrible idea, I'm going to try and modify `boot.c` as you've described.

sogaiu:
@AlecTroemel Regarding:

> I'm not sure how someone could find or use boot.c maliciously.. so this could be a non-issue?

I don't think `boot.c` is used for anything other than bootstrapping but that is just a guess.

May be this PR won't be merged without scrutiny along those lines...I do wonder a bit if there isn't a simpler approach, but if the path we're on now gets us to something working, perhaps we can revisit at that point?

AlecTroemel:
finally got `make` to succeed! Now on to `make test` ...

sogaiu:
https://github.com/janet-lang/janet/pull/1068/commits/47faa1b77f446e8f222283f08209eab830fc8419 built here too :+1: 

Still thinking about `make test`...I haven't tried it with other "limiting" flags in place.

I wonder if it's meant to work under those kinds of circumstances.

It does seem worth having some testing of sorts though.

Hmm...

bakpakin:
Way too much going on here, I will not merge this as is, even if builds start passing - needing to change the Makefile is a huge red flag here (not too mention, by convention new configuration parameters are added to janetconf.h and to meson as well).

You have like 3 different mechanisms going on here to configure the output. There is no need to override boot.janet in anyway - instead, look to how we do it for other optional stuff like net and ev - we use the handy macro `compwhen` for the most part.

As far as making boot.janet work without all IO primitives, all that boot.janet really needs to do it's job is write the contents of janet.c to stdout. In boot.c, we can use the replacements functionality to inject whatever extra stuff we need to complete bootstrapping, such as `boot/stdout` and calling `(setdyn :out boot/stdout)` in the bootstrap phase. Obviously there are some subtleties here but you should be able to get around this without changing the Makefile

bakpakin:
Closing this in preference to adding sandboxing functions to the C API - granular permissions that allow functionality to be disabled at runtime. Compiler defines should be reserved more for portability rather than permission based security.

AlecTroemel:
thank you for taking a look at this @bakpakin. the Granular permissions you described [here](https://github.com/janet-lang/janet/issues/1067#issuecomment-1419251319) pretty easily convinced me that that solution is the right call!

PR #1072, opened 2023-02-21
Add file/tell

sogaiu:
I'd like to reuse a single temp file by writing to it (e.g. via `os/execute`'s `{:out tf}`), seeking to the beginning of the file, and rereading just enough of the written content (and doing this repeatedly).

I'm not sure how to find out how much was written to the file though.

It seems like being able to use `ftell` (say wrapped in `file/tell`) could help with this as one could call `file/tell` before seeking to the beginning and then reading just enough bytes using the result from the `file/tell` call.

Does the addition of `file/tell` seem reasonable?

---

Regarding the code, IIUC, `ftell` returns a `long`, but I don't know how big that would be on the various platforms that Janet is supposed to run on.  I chose to go with `janet_wrap_number` and cast `ftell`'s return value to a `double`.  Does that seem reasonable?

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -348,11 +348,24 @@ JANET_CORE_FN(cfun_io_fseek,
     return argv[0];
 }
 
+JANET_CORE_FN(cfun_io_ftell,
+              "(file/tell f)",
+              "Get the current value of the file position for file `f`.") {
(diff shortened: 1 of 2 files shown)


bakpakin:
LGTM, and seems a very reasonable addition.

PR #1097, opened 2023-04-08
posix_spawn(p) doesn't always set errno

### LINKED ISSUE #1015: When a process is not found, `os/execute` should give a descriptive error
Both `os/execute` and `os/spawn` give unhelpful errors when the process being called does not exist.

Consider the following code:
```janet
(os/execute ["non-existent"])
(os/execute ["non-existent"] :p)
(os/spawn ["non-existent"])
(os/spawn ["non-existent"] :p)
```

Running this command should indicate that the underlying process doesn't exist.

Ideally, using the `:p` flag would contain a reference to the current path.

However, the current implementation gives the following messages:
```
Janet 1.23.0-meson macos/aarch64 - '(doc)' for help
repl:1:> (os/execute ["non-existent"])
error: ("non-existent"): Undefined error: 0
  in os/execute [src/core/os.c] on line 1115
repl:2:> (os/execute ["non-existent"] :p)
error: ("non-existent"): Invalid argument
  in os/execute [src/core/os.c] on line 1115
  in _thunk [repl] (tailcall) on line 2, column 1
repl:3:> (os/spawn ["non-existent"])
error: ("non-existent"): Invalid argument
  in os/spawn [src/core/os.c] on line 1128
  in _thunk [repl] (tailcall) on line 3, column 1
repl:4:> (os/spawn ["non-existent"] :p)
error: ("non-existent"): Invalid argument
  in os/spawn [src/core/os.c] on line 1128
  in _thunk [repl] (tailcall) on line 4, column 1
```

Interestingly, calling `(os/execute ["non-existent"])` returns different error messages the first time it's called. It says "Undefined error: 0" the first time and "Invalid argument" the second (and third) time.
Switching to `os/spawn` or adding the `:p` flag doesn't change this.

Sorry for filing so many issues about descriptive error-messages, just think of it as documenting room for improvement ❤️

CosmicToast:
This also sets the stage up for custom errno handling, in case more expressive error messages for other failure conditions end up being desirable.
Fixes #1015

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -1085,7 +1085,14 @@ static Janet os_execute_impl(int32_t argc, Janet *argv, int is_spawn) {
 
     os_execute_cleanup(envp, child_argv);
     if (status) {
-        janet_panicf("%p: %s", argv[0], strerror(errno));
+        switch (errno) {
+            // on macOS, the binary not existing does not cause errno to be set
+            case 0:
+                janet_panicf("%p: No such file or directory", argv[0]);
+                break;
+            default:
+                janet_panicf("%p: %s", argv[0], strerror(errno));
+        }
     }
 
 #endif
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -1085,14 +1085,7 @@ static Janet os_execute_impl(int32_t argc, Janet *argv, int is_spawn) {
 
     os_execute_cleanup(envp, child_argv);
     if (status) {
-        switch (errno) {
-            // on macOS, the binary not existing does not cause errno to be set
-            case 0:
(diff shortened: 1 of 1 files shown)


bakpakin:
A bit simpler and not triggering CodeQL is the statement `janet_panicf("%p: %s", argv[0], strerror(errno ? errno : ENOENT));`

CosmicToast:
Okay, that can be done. The main reason I kept it in the original form was because other edge-cases like that are likely to pop up eventually, + it could be nice to provide more complete error messages for specific errnos in the future.
Still, those are future considerations, so I've updated to do the above.

PR #1106, opened 2023-04-19
handle null-byte case in scanformat (fixes #1105)

### LINKED ISSUE #1105: Segmentation fault triggered via printf
I mistyped a format string and got a segfault:
```
$ janet
Janet 1.27.0-bef5bd72 linux/x64/gcc - '(doc)' for help
repl:1:> (printf "hello %" `dave`)
Segmentation fault (core dumped)
```

CosmicToast:
When there is no format to be found after a %, get_fmt_mapping returns NULL.
It then gets called against strlen, which is a typical SEGV.
Check for NULL aginst mapping, which signals a null format being specified.

Note that there are other edge-conditions present.
Namely, consider: `(printf "hello % " "dave")` (format is still null).
Meanwhile, `(printf "hello % s" "dave")` (outputs "hello dave").

### DIFF
--- a/src/core/pp.c
+++ b/src/core/pp.c
@@ -809,9 +809,9 @@ static const char *scanformat(
     *(form++) = '%';
     const char *p2 = strfrmt;
     while (p2 <= p) {
-        if (strchr(FMT_REPLACE_INTTYPES, *p2) != NULL) {
+        char *loc = strchr(FMT_REPLACE_INTTYPES, *p2);
+        if (loc != NULL && *loc != '\0') {
             const char *mapping = get_fmt_mapping(*p2++);
-            if (!mapping) janet_panic("invalid format (found null)");
             size_t len = strlen(mapping);
             strcpy(form, mapping);
             form += len;
--- a/src/core/pp.c
+++ b/src/core/pp.c
@@ -811,6 +811,7 @@ static const char *scanformat(
     while (p2 <= p) {
         if (strchr(FMT_REPLACE_INTTYPES, *p2) != NULL) {
             const char *mapping = get_fmt_mapping(*p2++);
+            if (!mapping) janet_panic("invalid format (found null)");
             size_t len = strlen(mapping);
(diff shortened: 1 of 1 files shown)


pepe (commented):
I would love to have a bit of information about what happened. And maybe that is just me.

pepe:
I mean in the error message.

sogaiu:
Ah, I finally get it, @pepe, you mean that may be `"invalid format (found null)"` is a bit on the terse side?

Do you have some specific ideas about what one might say in addition or instead?

sogaiu:
Ah I see.  That might be nice.

Though if you look above the newly added code, I think you'll see that there are a couple of other error messages which are very much in the same style (i.e. no positional info) :)

CosmicToast:
It's rather complicated because the information on the string format location is partial and lost (see: the "% " example), as well as disparate between the two functions (one of them doesn't hold an intermediary pointer), which is why the existing error messages don't hold location information.
To make this a bit more useful, I addressed the source of the error instead of the symptom (i.e the strchr was checked against NULL, but not against returning the location of `\0`, which it can and should according to the specification).
This means that the error handling falls through into `janet_formatbv` and `janet_buffer_format`, which then fall through to the default case.
The new error messages look like so:
```
repl:1:> (printf "hello % " 10)
error: invalid conversion '% ' to 'format'
  in printf [src/core/io.c] on line 626
repl:2:> (printf "hello %" 10)
error: invalid conversion '%' to 'format'
  in printf [src/core/io.c] on line 626
  in _thunk [repl] (tailcall) on line 2, column 1
```

This means that if a pass was ever done on those two functions (including to potentially deduplicate them), this case would benefit alongside all the other cases with location information.

pepe:
Oh thank you both.

bakpakin:
Looks like a good fix to me right now, we could definitely improve some of the error handling in some cases, but I think it is more important to fix segfaults and things before worrying about the error case flow

Issue #1013, opened 2022-08-18
peg/replace should accept functions for replacement


Techcable:
I would expect the following code to work.

```janet
(defn escape-string [target]
   (def escaped (peg/replace-all '(choice (set ``\\\"\n\t\0")) (fn [bad] (string/format "\\%s" bad)) target))
   (string/format ``"%s"`` escaped))
```

However, this code currently fails with the following error:
```
error: bad slot #1, expected string|symbol|keyword|buffer, got <function 0x6000037A84A0>
  in peg/replace-all [src/core/peg.c] on line 1733
  in escape-string [repl] on line 2, column 17
  in _thunk [repl] (tailcall) on line 4, column 1
```

I would say we should support "replacement functions" in addition to literal replacements. These functions would be called and would take one or two arguments ([matchedText &opt patternCaptures])

The return result of that function would then be expected to be a string. This would be much more flexible and consistent with the regex capabilities of other languages.

bakpakin:
Just to be clear, peg/match can be used for replacements as well, and can do function replacements - peg/replace is just a convenience function. 

I would be open to adding this though if it is easy to add.

autumnull:
It would also be useful if `peg/replace-all` had this feature too!

PR #1112, opened 2023-04-23
string and peg replacement functions can now take functions

ianthehenry:
Functions will be invoked with the matched text, and their result will be coerced to a string and used as the new replacement text. PEG replacement functions will also receive any captures as additional arguments after the full matching text.

This also allows passing non-function, non-byteviewable values, which will be converted to strings during replacement (only once, and only if at least one match is found).

- I standardized on `subst` instead of `repl` for this concept
- It seems a little silly to allow a function argument to `string/replace`, but I felt it was nicer if all the replacement functions behaved the same way. And it does mean that you can potentially save yourself work if no match is found. You can also pass an impure function to `string/replace-all` and return different replacements for each match, which might be useful.


ianthehenry:
Addresses #1013

pepe (approved):
LGTM

pepe:
I like the implementation.

PR #1113, opened 2023-04-24
add catseq

ianthehenry:
While `seq` is a `loop`-based alternative to `map`, `catseq` is a `loop`-based alternative to `mapcat`.

I've written this macro in multiple projects now, and I think it's a good fit to include in the standard library.


pepe (approved):
LGTM

CosmicToast (approved):

PR #1114, opened 2023-04-24
drop with a negative count now drops from the end

ianthehenry:
This is a functional change, as previously a negative count would be treated the same as 0. But it seems unlikely to cause problems in practice.

This brings a satisfying symmetry to the quadrants of this "table:"

- First element: `(first list)`
- Last element: `(last list)`
- Everything but the first element: `(drop 1 list)`
- Everything but the last element: `(drop -1 list)`

It might make sense to do the same thing for `take`, but because `take` works with fibers, we can't support it completely.

It might be preferable to have a separate `drop-last` function. The negative count thing seems nicer to me, but I don't know if there's precedent for it elsewhere in the standard library.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1133,16 +1133,17 @@
   (take-until (complement pred) ind))
 
 (defn drop
-  ``Drop the first n elements in an indexed or bytes type. Returns a new tuple or string
-  instance, respectively.``
+  ``Drop the first `n elements in an indexed or bytes type. Returns a new tuple or string
+  instance, respectively. If `n` is negative, drops the last `n` elements instead.``
   [n ind]
   (def use-str (bytes? ind))
   (def f (if use-str string/slice tuple/slice))
   (def len (length ind))
-  # make sure start is in [0, len]
-  (def m (if (> n 0) n 0))
-  (def start (if (> m len) len m))
-  (f ind start -1))
+  (def [start end]
+    (if (>= n 0)
+      [(min n len) len]
+      [0 (max 0 (+ len n))]))
+  (f ind start end))
 
 (defn drop-until
   "Same as `(drop-while (complement pred) ind)`."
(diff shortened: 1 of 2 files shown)


sogaiu:
I think the negative count thing is nice :+1:

I wonder if it might break some existing code though.

pepe:
In many places in the std lib, -1 means the last member, so this is a very accurate addition. And there hardly can be code that only works with the old behavior.

sogaiu:
> In many places in the std lib, -1 means the last member

One that I made the connection for only recently is in the context of PEG specials.  I had been puzzled for quite some time why `-1` is used for what it is, but I think the above quote might apply :)

zevv:
Not quite true: In the PEG module, `-1` is not some special convention; it just follows from the meaning of the pattern: positive numbers mean "matches if there are this many characters (and advance)", whereas negative numbers mean "matches if there are not this many characters (and not advance)". So `-1` will match if there is not at least 1 character available, which only is true for the end of the matched subject.

CosmicToast:
As a quick question re: this context. 
is my assumption that 0 means “always match” reasonable?
I was playing around with ircv3 parsers and needed an “always match” and “never match” primitive for… reasons. 
I used 0 and “! 0”.

CosmicToast:
After thinking about it I opened a PR to just add `true` and `false` PEG semantics, since I think those are reasonable and are less error-prone in the end: https://github.com/janet-lang/janet/pull/1187

sogaiu:
>  In the PEG module, -1 is not some special convention; it just follows from the meaning of the pattern: positive numbers mean "matches if there are this many characters (and advance)", whereas negative numbers mean "matches if there are not this many characters (and not advance)". So -1 will match if there is not at least 1 character available, which only is true for the end of the matched subject.

Ah, thanks for spelling it out like this.  Much appreciated.

Issue #1117, opened 2023-05-01
Misbehavior of logical operators and/or with splice


primo-ppcg:
Logical operators misbehave when passed a splice. For example:

```janet
(var a [nil nil nil 1 2])

# find the first truthy value in the tuple
(or (splice a))
```
Expected: `1`
Actual: `(nil nil nil 1 2)`

This behavior is documented only in the form of a community example: <https://janetdocs.com/or>
```janet
# note that `or` does not behave as you might expect
# when used with `apply` and `splice`:
(or 1 2 3)             # => 1
(or (splice [1 2 3]))  # => (1 2 3)
(apply or [1 2 3])     # => (if 1 1 (if 2 2 3))
```
However, this is not exactly correct. The result is not a tuple, but the splice of the tuple.

Witness:
```janet
(type (or (splice a)))
# error: arity mismatch, expected 1, got 5
```
Of course, the initial example could rewritten - arguably more legibly - as `(find truthy? a)`, but that's beside the point. This behavior appears to be inconsistent with the rest of the Core API. Both `and` and `or` are basic operators, and not special forms (or at least shouldn't be).

bakpakin:
This is expected behavior - if anything were to change, this perhaps should be compiler error. Splice only has meaning in certain locations, and notably macros can do whatever they want the splice form.

primo-ppcg:
I appreciate the responds.

> This is expected behavior
Is there any sound argument why the current behavior is intended or desirable?

bakpakin:
Here is some discussion for Clojure which is similar in this regard: https://stackoverflow.com/questions/2969543/why-doesnt-apply-or-true-false-work-in-clojure

Janet has as similar functions `any?`, `some`, and `all` for doing this in a functional, non-short circuiting manner.  Perhaps something like `all?` could be added to complement `any?`, but it is quite easy to call:

```janet
(any? a b ;(my-function-call) c) # functional form of `or`
(all identity a b ;(my-function-call) c) # functional form of `and`
```

primo-ppcg:
Thanks for the clarification. Perhaps I didn't read the docs closely enough, but I would have expected `any?` to always return a boolean. As it is, it covers this use case entirely. An `all` form without a predicate does seem like a missing symmetry to `some`/`any?`.

PR #1120, opened 2023-05-06
Tweak and / or

primo-ppcg:
- Fixes the following error: `and` fails to return the last truthy value, if that value is `true`.
```janet
(and true 1 true)
```
Expected: `true`
Actual: `1`
- Changes naming conventions and formatting of `or` to match that of `and`.
- Recursively expands splices in both `and` and `or`. #1117 
- Updates test cases.


bakpakin:
The handling of the splice code is wrong and useless - it can only handle splicing statically known arrays, which defeats the point of splice in the first place. For example:

```
(and ;(my-array-ctor))
```
expands to `my-array-ctor`, which is clearly wrong.

You cannot handle splice like this, the entire point of splice is that you don't know the number of parameters passed at compile time, and is likely something that can't even be determined. Also the compiler supports other handling of splice that this simple inspection-based macro implementation cannot handle. For example:

```
(+ (do (splice [1 2 3 4])))
```

Even though splice is inside `do`, it is expanded in. Admittedly this is non guaranteed behavior, but begs the question of why we are having special handling of splice in the `and` and `or` macros.

primo-ppcg:
```janet
(and true 1 true)
```
still stands as an open bug.

Best of luck with your project.

bakpakin:
fixed in 4cc4a9d38ba4bf9271f96a820decb64c33d61de5

bakpakin:
Thank you for the bug report, that bug has been sitting in plain sight for a while - it mostly went unnoticed since it doesn't change behavior when logical macros are used as conditions.

The splice behavior though was already closed as "not a bug", so I won't merge that. The bug fix for `(and true 1 true)` would have been better received if it was not intertwined with "fixes" for behavior that was already closed as "not a bug" in #1117. And didn't itself have bugs. The better fix is much simpler and more correct in this case.

I did not respond during the week since I have a day job, so lately I have been doing most of my Janet work on the weekends.

primo-ppcg:
I interpretted from your comments, specifically
> notably macros can do whatever they want the splice form

that the issue was closed as "will not fix" rather than "not a bug." I personally find the current behavior to be surprising, and I don't believe I'm the only one. That forms are passed to macros unevaluated, necessary in this case for short-circuiting behavior, does complicate the matter somewhat. Possibly more than is worth dealing with.

PR #1140, opened 2023-05-16
Enable debug symbols in janet binary; strip target at 'make install'

zevv:
This makes `make debug` work and `make valtest` show traces with with line numbers. Also valgrind now is run with `--guiet` to hide the banner popping up, resulting in cleaner output.

I saw the `-g` options have been added and removed a few times in the lifetime of the `Makefile`, so I'm not sure what the intention of the authors was with regards to debug symbols.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,11 @@ SONAME_SETTER=-Wl,-soname,
 HOSTCC?=$(CC)
 HOSTAR?=$(AR)
 CFLAGS?=-O2
-LDFLAGS?=-rdynamic
+LDFLAGS?=-rdynamic -g
 RUN:=$(RUN)
 
-COMMON_CFLAGS:=-std=c99 -Wall -Wextra -Isrc/include -Isrc/conf -fvisibility=hidden -fPIC
-BOOT_CFLAGS:=-DJANET_BOOTSTRAP -DJANET_BUILD=$(JANET_BUILD) -O0 -g $(COMMON_CFLAGS)
+COMMON_CFLAGS:=-std=c99 -Wall -Wextra -Isrc/include -Isrc/conf -fvisibility=hidden -fPIC -g
+BOOT_CFLAGS:=-DJANET_BOOTSTRAP -DJANET_BUILD=$(JANET_BUILD) -O0 $(COMMON_CFLAGS)
 BUILD_CFLAGS:=$(CFLAGS) $(COMMON_CFLAGS)
 
 # For installation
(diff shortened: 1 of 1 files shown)


CosmicToast (approved):

bakpakin:
LGTM - might even make sense to leave symbols in

bakpakin:
although, probably easier to just add -g to CFLAGS rather than LDFLAGS, etc.

sogaiu:
With respect to debugging, the combinations of flags I've tried all have one drawback or another.

With the latest commit in this PR, for example, I don't seem to be able to use `janet_csymbolv`:

```
No symbol "janet_csymbolv" in current context.
```

I think for a combination of `-Og` and `-g3` I can use `janet_csymbolv` but not `janet_type` -- should check this I guess.

zevv:
> Perhaps there is no all-around winning combo.

My goal was merely to have some basic debug info available in my builds to get a decent stack trace in gdb and valgrind when running into issues.

If you want to go all-in on the debug symbols, including at the preprocessor level, you indeed need an `-Og` build; that would be better for doing *actual* debugging because it'll make sure not to reorder and optimize out things. The price is if course that this is an unoptimized build, which you do not want to use as your final build asset.

We could add some simple logic to the Makefile to change the optimization to `-Og` when the `make debug` target is built? 
- regular builds -> `-O2` +  `-g`
- debug build -> dwarf + `-O0` + `-g3`

sogaiu:
I think it would be convenient personally but don't want the change if it disrupts other folks' situations (including maintenance burden).

It's been ok changing the `Makefile` manually up to now.

FWIW, with `-O0` and `-g3`, I think `janet_type` may not work:

```
(gdb) p janet_type(janet_next(janet_wrap_table(env), janet_wrap_nil()))
No symbol "__builtin_isnan" in current context.
```

though the following works:

```
(gdb) p janet_csymbolv("hi")
$5 = {
  u64 = 18445289786330884696,
  i64 = -1454287378666920,
  number = -nan(0xad5555564aa58),
  pointer = 0xfffad5555564aa58
}
```

In contrast, `CFLAGS?=-O0 -g` gives:

```
(gdb) p janet_type(janet_next(janet_wrap_table(env), janet_wrap_nil()))
$1 = JANET_SYMBOL
```

but:

```
(gdb) p janet_csymbolv("hi")
No symbol "janet_csymbolv" in current context.
```

This is part of what I meant by:

> Perhaps there is no all-around winning combo.

sogaiu:
On a side note, with the main change in 2f69cd42, it's probably easier to maintain / switch out customized / tweaked builds.

I haven't tried using `config.mk` yet but I suspect it will be helpful for using various combinations of debug-related flags.

PR #1141, opened 2023-05-18
Add CI test for mingw/wine on linux

zevv:
Not sure if this adds anything to the current CI, also the job is a tad more heavy than the others because it needs to install wine and mingw cross compilers first.

Feel free to not apply :)


bakpakin:
Nope, looks good to me. I am always looking for more CI - our current CI load is fairly small, and I already do extra CI on sr.ht for BSD testing

Issue #1092, opened 2023-04-06
Add functions to change objects to be mutable and immutable


Inc0n:
This would be effectively `transient` and `persistent!` in Clojure. 
`persistent!` would be `freeze` of Janet but with the deep freeze.

For example, let's define them as `mutable` and `freeze` (shallow version).
tuple and array `(mutable [1 2])` = `@[1 2]`
struct and dict `(mutable {1 2})` = `@{1 2}`
string and buffer ...

And `freeze` in the inverse operator of `mutable`

This would also imply that `@` is made into a macro reader character

bakpakin:
Reasonable idea. Perhaps for symmetry the function could be called `thaw` or `unfreeze` - maybe a bit too cutesy

sogaiu:
@Inc0n #1142 added by @tionis has `thaw` in it now, do you think that addresses the original issue?

Inc0n:
Hi, thanks for the new patch.

Originally I was trying to do this I believe (now with `thaw` in place, hopefully this would work now):
```
(defn cons [x xs]
  (freeze
   (array/push (thaw xs) x)))
```
For the reason that I did not enjoy the idea passing a mutable array across my functions, however, I also wanted linked list essentially...

CosmicToast:
Couldn't you just do `[;xs (freeze x)]` instead?

sogaiu:
Nice idea with the splice :+1: 

Perhaps though to emulate `cons` one might want the order to be `[(freeze x) ;xs]`?

Inc0n:
> Couldn't you just do `[;xs (freeze x)]` instead?

For my problem, this is a much better solution.

And for the details of the `xs` in my case.
Really, I was trying to make alist happen (Or a list of structs). As data structure characteristics, I was looking for:
- maintain insertion order, i.e. newest always at last or first, second newest always second or second to last
- adding new element is done without mutating the last list

It's screaming linked-list isn't it :)

And would the corresponding accessors looking like the following, if I were to use `[(freeze x) ;xs]`?
```
(def car [[p & more]] p)
(def cdr [[p & more]] more)
```

CosmicToast:
> And would the corresponding accessors ...

Yup.
Just be careful with first/last semantics, but 5 minutes in a REPL clears those up.

sogaiu:
> And would the corresponding accessors looking like the following, if I were to use [(freeze x) ;xs]?

Those seem like they could work, but...

[...]

PR #1142, opened 2023-05-19
added thaw to complement freeze

tionis:
Relevant issue: #1092
Not sure if all cases are handled here.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -2142,6 +2142,19 @@
     :buffer (string x)
     x))
 
+(defn thaw
+  `Thaw an object (make it mutable) and do a deep copy, making
+  child value also mutable. Closures, fibers, and abstract
+  types will not be recursively thawed, but all other types will`
+  [ds]
+  (case (type ds)
+    :table (walk-dict thaw ds)
+    :struct (walk-dict thaw ds)
+    :array (walk-ind thaw ds)
+    :tuple (walk-ind thaw ds)
+    :string (buffer ds)
+    ds))
+
 (defn macex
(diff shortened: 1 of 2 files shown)


sogaiu:
May be prototype stuff is not being handled: 

```janet
(def ds @{:a 1})

(table/setproto ds @{:b "2"})

(get ds :b)
# =>
"2"

(get (thaw ds) :b)
# =>
nil
```

IIUC a struct can be created with a prototype via `struct/with-proto` so may be it makes sense to try to account for that sort of thing?

tionis:
Interesting, I'll look into it. Thanks!

tionis:
OK, prototypes don't appear in normal lookups, I guess I could check recursively for prototypes in structs and maps and inline them with the uppermost struct. Not sure about the performance implications of that approach, but I'll implement it quickly.

tionis:
Oh nevermind, I can just use proto-flatten

tionis:
Ok, should be fixed now

sogaiu:
Worked here too :+1: 

Will you be adding any tests?

zevv:
Minor comment: maybe handle the different types in the same order as `(freeze)` does, which makes it harder to compare the implementations and maybe later add different types.

tionis:
I added simple tests for thaw and freeze and reordered the types in the implementation

sogaiu:
On a side note, going over the tests I started to wonder whether it would be worth having functions like `mutable?` / `immutable?` and may be `deep-*` versions.

I didn't see any functions like that already, but I don't know if they'd be that useful.

PR #1152, opened 2023-05-22
Improved various error messages when handling unexpected types.

zevv:
```
error: bad slot #1, expected string|symbol|keyword|buffer, got ...
error: bad slot #1, expected a string, symbol, keyword or buffer, got ...
```

```
bad s64 initializer: "donkey"
can not convert string "donkey" to s64
```

In other places error message do not talk about "s64"  but about "64 bit signed integer", would that be preferable here as well?

### DIFF
--- a/src/core/corelib.c
+++ b/src/core/corelib.c
@@ -458,7 +458,7 @@ JANET_CORE_FN(janet_core_getproto,
                ? janet_wrap_struct(janet_struct_proto(st))
                : janet_wrap_nil();
     }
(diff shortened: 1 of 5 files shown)


bakpakin:
The pipe syntax for a type union may appear in other places in the code, but since the new message is very clear, I don't think this will cause _more_ confusion. These error messages are certainly no worse - as far s64 vs. "64 bit signed integer", I don't mind either way but if we are going the route of having as simple, readable prose as possible for error messages, "64 bit signed integer" is definitely better.

PR #1157, opened 2023-05-24
Add file/lines iterator

zevv:
That's what you get for coming up with a nice implementation. Still trying to get this into the stdlib :smile:

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1749,18 +1749,13 @@
   (printf (dyn *pretty-format* "%q") x)
   (flush))
 
+
 (defn file/lines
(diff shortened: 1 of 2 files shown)


primo-ppcg (commented):
I would like to reiterate my objections here.

I agree that this would be a useful additon, in particular to iterate lines of `stdin` as a command line tool. However, every other `file/*` function requires a valid file descriptor as its argument, and I don't agree that this function is special enough or important enough to be an exception.
```janet
(defn file/lines
  "Return an iterator over the lines of a file."
  [file]
  (coro
    (while (def line (file/read file :line))
      (yield line))))
```
I think this is a sufficient implementation.

PR #1161, opened 2023-05-25
Change  pp to return the passed value; this allows  to be used similar to Elixirs IO.inspect

zevv:
which makes it a handy tool to "spy" on data by inserting `pp` somewhere in code without affecting behavior.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1747,7 +1747,8 @@
   ``Pretty-print to stdout or `(dyn *out*)`. The format string used is `(dyn *pretty-format* "%q")`.``
   [x]
   (printf (dyn *pretty-format* "%q") x)
-  (flush))
+  (flush)
+  x)
 
 ###
 ###


CosmicToast:
You may be interested [tracev](https://janet-lang.org/api/index.html#tracev).
It's particularly relevant since `pp` prints to stdout while tracev does so to stderr, and also informs you of the source mapping location of the print (so you can have multiple of them active at a time).

sogaiu:
In addition to what @CosmicToast said, the following thoughts came up:

* The print-types of functions (`pri*`, `epr*`, `xpr*`) generally return `nil` so possibly some might find it a bit surprising (though perhaps one could say `pp` is not really similar to those functions).
* I'd be wary of changing `pp` in the suggested manner at this point as I have [code that depends on `print` returning `nil`](https://github.com/sogaiu/janet-ref/blob/c6c66fe1b17abc2b3943c3441a6b640a636f9dff/janet-ref/completion.janet#L33-L45) so possibly someone else has done something similar for `pp`.

If there weren't a `tracev` callable though, I'd think it would be a good idea :)

Though I'd like something that could have its behavior customized to do other things like measure execution time.

zevv:
Excellent, thanks!

Issue #1154, opened 2023-05-24
Reorganize test suite


bakpakin:
Janet's test suite is hard to navigate and disorganized. The coverage is fairly broad and good for catching most serious regressions, but it is not obvious where to put new tests - Better organization is needed in the long term.

One way would to be to break up the suites by module.

pepe:
I come to that organization in my codes.

tionis:
I also think that naming tests after modules would be very useful. (In general, so for spork too).
Every time I add a new test, I have to grep to find a fitting suite.

Perhaps an argument could be added to `jpm test` to specify which suite to execute. (Even though it brings no real benefit over janet test/some-module.janet` as far as I can see)

sogaiu:
To get some sense of what might be involved, I tried splitting `suite0000.janet` up.

[Here](https://github.com/sogaiu/janet/commit/2a34a10d971e25573edea3ec6838c38a2700a2ea) is a commit with an attempt.

Some things that happened:

* Changed `helper.janet` so it can work with a string instead of a number (`suite-name` instead of `suite-number`)
* Temporarily named new files `suite-<module-name>.janet` -- this was to work with the existing `Makefile`
* I tried to place tests in a file associated with what I guessed was the most relevant module (e.g. `not` is actually defined in `math.c` so [this line](https://github.com/janet-lang/janet/blob/909c9060809303ddb4533cf841f51e7b8694e266/test/suite0000.janet#L63) and [this line](https://github.com/janet-lang/janet/blob/909c9060809303ddb4533cf841f51e7b8694e266/test/suite0000.janet#L65) end up in `suite-math.janet`) -- but see below for more on these lines.
* Some tests that had lines interleaved got split up -- e.g. there were [some lines for structs and tables that were together](https://github.com/janet-lang/janet/blob/909c9060809303ddb4533cf841f51e7b8694e266/test/suite0000.janet#L313-L328), but as a test (no pun intended), I moved one set of lines to a file for structs and the remaining set of lines to a file for tables
* Changed some names (e.g. [here](https://github.com/janet-lang/janet/blob/909c9060809303ddb4533cf841f51e7b8694e266/test/suite0000.janet#L227) the name `merge` is used, but I think this could potentially shadow an actual use of the built-in `merge` at some point so it seemed better to rename for future-proofing) [1]
[...]

PR #1162, opened 2023-05-27
Reorganize tests

sogaiu:
This is a draft PR related to #1154.

I'll start by saying that I'm fine with this just being an ingredient in figuring out a better organization for tests.  Please don't hesitate to give critical feedback -- the goal of a better test organization is most important here IMO and not the merging of this particular PR.

Now that that's out of the way...

This PR includes the following changes:

* Moved tests out from the `suite00XY.janet` files into `suite-<module>.janet` files, where `<module>` is a name from a file living in `src/core` (no extension), like: `string`, `buffer`, etc.  Two exceptions to this are:
    * `suite-boot.janet` - contains tests most closely associated with `boot.janet`
    * `suite-unknown.janet` - contains tests I did not succeed in fitting in elsewhere (around 5% (c. 60) of the 1858 tests [1])
* `helper.janet` has been modified so that `start-suite` can take a string and further it doesn't require an argument -- `(start-suite)` is enough (see [here](https://github.com/janet-lang/janet/issues/1154#issuecomment-1562512858) for more details).
* A large number of the tests [2] have been annotated with commit hashes (and in a smaller number of cases issue numbers as well) so it's possible to mostly figure out which commit added a test or at least was the origin of a grouping of tests.  This should be helpful in assessing whether a test really belongs in the file it has been put in and further, the remaining ones in `suite-unknown.janet` should be easier to review if there is a desire to move them elsewhere [3].
* Removed all of the `suite00XY.janet` files.
* Changed the names of some identifiers to prevent potential shadowing built-in names (e.g. `merge` -> `merge-sort`)
* Updated some description strings that seemed off
* Tweaked some tests that seemed off
* Changed some formatting (e.g. line lengths) to make comparison easier

I verified the number of tests (at least the `assert`-like ones) did not change via a script that parsed the output of `make test` [4].

---

[1] At the time of writing I (or my janet script) counted 1858 assert-like tests using `helper.janet`.  There are some things in the test files that are not wrapped in `assert`-like forms which I believe will cause testing to stop if those forms / expressions are not successfully evaluated, so such things are "tests" in some sense, even if they don't contribute to the overall count.  I don't think there are that many though.

[2] I want to say "nearly all", but I didn't count :)

[3] Having issue numbers is really handy for establishing a connection between a test and relevant context.  I might try more often to file issues before making PRs (if there isn't already a relevant issue).

[4] It actually caught some that went missing when I tried to split out the network-related stuff from the ev-related tests.  I didn't figure out what was going on there so the net tests still live together with the ev ones.


zevv:
I notice git has been able to figure out the history for some of the split-off files, but the majority of the lines (82%) is now attributed to your commit 0eeab4a62. Is this something we would care about?

sogaiu:
> I notice git has been able to figure out the history for some of the split-off files, but the majority of the lines (82%) is now attributed to your commit https://github.com/janet-lang/janet/commit/0eeab4a62cce78978aac47d748d4eaf073bee528. Is this something we would care about?

I care :)

I think I tracked dow a fair number of the tests, but as this was a manual effort and tests had moved around and merged (before I got to them), it's likely I have missed some / gotten some wrong.  May be there's a more error-free way...

Please don't hold back!

bakpakin:
Looks pretty good, thanks for all of the work!

One cool feature to have, especially now that we have feature specific suites, is to run testing when certain features are disabled - for example, to be able to run `make test` and pass, even if FFI is disabled. Ideally have something that prints a "Skipped suite-ffi" message to stderr but passes so that we can run run testing on mimimized implementations. Currently, we only do compilation testing on small builds which has been less than fun for porters to limited platforms (Nintendo switch, I think there was also an Amiga port).

Something like:

```
...
(start-suite)

(feature-gate 'ffi/context)
...
```
Which would encapsulate this logic.

sogaiu:
Thanks for taking a look.

I've pushed [a commit](https://github.com/janet-lang/janet/pull/1162/commits/f3beb6bdee64edb58b30a4940425a126c0081cba) with a sketch of `feature-gate` and supporting bits.

With those changes, and putting:

```c
#define JANET_NO_FFI
#define JANET_NO_FFI_JIT
```

in `src/conf/janetconf.h`, some of the `make test` output is:

```
Starting suite ev...
Finished suite ev in 0.300 seconds - 495 of 495 tests passed.
Starting suite ffi...
Skipping suite ffi.
Starting suite inttypes...
Finished suite inttypes in 0.005 seconds - 87 of 87 tests passed.
```

---

Regarding other suites...

IIUC, the above `feature-gate` arrangement works for `ffi` partly because essentially each of the tests in `suite-ffi.janet` is wrapped in `(compwhen has-ffi ...)` (where `has-ffi` was defined as `(def has-ffi (dyn 'ffi/native))`).

Almost all of the rest of the tests in the other `suite-*.janet` files doesn't use `compwhen` ATM.

`suite-intypes.janet` (in its current form) is a specific example where the `feature-gate` construct won't work with `#define JANET_NO_INT_TYPES` because syntactically top-level `int/s64` / `int/u64` would not be available (at compilation time) and none of the test content accounts for that.

So to run testing when certain features are disabled in `suite-inttypes.janet` and elsewhere, would it work to just wrap most of the content of the suite file in a single appropriate `compwhen`?

Since `compwhen` upscopes anyway, it seems like this could work -- I think this sort of thing is done in parts of `boot.janet`.  May be this wouldn't require an explicit `feature-gate` construct either.

With that approach I was able to get `suite-inttypes.janet` to cooperate.

sogaiu:
I've removed the `feature-gate` bits and rebased.

Number of tests increased by 1 compared to previously because a new one was added in https://github.com/janet-lang/janet/commit/4782a76bcab1d6ad6af4a1bc821d6b3604d5300b.

sogaiu:
Rebased onto 644ac8caf8e8a1fa5b91d1827769c99d8009d37a.

sogaiu:
Looks like there are some more changes to incorporate :)

sogaiu:
Ok, made an attempt to incorporate changes up through 5de889419ff26b710b706958bf99e180d084f564.

(Note to self: consider using stacked git next time :) )

sogaiu:
Updated to include changes through e97299f.

PR #1164, opened 2023-05-27
More portable method of installing janet.h -> janet/janet.h symlink

dressupgeekout:
The problem that #994 was trying to solve (the symlink should be a relative path and not absolute) doesn't actually require the use of the GNU-only `-t`/`-T` option to ln(1).


sogaiu:
To spell things out a bit...

The original proposed change looks like it was:

```makefile
ln -sf -t '$(DESTDIR)$(INCLUDEDIR)' janet.h janet/janet.h
```

What's currently in the `Makefile` is:

```makefile
ln -sf -T ./janet/janet.h '$(DESTDIR)$(INCLUDEDIR)/janet.h' || true #fixme bsd
```

So it looks like there was a change to the original at some point.

Doing some digging, came across [this](https://github.com/janet-lang/janet/commit/e001efa9fdd6468e9589dfe1dcedb075dcf43048).

Issue #1163, opened 2023-05-27
Why is forv so much faster than for (and should something be done about it)?


primo-ppcg:
A bodyless `forv` loop appears to be over 20% faster than the equivalent `for` loop. This surprised me. Perhaps it's just my experience with other languages, but I would have expected immutable forms to always be faster than mutable forms.

If we expand these loops, we can see why this is:

```janet
(for i 0 100)
# ->
(do (var _000034 0) (def _000035 100) (while (<function <> _000034 _000035) (def i _000034) (set _000034 (<function +> _000034 1))))
```
```janet
(forv i 0 100)
# ->
(do (var i 0) (def _000034 100) (while (<function <> i _000034) (set i (<function +> i 1))))
```

A `for` loop defines a locally scoped variable every single iteration, and therefore has more overhead. My question is: must it necessarily be so?

I don't really have a concrete suggestion. Although, one idea that might work is to add some sort of meta-data to the `var` special form, that exposes its current value, immutably, under another name. Something like a symbolic link. For example:

```janet
(var _000034 0 :alias i)
```

So that when `i` is looked up in the symbol table, it would return the current value of `_000034`, without ever needing to be explicitly set.

primo-ppcg:
So, variable references are captured as references, and not as their current value. Not so surprising, really, just something that the implementer (which, if it's going to be anyone, it's going to be you 😉) would need to keep in mind.

I also think that other forms could benefit from this as well. `eachp` in particular seems like a great candidate for a performance boost.

bakpakin:
So the cleanest way to do this sort of thing is actually just add some simple optimization passes to the bytecode to remove redundant reads - and to do this in a way that doesn't impede debugging.

For example:

```janet
(setdyn *pretty-format* "%.99M")

(pp (disasm |(for i 0 100)))
(pp (disasm |(forv i 0 100)))
```
Outputs the following:

```
{ :arity 0
  :bytecode @[(ldi 0 0) (ldi 1 100) (lt 2 0 1) (jmpno 2 4) (movn 3 0) (addim 0 0 1) (jmp -4) (retn)]
  :constants @[]
  :defs @[]
  :environments @[]
  :max-arity 0
  :min-arity 0
  :slotcount 4
  :source "temp2.janet"
  :sourcemap @[(3 13) (3 13) (3 13) (3 13) (3 13) (3 13) (3 13) (3 13)]
  :structarg false
[...]

PR #1170, opened 2023-05-30
Rework eachp

primo-ppcg:
Removes `:pairs` from the `each-template`, and special cases it to avoid a destructuring assignment each iteration. The result is over 3x faster than the current implementation with an empty body, and further benefits from the in-progress bytecode optimizations.

`loop :pairs` and `seq :pairs` are similarly improved.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -444,9 +444,21 @@
          (def ,binding
            ,(case kind
               :each ~(,in ,ds ,k)
-              :keys k
-              :pairs ~(,tuple ,k (,in ,ds ,k))))
+              :keys k))
          ,;body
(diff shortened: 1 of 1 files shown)


bakpakin:
While a good optimization in most cases, I think this breaks code of the form:

```janet
repl:2:> (eachp p (range 10) (pp p))
(0 0)
(1 1)
(2 2)
(3 3)
(4 4)
(5 5)
(6 6)
(7 7)
(8 8)
(9 9)
nil
repl:3:> 
```

primo-ppcg:
The [proposed compiler changes](https://github.com/janet-lang/janet/issues/1163#issuecomment-1570164692) will perform this optimization automatically, and everywhere else applicable.

PR #1171, opened 2023-05-30
Fixed net/connect binding address

zevv:
`net/connect` uses the destination address instead of the binding address for the `bind()` after the `connect()`


bakpakin:
Interesting, good catch!

PR #1175, opened 2023-06-02
Added JANET_NO_AMALG flag to Makefile 

zevv:
(Not sure if people are willing to pull in this added complexity, but giving it a shot anyway)

This change allows building Janet from the individual source files instead of from the amalgamated `janet.c`; this considerably speeds up parallel builds on modern CPUs;

```
 make clean; time make -j NO_AMALG=0 
[...]
real	0m7.882s
```

vs

```
make clean; time make -j NO_AMALG=1 
[...]
real	0m1.481s
```

For me this make the difference between "comfortably and quickly running my tests" and "getting slightly annoyed staring at my screen",

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ JANET_PATH?=$(LIBDIR)/janet
 JANET_MANPATH?=$(PREFIX)/share/man/man1/
 JANET_PKG_CONFIG_PATH?=$(LIBDIR)/pkgconfig
 JANET_DIST_DIR?=janet-dist
+JANET_BOOT_FLAGS:=. JANET_PATH '$(JANET_PATH)'
+JANET_TARGET_OBJECTS=build/janet.o build/shell.o
 JPM_TAG?=master
 DEBUGGER=gdb
 SONAME_SETTER=-Wl,-soname,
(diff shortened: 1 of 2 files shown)


sogaiu:
The speedup is nice :+1:

bakpakin:
The added complexity is ok, since it is not much added complexity. We used to have this, however the requirement for distributing as `janet.c` and `janet.h` meant that sometimes the amalgamation would break, so I made the amalgamation the default in the Makefile which reduced that problem.

But a lot has changed and stabilized, I guess I just got use to waiting for the compiler :sweat_smile:

disruptek:
It would probably be sufficient to make the amalgamation a build/ci artifact, right?

bakpakin:
Yes, but the idea was to have one fewer way to build things and I would get PRs that would break the amalgamated build (and break them myself). Basically, it was just one less thing to think about. Still, it's like a breath of fresh air to have incremental compilation nice and fast again.

zevv:
Is the current implementation acceptable, in particular the way the option is passed to and handled in boot. janet? Also maybe change NO_AMALG into something nicer like JANET_DISABLE_AMALGAMATION? (I will not be typing that more then once (in my .bashrc only)

sogaiu:
How about `JANET_NO_AMALG`?

bakpakin:
Current implementation looks good, but it is probably better to have a name that is unlikely to collide with anything as the trigger to avoid surprises.

zevv:
env var  changed to `JANET_NO_AMALG`

CosmicToast:
I'm curious on what the speed delta would look like when running with a more parallelized linker. (e.g mold)

zevv:
Linking really is not the problem, that's typically done in tens of milliseconds for a project this size.

sogaiu:
May be a minor thing but I think the construction `(if-not (...) (do ...` in the following:

```janet
    (if-not (has-value? boot/args "image-only") (do
      (print "\n/* " fname " */")
      ...)
```

[here](https://github.com/janet-lang/janet/commit/bb4ff05d35fdd01dbd4cd61ae9740c01830d6d86#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR4160-R4164) might be better expressed as either:

```janet
    (when (not (has-value? boot/args "image-only"))
      (print "\n/* " fname " */")
      ...)
```

or:

```janet
    (unless (has-value? boot/args "image-only")
      (print "\n/* " fname " */")
      ...)
```

The latter (using `unless`) is briefer, but I find it often takes me much longer to convince myself at a gut level that it's the "right thing" particularly when what's getting tested gets complicated.

(I've sometimes wanted a `when-not`, but I have been hesitant to suggest that kind of construction for potential bloat reasons (^^; )

zevv:
Oh sorry @sogaiu, I saw your earlier remark and changed the code accordingly; this was on "my other machine" though and I forgot to push it.

sogaiu:
Ah well, at least the code works :)

PR #1177, opened 2023-06-02
added os/isatty, do not enable colors if stdout is not a tty

### LINKED ISSUE #1176: Turn off ANSI colors by default if output is not a terminal
I know that this can be achieved with `-n`, but most other command line tools only enable color output if writing to a terminal and disable it otherwise.

It it possible to make janet adhere to this common practice?

zevv:
Fixes #1176

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3903,7 +3903,9 @@
 
   (if-let [jp (getenv-alias "JANET_PATH")] (setdyn *syspath* jp))
   (if-let [jprofile (getenv-alias "JANET_PROFILE")] (setdyn *profilepath* jprofile))
-  (set colorize (not (getenv-alias "NO_COLOR")))
+  (set colorize (and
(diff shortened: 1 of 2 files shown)


zevv:
would't be surprised if this breaks color on windows again :)

zevv:
updated to use `_fileno()` on win32

sogaiu:
I'd like to verify it works on Windows but am coming up short on how.

For what I've tried with `janet.exe` without entering the REPL, I don't get color (with or without these changes).  Similar on Linux.

Any hints?

zevv:
t

sogaiu:
Thanks, that helped :)

I built https://github.com/janet-lang/janet/pull/1177/commits/e0ea844d50690ea54c4faa02e34a51f7176fbcc9 on Windows and Linux and the changes seem to work as advertised :+1: 

No breakage of color on Windows AFAICT.

zevv:
cool, thanks for verifying.

PR #1182, opened 2023-06-07
Fix resumption values when closing a channel

chris-chambers:
When suspended in `ev/give` or `ev/take`, closing the channel should cause the result of `ev/give` or `ev/take` to be `nil`. 
[ref](https://janet-lang.org/api/ev.html#ev/chan-close)

When suspended in `ev/select`, closing the channel should cause the result of `ev/select` to be `[:close ch]`.
[ref](https://janet-lang.org/api/ev.html#ev/select)

The results were flipped before.


bakpakin:
Interesting, was this a regression? Either way, tests look good.

chris-chambers:
I looked through the history a little to see if it was a regression, but it looks like it's been there all along.  Definitely surprising.  Must not be many people using channels that way.

Issue #1178, opened 2023-06-02
The asymmetry of drop and take


primo-ppcg:
The current implementations of `drop` and `take` are asymmetric:

`drop`
Allows dropping from the end.
Accepted types: bytes, indexed

`take`
Does not allow taking from the end.
Accepted types: bytes, indexed, dictionary, fiber, any other data type for which `next` is valid (aside: can we agree to call these all `iterable?`)

This is more of a request for comments than an open issue, although I do think that these functions should be symmetrically defined. My thoughts are the following:

1. For data types with a clearly defined end (bytes or indexed), taking or dropping from the end is a sensible thing to want to do, and should be allowed.
2. For data types without a clearly defined end (dictionary or fiber), taking or dropping from the end is not a sensible thing to want to do and should either:
  a. return the argument unmodified (`drop`), return an empty array (`take`)
  b. produce an error.
3. Taking or dropping from the front is a sensible thing to want to do for any data type for which `next` is valid, and should be allowed.

I could be swayed on any of these points (with exception to perhaps point 3), although as stated previously, I do think that these functions should be symmetrically defined.

primo-ppcg:
> If you consider the dictionary to have a clearly defined front, it also has a clearly defined end, and vice versa.

I don't disagree, necessarily. Except that the core library defines `next` (and `next nil` as first), but does not define `last` or `previous`. As a result, a dictionary types can be easily iterated front to back, in some perhaps arbitrary order, but not back to front. And because this front to back ordering exists, it is sensible to allow it for  both drop and take.

> I'm curious, what exactly is the utility of dropping from the front on a dictionary?

I don't know that either of these functions must work with dictionary types. However, a consistent behavior between `take n` and `drop n` could definitely be useful in the implementation of various algorithms.

primo-ppcg:
To elaborate a bit more on what I mean by asymmetry, if `take` were defined analogously to `drop`, the following assertions would hold in any applicable case:

```janet
(defn take
  ``Take the first `n` elements in an indexed or bytes type. Returns a new tuple or string
  instance, respectively. If `n` is negative, takes the last `n` elements instead.``
  [n ind]
  (def use-str (bytes? ind))
  (def f (if use-str string/slice tuple/slice))
  (def len (length ind))
  (def nonnegn (>= n 0))
  (def start (if nonnegn 0 (max 0 (+ len n))))
  (def end (if nonnegn (min n len) len))
  (f ind start end))

[...]

PR #1183, opened 2023-06-08
Make take and drop more symmetric

primo-ppcg:
Related issue: #1178 

Allow `take` from the end of bytes or indexed (as `drop` does).
Allow `drop` from fibers (as `take` does).
Return table for `take` from dictionary types.
Allow `drop` from dictionary types (as `take` does).
Increase efficiency for `take` and `drop` with slices.
Check indexed types before bytes.

It isn't clear to me that there is a common use case for taking and dropping from dictionary types. However, the implementation is relatively straight-forward, particularly for take, so I don't know that it should be explicitly forbidden either.

`take` for indexed types isn't slowed any by this change, and `drop` has become significantly faster. `take` has become slightly slower for `bytes` objects due to the type check reordering. I believe that indexed types are a more common case, and in particular when efficiency is a concern.

```janet
(def ind (range 200))
(repeat 10_000_000 (take 100 ind))
master     8.399327
take-drop  7.313185

(repeat 10_000_000 (drop 100 ind))
master    11.91337
take-drop  6.92769

(repeat 1_000_000 (take-until |(> $ 100) ind))
master     7.40999
take-drop  7.256637

(repeat 1_000_000 (drop-until |(> $ 100) ind))
master     7.251279
take-drop  7.15903
```

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1083,9 +1083,9 @@
 (defn- take-n-slice
   [f n ind]
   (def len (length ind))
-  (def negn (< n 0))
-  (def start (if negn (max 0 (+ len n)) 0))
-  (def end (if negn len (min n len)))
+  (def m (+ len n))
+  (def start (if (< n 0 m) m 0))
+  (def end (if (<= 0 n len) n len))
   (f ind start end))
 
 (defn take
@@ -1093,8 +1093,8 @@
   respectively. If `n` is negative, takes the last `n` elements instead.``
   [n ind]
   (cond
-    (bytes? ind) (take-n-slice string/slice n ind)
     (indexed? ind) (take-n-slice tuple/slice n ind)
+    (bytes? ind) (take-n-slice string/slice n ind)
     (dictionary? ind) (do
                         (var left n)
                         (tabseq [[i x] :pairs ind :until (< (-- left) 0)] i x))
@@ -1117,8 +1117,8 @@
   "Same as `(take-while (complement pred) ind)`."
   [pred ind]
   (cond
-    (bytes? ind) (take-until-slice string/slice pred ind)
     (indexed? ind) (take-until-slice tuple/slice pred ind)
+    (bytes? ind) (take-until-slice string/slice pred ind)
     (dictionary? ind) (tabseq [[i x] :pairs ind :until (pred x)] i x)
     (seq [x :in ind :until (pred x)] x)))
 
@@ -1131,10 +1131,10 @@
 (defn- drop-n-slice
   [f n ind]
   (def len (length ind))
-  (def negn (< n 0))
-  (def start (if negn 0 (min n len)))
-  (def end (if negn (max 0 (+ len n)) len))
-  (f ind start end))
+  (cond
+    (<= 0 n len) (f ind n)
+    (< (- len) n 0) (f ind 0 (+ len n))
+    (f ind 0 0)))
 
 (defn- drop-n-dict
   [f n ind]
@@ -1148,8 +1148,8 @@
   instance, respectively. If `n` is negative, drops the last `n` elements instead.``
   [n ind]
   (cond
-    (bytes? ind) (drop-n-slice string/slice n ind)
     (indexed? ind) (drop-n-slice tuple/slice n ind)
+    (bytes? ind) (drop-n-slice string/slice n ind)
     (struct? ind) (drop-n-dict struct/to-table n ind)
     (table? ind) (drop-n-dict table/clone n ind)
     (do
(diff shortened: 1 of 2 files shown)


sogaiu:
I tried the code out.  Below are some preliminary remarks.

FWIW, I did something like:

```janet
(def ind (range 200))

## take

(def take-results @[])

(for i 0 10
  (def start (os/clock)) 
  (repeat 10_000_000 (take 100 ind))
  (array/push take-results
              (- (os/clock) start)))
```

So I could see individual batch run results.

For `take` and `drop` I got similarish looking results.

However, for `take-until` and `drop-until`, the results were not so clearly different.  One thing I noticed in the results was that there could sometimes be a "spike" result or a result beyond which things took more time consistently, perhaps this was some throttling kicking in.

If I throw out outlier-ish things `take-until` looks slightly faster on the `take-drop` branch, but `drop-until` I couldn't really tell.

In any case, the point of this PR is symmetry (though of course it's nice to get a boost for some cases and/or maintain a similar level of efficiency for others).

---

A rather minor thing -- there was a change that [got lost in one of zevv's recent PRs](https://github.com/janet-lang/janet/pull/1175#issuecomment-1575785211) and it looks like the same lines are touched here so I wonder if we might make a change: https://github.com/janet-lang/janet/pull/1183/commits/472ec730b5c22470535f1fcc707915172ad2a864#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR4171-R4172

If you don't mind, may be we can go for one of the suggested options.  No big deal though, just noticed a chance to tweak :)

sogaiu:
If we're going the route of handling dictionaries [1] for `drop-until`, I wonder if it's worth the predicate getting access to both the key and value per encountered pair.  At the moment it looks like to me like [the predicate only gets access to the value](https://github.com/janet-lang/janet/pull/1183/commits/b5407ac708ff38b30655423c9ab1761b7d3a2214#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR1171).

Not sure though, just bringing it up.

If so, I guess it would make sense to do similarly for `take-until`: https://github.com/janet-lang/janet/pull/1183/commits/b5407ac708ff38b30655423c9ab1761b7d3a2214#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR1122

If those changes were made then `take-while` and `drop-while` are also affected, but I guess that doesn't involve any code changes.

---

[1] I'm in favor of leaving the functionality undocumented initially :)

primo-ppcg:
I appreciate the feedback!

`drop` became significantly slower after #1114, which is understandable because it added extra functionality. My primary goal was to win back some of the lost performance, and also not to slow `take` any.

`drop-until` and `take-until` are essentially the same implementation copy-pasted, just restructured slightly. My point in including them is to demonstrate that they haven't been slowed by the refactoring. On my system, both are approximately 0.1µs per call  faster for indexed types, and 0.05µs slower per call for bytes, pretty consistently. This is due primarily to the re-ordering of the checks.

For profiling, I use something similar to this: [timeit.janet](https://gist.github.com/primo-ppcg/ccf7b153b2bc970900db9089699bdf71)
```janet
(defmacro timeit
  ``Similar to `loop`, but outputs performance statistics after completion.``
  [head & body]
  (with-syms [clk cnt elp run]
    ~(do
       (var ,cnt 0)
       (def ,clk (os/clock))
       (loop ,head (++ ,cnt) ,;body)
       (def ,elp (- (os/clock) ,clk))
       (def ,run (/ ,elp ,cnt))
       (cond
         (< ,run 1e-3) (printf "elapsed %gs, %.4gµs/body" ,elp (* ,run 1_000_000))
         (< ,run 1) (printf "elapsed %gs, %.4gms/body" ,elp (* ,run 1_000))
         (printf "elapsed %gs, %.4gs/body" ,elp ,run)))))

(def ind (range 200))
(timeit [:repeat 10_000_000]
  (drop 100 ind))
```
---

> A rather minor thing -- there was a change that [got lost in one of zevv's recent PRs](https://github.com/janet-lang/janet/pull/1175#issuecomment-1575785211) and it looks like the same lines are touched here

I remembered to run my changes through `format-file` (unlike [last time](https://github.com/janet-lang/janet/commit/c4c86f8671d5fea60e6461816b2018d9caf8d367)). Should I revert those lines?

primo-ppcg:
> If we're going the route of handling dictionaries [1] for `drop-until`, I wonder if it's worth the predicate getting access to both the key and value per encountered pair. At the moment it looks like to me like [the predicate only gets access to the value](https://github.com/janet-lang/janet/pull/1183/commits/b5407ac708ff38b30655423c9ab1761b7d3a2214#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR1171).
> 
> Not sure though, just bringing it up.

All other functions that accept a predicate (`find`, `filter`, etc.) treat dictionary types the same as any other, only passing the value to the predicate. For consistency, these should do the same.

PR #1190, opened 2023-06-11
update if-let

### LINKED ISSUE #1189: if-let behavior appears different in #1188
The following code used to not error for me:

```janet
(if-let [[l r] (peg/match "a" "b")]
  :a
  :b)
```

With https://github.com/janet-lang/janet/commit/2c3ca2984ec731f9804abee1f6b52031bbbd62a4 I get:

```
$ janet
Janet 1.28.0-dev-2c3ca298 linux/x64/gcc - '(doc)' for help
repl:1:> (if-let [[l r] (peg/match "a" "b")]
repl:2:(>   :a
repl:3:(>   :b)
error: expected string, symbol, keyword, array, tuple, table, struct or buffer, got nil
```

With https://github.com/janet-lang/janet/commit/94722e566c6f9bab9b0c8e630b0fd1aed4e10739 I get:

```
$ ./build/janet
Janet 1.28.0-dev-94722e56 linux/x64/gcc - '(doc)' for help
repl:1:> (if-let [[l r] (peg/match "a" "b")]
repl:2:(>   :a
repl:3:(>   :b)
:b
```

### LINKED ISSUE #1191: if-let's else branch appears multiple times in its expansion
This is unrelated to the recent `if-let` changes -- `if-let` has always worked this way, but I found this very surprising:

```
repl:1:> (defmacro hello [] (print "hello") 1)
<function hello>
repl:2:> (if-let [a true] "ok" (hello))
hello
"ok"
repl:3:> (if-let [a true b true] "ok" (hello))
hello
hello
"ok"
repl:4:> (if-let [a true b true c true] "ok" (hello))
hello
hello
hello
"ok"
```

I would expect the else branch to only appear once in the expansion of `if-let` -- which I believe is not compatible with the desire to be able to reference the first binding form in the else branch.

primo-ppcg:
Fixes #1189

Code of the following form should not error:
```janet
(if-let [[a b] (possibly-nil-or-tuple)] ...)
```
in accordance with the documentation: _If any are false or nil, evaluate the fal form._

Fixes #1191

Macros with side-effects should not be expanded more than once. This is an issue that pre-dates the most recent change.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -668,18 +668,20 @@
   (def len (length bindings))
   (if (= 0 len) (error "expected at least 1 binding"))
   (if (odd? len) (error "expected an even number of bindings"))
+  (def res (gensym))
   (defn aux [i]
     (if (>= i len)
-      tru
+      ~(do (set ,res ,tru) true)
       (do
         (def bl (in bindings i))
         (def br (in bindings (+ 1 i)))
         (if (symbol? bl)
-          ~(if (def ,bl ,br) ,(aux (+ 2 i)) ,fal)
+          ~(if (def ,bl ,br) ,(aux (+ 2 i)))
           ~(if (def ,(def sym (gensym)) ,br)
-             (do (def ,bl ,sym) ,(aux (+ 2 i)))
-             ,fal)))))
-  (aux 0))
+             (do (def ,bl ,sym) ,(aux (+ 2 i))))))))
+  ~(do
+     (var ,res nil)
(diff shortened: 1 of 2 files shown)


primo-ppcg:
The use of `symbol?` should possibly be replaced with `idempotent?`, pending the resolution of #1192.

sogaiu:
@primo-ppcg I gave 2a7ea27b a try.

The 20 or so repositories I run tests against are now back to having their tests pass :+1:

Issue #1192, opened 2023-06-11
symbols are not idempotent


primo-ppcg:
Shouldn't they be?

In the "Prepare for 1.27.0" commit, `idempotent?` was refactored, with `:symbol` being added to the list of `non-atomic-types`:

https://github.com/janet-lang/janet/commit/01aab666676c5f93529ba8db5d97b9db1f4c499a#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbcc

I think this may have been inadvertent. Throughout the core library, this function is used to determine if a gensym is necessary to prevent a form from being evaluated twice.

As a result, we have more complicated expansions than necessary:

```janet
(def a 1)
(def b 2)
(macex ~(or a b))
# ->
(do (def _00000r a) (if _00000r _00000r b))
```
where
```janet
(if a a b)
```
would clearly be sufficient.

bakpakin:
So yes, the issue is fundamentally  that symbols are only idempotent in certain contexts (we know that the symbol is not redefined or mutated between all evaluations of the symbol)

We probably could have a separate (possibly internal) macro with such a constraint and use it there.

primo-ppcg:
> Currently, this prints `0123456789`. If I reimplement `loop` so that symbols are considered idempotent, it prints `0259`.

I've confirmed that this was the behavior in 1.26.0, that is, until just a few months ago. One or the other is not obviously more correct to me.

primo-ppcg:
`idempotent?` is used to answer several different questions throughout the core library.

As it relates to symbols specifically:
1. Q: Can this appear multiple times in an expansion without potentially causing side-effects? (which includes `and`, `or`, `case`, `match`, `each` and any other place where these are used)
A: Yes.
2. Q: Does this represent an atomic binding target? (which includes `if-let`, `when-let`, `if-with`, `when-with` and any other place where these are used)
A: Yes.
3. Q: Is this known not to change throughout subsequent iterations of a user provided body? (which very specifically applies to the `step` argument of a range iterator, and nowhere else as far as I have been able to determine)
A: No.

For case 2., `symbol?` could be used instead (in fact [I've done so as a work-around](https://github.com/janet-lang/janet/pull/1190/files#diff-e120880268b4f0f04177470180f50ee0d2c7ac13cb83bb778b6d81efda1cbbccR678)). For case 3., `number?` could be used instead if it's really a concern.

[...]

PR #1195, opened 2023-06-17
Move idempotent to corelib

primo-ppcg:
Related issue: #1192 

Moves `idempotent?` from `boot.janet` to `corelib.c`. This is done so that `var` bindings and `def` bindings can be distinguished within macros.

```janet
(defmacro macro-idempotent
  [x]
  (idempotent? x))

(var var-x 1)
(assert (= (macro-idempotent var-x) false) "macro idempotent var")

(def def-x 1)
(assert (= (macro-idempotent def-x) true) "macro idempotent def")
```

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -129,19 +129,6 @@
 # For macros, we define an imcomplete odd? function that will be overriden.
 (defn odd? [x] (= 1 (mod x 2)))
 
-(def- non-atomic-types
-  {:array true
-   :tuple true
-   :table true
-   :buffer true
-   :symbol true
-   :struct true})
-
-(defn idempotent?
-  "Check if x is a value that evaluates to itself when compiled."
-  [x]
-  (not (in non-atomic-types (type x))))
-
 # C style macros and functions for imperative sugar. No bitwise though.
 (defn inc "Returns x + 1." [x] (+ x 1))
 (defn dec "Returns x - 1." [x] (- x 1))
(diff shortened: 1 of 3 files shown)


sogaiu:
I tried it out a bit.

The tests I mentioned [here](https://github.com/janet-lang/janet/pull/1190#issuecomment-1586408806) all passed at least :)

ianthehenry:
This doesn't quite do what you want, because you're only looking at the symbol in the fiber's environment, which is not necessarily the scope that the macro will be expanded in. Consider:

```
Janet 1.28.0-dev-2e2d792e macos/aarch64/clang - '(doc)' for help
repl:1:> (idempotent? 'x)
true
repl:2:> (var x 1)
1
repl:3:> (idempotent? 'x)
false
```

So far so good. But when you make non-top-level bindings:

```
repl:4:> (let [x 1] (idempotent? 'x))
false
repl:5:> (do (var y 1) (idempotent? 'y))
true
```

You get unexpected results.

I don't think there's a way to write the thing that you want in general. `idempotent?` is not really a function of a symbol by itself, it's a function of a symbol and a scope. But macros aren't aware of the lexical environment into which they're being expanded -- Janet doesn't have "first-class scopes," apart from the fiber environment, which is its own special thing.

primo-ppcg:
Thanks for taking the time to review.

I think this optimization would need to happen in the compiler itself (remove def bindings of def bindings).

primo-ppcg:
Thanks for taking the time to review.

I think this optimization would need to happen in the compiler itself (remove def bindings of def bindings).

PR #1203, opened 2023-06-25
Add C escape sequences

czkz:
Some C escape sequences are not supported in Janet strings, namely:
- `\'`
- `\?`
- `\a`
- `\b`

Full list at https://en.cppreference.com/w/c/language/escape.
Supporting them could be useful to copy-paste strings from other languages (and I need `\b`).

### DIFF
--- a/src/core/parse.c
+++ b/src/core/parse.c
@@ -259,6 +259,14 @@ static int checkescape(uint8_t c) {
             return '\f';
         case 'v':
             return '\v';
+        case 'a':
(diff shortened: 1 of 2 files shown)


ml-2:
You probably already know this, but you can use `\x08` instead of `\b` if you need it right now in Janet.

bakpakin:
We don't really need things like `\?` and `\'` but I suppose they don't hurt for copy pasting.

Issue #1206, opened 2023-06-29
floored division, variadic mod


primo-ppcg:
### Consistent modular arithmetic
A modular arithmetic is said to be consistent if the following property holds for any _a_ and non-zero _b_:
```janet
(= a (+ (* (div a b) b) (mod a b)))
```
where `div` represents an appropriate integer division. From the C99 standard onward, modular arithmetic is consistent, and therefore also with Janet int types:
```janet
(for a (int/s64 -99) (int/s64 100)
  (for b (int/s64 -99) (int/s64 100)
    (unless (compare= b 0)
      (assert (= a (+ (* (/ a b) b) (% a b)))))))
```
I find this usage to be intuitive. What isn't so intuitive is when using the default numeric type:
```janet
(def quo (comp math/trunc /))
(def div (comp math/floor /))
(for a -99 100
  (for b -99 100
    (unless (= b 0)
      (assert (= a (+ (* (quo a b) b) (% a b))))
      (assert (= a (+ (* (div a b) b) (mod a b)))))))
```
in particular that `%` is consistent with truncated division may not be expected. Defining integer division in this way is also very slow. It can also be defined like this:
```janet
(defn div [a b] (math/floor (/ a b)))
```
which is a lot faster, but no longer variadic. I've found it's best to inline `math/floor` everywhere, which can get a bit messy. I think it would be useful to add a floored division (`div`) which is consistent with `mod`. I don't know that it's necessary to have a symbolic truncated division that agrees with `%`; in most cases, you would want to be using `div` and `mod` anyway, or at very least it would be sufficient.

### Mod zero
Modulo zero is a well-defined mathematical operation. Two values _a_ and _b_ are said to be equivalent _mod m_ if there exists some integer _k_ for which _a = b + k·m_. To state that _a_ is equivalent to _b mod 0_ is to state that _a_ is equal to _b_, and nothing more. Programming languages that care very much for mathematical correctness (APL, J, K, others) all agree that _x mod 0 = x_, because it is. [Even Donald Knuth agrees](https://books.google.co.th/books?id=x9AsAwAAQBAJ&pg=PA15). I'm not saying that Janet must necessarily do this (and given that the default numeric type is IEEE floating point, very probably _shouldn't_), but I am saying that it is correct, and relevant to the next point.

### Variadic mod
[...]

PR #1207, opened 2023-06-30
floor div, variadic mod

primo-ppcg:
Related issue: #1206

- Adds floored division to corelib (`div`)
- Makes `mod` and `%` variadic
- Defines `(mod x 0)` as `x`, `(% x 0)` is unchanged


sogaiu:
I wanted to bring up the issue of where tests related to `int/*64` ought to live.

When working on the reorganization of tests recently, I made some changes in that direction, namely to place some things with `int/s64` or `int/u64` in `suite-inttypes.janet`.  This was motivated by [this comment](https://github.com/janet-lang/janet/pull/1162#issuecomment-1565584655) having to do with (eventually?) arranging things so that we can:

> run testing when certain features are disabled - for example, to be able to run make test and pass, even if FFI is disabled.

I don't think I got everything relevant moved in to `suite-inttypes.janet` so if this is a direction we want to go in, there are still some changes to be made.

Perhaps the comments above belong in a separate issue, but if there are to be tests in this PR that use `int/*64` perhaps it would be worth thinking about where they should go.  If it's not too much trouble may be they can go in `suite-inttypes.janet`?

primo-ppcg:
> I don't think I got everything relevant moved in to `suite-inttypes.janet` so if this is a direction we want to go in, there are still some changes to be made.

I'll be adding a few more tests to this suite. While I'm there, I can look into moving any remaining tests as well.

sogaiu:
Thanks for the test moving / shielding in [b3db367](https://github.com/janet-lang/janet/pull/1207/commits/b3db367ae724bb9aa94a870937d266000729794c).

Looking over the proposed changes in 8a62c74, I was reminded of something @pyrmont mentioned recently about how [`int/64` and `int/u64` do not provide "NOT"](https://github.com/janet-lang/janet/blob/b125cbeac9e43ca7adb3f066fef679ac58a91c71/src/core/inttypes.c#L549-L597).  I didn't figure out whether there is some reason it isn't doable.  Perhaps it's something that should have a separate issue.

primo-ppcg:
> I was reminded of something @pyrmont mentioned recently about how [`int/64` and `int/u64` do not provide complement / not](https://github.com/janet-lang/janet/blob/b125cbeac9e43ca7adb3f066fef679ac58a91c71/src/core/inttypes.c#L549-L597).

I can't immediately think of a reason why `bnot` shouldn't be supported by int types... although if I were to attempt to implement it, I suspect I may quickly discover the reason 😉

Firstly, the vm asserts that the type is `:number` rather than deferring to abstract implementations:
https://github.com/janet-lang/janet/blob/b125cbeac9e43ca7adb3f066fef679ac58a91c71/src/core/vm.c#L731-L736
I _think_ that's all that would need to change, apart from the implementation in `inttypes`, of course.

Definitely as separate PR, though.

Issue #1219, opened 2023-07-12
Negative index in array.remove off by one error?


narimiran:
Is this intended? (Because definitely it isn't _expected_)

```janet
(array/remove @[10 20 30 40 50] -1) # doesn't remove anything
(array/remove @[10 20 30 40 50] -2) # removes last element
```

This `+ 1` in array.c looks suspicious:
https://github.com/janet-lang/janet/blob/3a4d56afca2eee0e4dbca9a6400757d198e6d6f3/src/core/array.c#L300

pepe:
Check the doc here https://janet-lang.org/api/array.html#array/slice please

sogaiu:
I think what @pepe might have been hinting at is that this is expected behavior.  The negative index convention in Janet might be different from what one might expect from elsewhere but glancing at what has been posted above, they look consistent with what I'm used to.

See [here](https://janetdocs.com/array%2fremove) for some examples.

narimiran:
> Check the doc here https://janet-lang.org/api/array.html#array/slice please

Ok, it seems it is intended. But then it should be mentioned and emphasized in the documentation of every function that uses that convention, because Janet is the only language I know to use `-2` for the last element, and I'm sure I'm not the only one who was surprised with it.

primo-ppcg:
I'm going to throw my hat in the ring, and say that the behavior of `array/remove` is probably wrong, and certainly doesn't follow the principle of least surprise.

On the other hand, I think that the behavior of `array/slice` is correct, however the docstring could, and probably should be reworded:
```diff
- Note that index -1 is synonymous with index `(length arrtup)` to allow a full negative slice range.
+ Note that if the range is negative, it is taken as (start, end] to allow a full negative slice range.
```
This defines the same behavior, without need to define index -1 to be "outside" the array, which doesn't make sense for any other function.

narimiran:
> Check the doc here https://janet-lang.org/api/array.html#array/slice please

Now that I thought about it a bit more — I don't buy it.

Slice can be used without `end`, so there's no need for this "off by one" so that `-1` is the length of the array.

```janet
(array/slice @[10 20 30 40 50] 1)
(array/slice @[10 20 30 40 50] -2)
```

No need to do:
```janet
(array/slice @[10 20 30 40 50] 1 -1)
(array/slice @[10 20 30 40 50] -2 -1)
```

pepe:
@narimiran, you do not need to, indeed. But it is like that, and I do not think it will be changed.

[...]

PR #1224, opened 2023-07-14
`array/remove`: update final array index to be -1

primo-ppcg:
Related issue: #1219

The problem with array index -1: Because `array/slice` does not include its end index in the result, it is not possible to slice to the end of the array. Also, because `array/insert` inserts before the given index, it is not possible to insert to the end of the array.

There are two possible solutions to this problem:
1. (Current Solution) Redefine the final index of an array to be -2.
2. (Proposed Solution) Redefine the behavior of `array/slice` and `array/insert` for negative indices.

Notably, the behavior of these two functions is identical for either solution.

I submit for consideration that the proposed solution is more correct and desirable. When indexing from the end of the array, the array should be viewed reversed, as if its end were its front. In other words, any action that would have occurred before the index should occur after the index, and vice versa.

Furthermore, the current solution breaks the symmetry between `array/insert` and `array/remove`:

Current solution:
```janet
(def a @[1 2 3 4 5])
(array/insert a -1 0) # -> @[1 2 3 4 5 0]
(array/remove a -1)   # -> @[1 2 3 4 5 0]

(def b @[1 2 3 4 5])
(array/insert b -2 0) # -> @[1 2 3 4 0 5]
(array/remove b -2)   # -> @[1 2 3 4 0]
```

Proposed solution:
```janet
(def a @[1 2 3 4 5])
(array/insert a -1 0) # -> @[1 2 3 4 5 0]
(array/remove a -1)   # -> @[1 2 3 4 5]

(def b @[1 2 3 4 5])
(array/insert b -2 0) # -> @[1 2 3 4 0 5]
(array/remove b -2)   # -> @[1 2 3 4 5]
```

Inserting at an index, and then immediately removing at the same index should certainly be a no-op.

### DIFF
--- a/src/core/array.c
+++ b/src/core/array.c
@@ -211,7 +211,7 @@ JANET_CORE_FN(cfun_array_slice,
               "Takes a slice of array or tuple from `start` to `end`. The range is half open, "
               "[start, end). Indexes can also be negative, indicating indexing from the "
               "end of the array. By default, `start` is 0 and `end` is the length of the array. "
-              "Note that index -1 is synonymous with index `(length arrtup)` to allow a full "
+              "Note that if the range is negative, it is taken as (start, end] to allow a full "
               "negative slice range. Returns a new array.") {
     JanetView view = janet_getindexed(argv, 0);
     JanetRange range = janet_getslice(argc, argv);
@@ -259,8 +259,8 @@ JANET_CORE_FN(cfun_array_insert,
               "(array/insert arr at & xs)",
               "Insert all `xs` into array `arr` at index `at`. `at` should be an integer between "
               "0 and the length of the array. A negative value for `at` will index backwards from "
-              "the end of the array, such that inserting at -1 appends to the array. "
-              "Returns the array.") {
+              "the end of the array, inserting after the index such that inserting at -1 appends to "
+              "the array. Returns the array.") {
     size_t chunksize, restsize;
     janet_arity(argc, 2, -1);
     JanetArray *array = janet_getarray(argv, 0);
@@ -297,7 +297,7 @@ JANET_CORE_FN(cfun_array_remove,
     int32_t at = janet_getinteger(argv, 1);
     int32_t n = 1;
     if (at < 0) {
-        at = array->count + at + 1;
+        at = array->count + at;
     }
     if (at < 0 || at > array->count)
         janet_panicf("removal index %d out of range [0,%d]", at, array->count);
--- a/src/core/string.c
+++ b/src/core/string.c
@@ -175,8 +175,9 @@ JANET_CORE_FN(cfun_string_slice,
               "Returns a substring from a byte sequence. The substring is from "
               "index `start` inclusive to index `end`, exclusive. All indexing "
               "is from 0. `start` and `end` can also be negative to indicate indexing "
-              "from the end of the string. Note that index -1 is synonymous with "
-              "index `(length bytes)` to allow a full negative slice range. ") {
+              "from the end of the string. Note that if `start` is negative it is "
+              "exclusive, and if `end` is negative it is inclusive, to allow a full "
+              "negative slice range.") {
     JanetByteView view = janet_getbytes(argv, 0);
     JanetRange range = janet_getslice(argc, argv);
     return janet_stringv(view.bytes + range.start, range.end - range.start);
--- a/src/core/tuple.c
+++ b/src/core/tuple.c
@@ -69,9 +69,9 @@ JANET_CORE_FN(cfun_tuple_slice,
               "inclusive to index `end` exclusive. If `start` or `end` are not provided, "
               "they default to 0 and the length of `arrtup`, respectively. "
               "`start` and `end` can also be negative to indicate indexing "
-              "from the end of the input. Note that index -1 is synonymous with "
-              "index `(length arrtup)` to allow a full negative slice range. "
-              "Returns the new tuple.") {
+              "from the end of the input. Note that if `start` is negative it is "
+              "exclusive, and if `end` is negative it is inclusive, to allow a full "
+              "negative slice range. Returns the new tuple.") {
     JanetView view = janet_getindexed(argv, 0);
     JanetRange range = janet_getslice(argc, argv);
     return janet_wrap_tuple(janet_tuple_n(view.items + range.start, range.end - range.start));
--- a/test/suite-array.janet
+++ b/test/suite-array.janet
@@ -44,7 +44,7 @@
 (assert (deep= (array/remove @[1 2 3 4 5] 2) @[1 2 4 5]) "array/remove 1")
 (assert (deep= (array/remove @[1 2 3 4 5] 2 2) @[1 2 5]) "array/remove 2")
 (assert (deep= (array/remove @[1 2 3 4 5] 2 200) @[1 2]) "array/remove 3")
-(assert (deep= (array/remove @[1 2 3 4 5] -3 200) @[1 2 3]) "array/remove 4")
+(assert (deep= (array/remove @[1 2 3 4 5] -2 200) @[1 2 3]) "array/remove 4")
 
 
 # array/peek


zevv:
@narimiran: say what you want, but the Janet folks are pretty reasonable at times, eh? :)

narimiran:
> @narimiran: say what you want, but the Janet folks are pretty reasonable at times, eh? :)

Hey there, long time no see :)
Believe it or not: when I saw your name in the contributors' list, that's what finally convinced me I should give Janet a try.

----

Hopefully this change doesn't break too much code in the wild, but IMO it is well worth changing this.

sogaiu:
While I don't personally mind rewriting my own code (if I have anything that is affected), if there is rewritten code, is there not an issue of how to handle running with older versions of Janet vs newer versions?

I suppose one might provide multiple versions of one's program that are tailored to different versions of Janet.  Alternatively, one might also have code that examines `janet/version` and acts accordingly.

Another option might be to avoid using negative indeces completely for a while.

zevv:
> Hey there, long time no see :) Believe it or not: when I saw your name in the contributors' list, that's what finally convinced me I should give Janet a try.

Yeah, as if I have a good taste with these kind of things. The last language I invested in ended up a total dumpster fire, remember.

zevv:
> Another option might be to avoid using negative indeces completely for a while.

I don't hate this. I expect this feature is not used often, as it seems there were no complaints about it earler, so my guess is that this would not affect too many code bases. I know I'm a bit of a whiner myself, and I would have spoken out if I would have needed to use `-2` for the last index.

I agree that `-2` is just a bit too weird, and I'd rather see this "fixed" sooner than later.

sogaiu:
I'm fuzzy on whether plain `slice` would be affected.

I use `-2` in some of my pegs with `slice`, like [here](https://github.com/sogaiu/janet-peg/blob/e4785b3111dbf1fbb78d28ca87953133c6af3dff/janet-peg/rewrite.janet#L14).

narimiran:
> I use -2 and other negative indeces in some of my pegs with slice, like [here](https://github.com/sogaiu/janet-peg/blob/e4785b3111dbf1fbb78d28ca87953133c6af3dff/janet-peg/rewrite.janet#L14) and [here](https://github.com/sogaiu/janet-usages-as-tests/blob/9f58b2a42ba8c6180f3546616db97229149330aa/janet-usages-as-tests/location.janet#L24).

But this will not have to be changed.
You can still use `(slice $& -4 -2)` and it will do the same thing as before, capturing the same two elements. The only thing changed in `slice` is its documentation: when using negative indices, the range is now explained as open-closed, making the index of the last element -1 (not closed-open, as before, with the last element as -2).

Before:
```
 [1  2  3  4  5]
 -6 -5 -4 -3 -2
      [     ]
```

After:
```
 [1  2  3  4  5]
 -5 -4 -3 -2 -1
      [     ]  
```

bakpakin:
The inability to remove stuff from the end of the array and the lack of symmetry is definitely a problem of a bad interface, I support this change, at least for `array/insert`.

As far as making slices using -1 as the end of the array instead of the -2 issues, I agree that it is confusing and unfortunate (along with os/date changing indices to be 0-indexed), but I also believe that constantly changing interfaces that really don't have _strong_ mathematical reasons behind and are more about "hygine" or "clean code" just causes churn.

On the other hand, if one were to make this change, sooner is always better than later to just "rip the bandaid off."

tionis:
Perhaps such improvements could be collected and bundled together into one large backwards-incompatible patch?

CosmicToast:
I would recommend having a "2.0" branch for janet where all such changes can be accumulated until it's ready for release.
It should be possible to rebase it on every point release, meaning people could "opt-in" early as well.

bakpakin:
> I would recommend having a "2.0" branch for janet where all such changes can be accumulated until it's ready for release. It should be possible to rebase it on every point release, meaning people could "opt-in" early as well.

Not a bad idea, but I think the more interesting question is how much should we consider changing in the first place. Lots of small changes could be made, but there is also plenty on could do 

. My gut tells me that changing slice semantics will break a lot more code that `array/insert` and `array/remove`, since for the usecase of inserting at the end of the array, `array/push` and `array/pop` is more natural. The indexing scheme where -1 refers to imaginary "element after the last one" also affects buffers and strings as well.

primo-ppcg:
> My gut tells me that changing slice semantics will break a lot more code that `array/insert` and `array/remove`, since for the usecase of inserting at the end of the array, `array/push` and `array/pop` is more natural. The indexing scheme where -1 refers to imaginary "element after the last one" also affects buffers and strings as well.

I think there's a bit of confusion. I haven't proposed changing the behaviors of `array/slice` or `array/insert`. I believe that they were already correct, but perhaps for the wrong reason. The current behavior has been explained differently, in terms of -1 as the final array index.

As an example, the following code snippets should be equivalent:
```janet
(var a @[:a :b :c :d])
(set a (reverse a))
(array/insert a 0 :e)
(set a (reverse a))
```
```janet
(var a @[:a :b :c :d])
(array/insert a -1 :e)
```
And they are, currently. However the _reason given_ is that -1 is an index "outside" the array, and `array/insert` inserts before that index. What I'm proposing is that this is not as correct as saying that negative indices view the array in reverse, such that inserting at a negative index would actually insert after that index, precisely as the first snippet does. I have likewise changed the description of `array/slice` with that understanding.

The only function that I have proposed changing is `array/remove`. Had `array/insert` and `array/remove` been consistent, I would not have recommended a change, despite the final index of -2 being somewhat weird, and I do believe unprecedented.

bakpakin:
Negative indexing is a weird python-ism that can seem ubiquitous in scripting languages, but it comes with it's own can of worms, and a lot of languages just don't bother. I don't know if it comes from Python originally but it really is more of a hack than any sort of elegant construction, regardless of the particulars.

Also, the original explanation of array/insert and array/remove actually makes less sense to me now. After a call to array/insert, the array is one longer, the same negative index would return a different value. In that sense, the current implementation is more consistent.

nstgc:
I would like to point out that -1 mod n = n-1, making it the last element of an array. Like like, on a ring, -2 is the second to last, and so on. So, from a mathematical standpoint, at least to me, it makes sense. Also, adding my two cents to something related, one of my issues with 1-indexed arrays is how they interact with modular arithmetic. (Looking at you Julia!)

bakpakin:
Getting slightly off topic here, I did a little experiment with redoing negative indexing for slices here: https://github.com/janet-lang/janet/compare/master...negative-indexing-redo

It basically breaks almost all reasonably sized projects I can find written in Janet (not surprisingly).

primo-ppcg:
> I did a little experiment with redoing negative indexing for slices here: [master...negative-indexing-redo](https://github.com/janet-lang/janet/compare/master...negative-indexing-redo)

Of the functions that refer to negative indexing, which includes:
```janet
array/insert
array/remove
array/slice
buffer/blit
buffer/slice
keyword/slice
slice
string/slice
symbol/slice
tuple/slice
```
the implementations of all of these are already correct and do not need to be changed, with exception to `array/remove`. And surely if they hadn't been correct, it would have been mentioned long before now. This PR is Ready to Merge; no further changes are required. If it hadn't been, I wouldn't have marked it as such.

> Also, the proposed explanation of array/insert and array/remove actually makes less sense to me now.

I would like you to imagine a bijective one-dimensional coordinate space translation, _x_ ↦ _x'_, such that _0_ ↦ _-1_, _1_ ↦ _-2_, _2_ ↦ _-3_, etc., and that parallel functions are defined which reference array indices according to this translation. When using these functions, arrays would be indexed in the following way:
```janet
#x' -1 -2 -3 -4 -5
 @[ :a :b :c :d :e ]
```
When inserting an item at index -2, the item would be inserted at index -2, shifting the contents and everything after by one cell. The position of -2 will not have changed. If the contents at -2 are examined, it would be precisely what was inserted. Removing it again would result in the original array.

When indexing from the end of an array, precisely this bijection is used, except that the items of the array are enumerated in reverse order:
```janet
#    0  1  2  3  4
 @[ :a :b :c :d :e ]

#   -1 -2 -3 -4 -5
 @[ :e :d :c :b :a ]
```
such that inserting at -2 would insert between `:e` and `:d`, such that slicing from -2 to -4 would contain `:d` and `:c`, but not `:b`, and such that removing at -2 would remove `:d`.

Note that when referencing a range of cells, as slicing does, or a gap between cells, as inserting does, the behavior is precisely the same as always viewing the array from the front, and pretending that the final index is -2. Janet already has a correct implementation for these; nothing needs to change.

It only makes a difference when referring to a cell location directly, which currently only applies to `array/remove`, although fetching from a negative index for indexed or bytes types could theoretically be allowed.

narimiran:
> Getting slightly off topic here, I did a little experiment with redoing negative indexing for slices here: https://github.com/janet-lang/janet/compare/master...negative-indexing-redo

You changed things that don't need change and shouldn't be changed, i.e. you've changed their behavior and introduced several off-by-one bugs.

bakpakin:
@primo-ppcg This makes sense.

Just to be clear, I will not redo negative indexing for slicing, I just wanted to have a more concrete understanding of how much it was actually used

sogaiu:
So may be the following [code in mendoza](
https://github.com/bakpakin/mendoza/blob/1cb7186d4b0ab892593a407cfacd19e76c2618b2/mendoza/template-env.janet#L60) is affected?

```janet
(defn relative-url
  "Express a url relative to (dyn :url). The given url should
  be absolute from the site root, like /index.html."
  [url]
  (def b-parts (string/split "/" (dyn :url)))
  (array/remove b-parts -2)                      # the form on this line
  (def a-parts (string/split "/" url))
  (while (and (not= 0 (length a-parts))
              (not= 0 (length b-parts))
              (= (a-parts 0) (b-parts 0)))
    (array/remove a-parts 0)
    (array/remove b-parts 0))
  (string (string/repeat "../" (length b-parts)) (string/join a-parts "/")))
```

May be this can be re-expressed in a way to not use a negative index?

---

I'm posting this here rather than in the mendoza repository issues because I think more interested parties (issue participants) will see it.

PR #1244, opened 2023-08-08
Move `bytes?`, `indexed?`, `dictionary?` to corelib

primo-ppcg:
I understand that it's not desirable to move everything to `corelib.c`, especially things that can be implemented simply and concisely in Janet. However, for these three functions I think it may be warranted. Each of these functions compares multiple types, and are used in several places throughout `boot.janet`, and not only in compile-time expanded macros.

For example, in `flatten`:
```janet
(use spork/test)
(timeit-loop [:timeout 1] :flatten (flatten [[[1 2] 3] 4 [5 [6 7 8] 9]]))
```
master:
```janet
flatten 1.000s, 1.556µs/body
```
branch:
```janet
flatten 1.000s, 0.8098µs/body
```
Nearly half the time taken is spent checking the type.

For the interested, moving these functions increases the size of the executable by 128 bytes, at least on my system:
master:
```janet
$ du -b /usr/local/bin/janet
797104	/usr/local/bin/janet
```
branch:
```janet
$ du -b /usr/local/bin/janet
797232	/usr/local/bin/janet
```

PR #1241, opened 2023-08-06
Rework `keys`, `values`, `pairs`

primo-ppcg:
These functions may not be overly important, but that doesn't mean that they can't be optimized. The main performance gain here is creating an array with the correct capacity before iteration begins, rather than pushing to an empty array.
```janet
(use spork/test)

(def xs (range 1000))

(timeit-loop [:timeout 1] "keys  " (keys xs))
(timeit-loop [:timeout 1] "values" (values xs))
(timeit-loop [:timeout 1] "pairs " (pairs xs))
```
master:
```janet
keys   1.000s, 36.86µs/body
values 1.000s, 42.86µs/body
pairs  1.000s, 84.27µs/body
```
branch:
```janet
keys   1.000s, 14.75µs/body
values 1.000s, 19.11µs/body
pairs  1.000s, 56.36µs/body

```
`keys` and `values` are each over twice as fast, and `pairs` is approximately 50% faster.

PR #1247, opened 2023-08-11
Add `lengthable?`

primo-ppcg:
- Adds `lengthable?` to corelib, as mentioned [here](https://github.com/janet-lang/janet/pull/1244#issuecomment-1671333649).
```
(lengthable? x)

Check if x is a bytes, indexed, or dictionary.
```
- Allows potentially non-lengthable items for `keys`, `values`, and `pairs`, as used for example in [spork/generators](https://github.com/janet-lang/spork/blob/ce3e4377b61f14cf4221bcb3c0c10a96382aede8/spork/generators.janet#L18-L21).

There's a few other core functions that I think could benefit from knowing whether an iterable has a pre-determined length.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1554,32 +1554,41 @@
 (defn keys
   "Get the keys of an associative data structure."
   [x]
-  (def arr (array/new-filled (length x)))
-  (var i 0)
-  (eachk k x
-    (put arr i k)
-    (++ i))
-  arr)
+  (if (lengthable? x)
+    (do
+      (def arr (array/new-filled (length x)))
+      (var i 0)
+      (eachk k x
+        (put arr i k)
+        (++ i))
+      arr)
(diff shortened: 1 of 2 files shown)


sogaiu:
I tried out the changes a bit.

The spork test bit mentioned [here](https://github.com/janet-lang/janet/pull/1241#issuecomment-1674138478) seems to be resolved.  Ran the spork test suite and that was fine too.

Also ran the tests I've been running across some other repositories with no issues :+1:

primo-ppcg:
Actually, these functions don't make much sense on a fiber.

For example:
```janet
(defn gen []
  (coro
    (for i 0 10 (yield i))))

(keys (gen))
# ->
@[0 0 0 0 0 0 0 0 0 0]
```
This is true of every major Janet release; the next key of an active fiber is always 0.

Instead, I think `generators/to-array` should be rewritten to do as it says: consume the values into an array.

PR #1255, opened 2023-08-15
Special case common `sort` usages

primo-ppcg:
- Special case common sort usages (`<`, `>`) to use vm ops directly.
- Rewrite `median-of-three` to use 2.667 comparisons on average, rather than 5. It's not as clever, but it is faster.
- Avoid recursing if unnecessary.

The result is approximately 3x faster for these usages, and roughly unchanged for others.
```janet
(use spork/test)
(use spork/misc)

(def rng (math/rng 12345))
(def a (randomize-array (range 100) rng))

(timeit-loop [:timeout 1]
  "sorted "
  (sorted a))

(timeit-loop [:timeout 1]
  "sort   "
  (def b (array/slice a))
  (sort b))

(timeit-loop [:timeout 1]
  "reverse"
  (def b (array/slice a))
  (sort b >))

(timeit-loop [:timeout 1]
  "compare"
  (def b (array/slice a))
  (sort b compare<))
```
master:
```janet
sorted  1.000s, 99.43µs/body
sort    1.000s, 100.1µs/body
reverse 1.000s, 98.23µs/body
compare 1.000s, 131.7µs/body
```
branch:
```janet
sorted  1.000s, 28.41µs/body
sort    1.000s, 28.68µs/body
reverse 1.000s, 31.29µs/body
compare 1.000s, 126.9µs/body
```

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -805,21 +805,31 @@
 ###
 ###
 
-(defn- median-of-three [a b c]
-  (if (not= (> a b) (> a c))
-    a
-    (if (not= (> b a) (> b c)) b c)))
+(defmacro- median-of-three
+  [x y z]
+  ~(if (<= ,x ,y)
+     (if (<= ,y ,z) ,y (if (<= ,z ,x) ,x ,z))
+     (if (<= ,z ,y) ,y (if (<= ,x ,z) ,x ,z))))
+
+(defmacro- sort-partition-template
+  [ind before? left right pivot]
+  ~(do
+     (while (,before? (in ,ind ,left) ,pivot) (++ ,left))
+     (while (,before? ,pivot (in ,ind ,right)) (-- ,right))))
 
 (defn- sort-help [a lo hi before?]
   (when (< lo hi)
-    (def pivot
-      (median-of-three (in a hi) (in a lo)
-                       (in a (math/floor (/ (+ lo hi) 2)))))
+    (def [x y z] [(in a lo)
+                  (in a (div (+ lo hi) 2))
+                  (in a hi)])
+    (def pivot (median-of-three x y z))
     (var left lo)
     (var right hi)
     (while true
-      (while (before? (in a left) pivot) (++ left))
-      (while (before? pivot (in a right)) (-- right))
+      (case before?
+        < (sort-partition-template a < left right pivot)
+        > (sort-partition-template a > left right pivot)
+        (sort-partition-template a before? left right pivot))
       (when (<= left right)
         (def tmp (in a left))
         (set (a left) (in a right))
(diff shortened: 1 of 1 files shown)


sogaiu:
Tangential question...

Regarding:

```janet
      (if (>= left right) (break)))
    (if (< lo right)
      (sort-help a lo right before?))
    (if (< left hi)
      (sort-help a left hi before?)))
```

I've been wondering why `if` over `when` sometimes (assuming no "else"-ish things).  I know there are other cases of using `if` in `boot.janet` in a similar manner but I remain curious about motivations.  Less to type?

sogaiu:
Below are the numbers I got.

master (8df73643):

```
$ janet ~/scratch/pr-1255-sort.janet 
sorted  1.000s, 142.8µs/body
sort    1.000s, 138.2µs/body
reverse 1.000s, 149.5µs/body
compare 1.000s, 175.5µs/body
```

branch (cdd7083c):

```
$ JANET_PATH=$HOME/.local/lib/janet ~/src/janet.primo-ppcg/build/janet ~/scratch/pr-1255-sort.janet 
sorted  1.000s, 44.15µs/body
sort    1.000s, 43.94µs/body
reverse 1.000s, 47.01µs/body
compare 1.001s, 172.4µs/body
```

pepe:
> Tangential question...

@sogaiu I am using `when` only when there is more than one form in the body. My 2 cents tho :-).

sogaiu:
> I am using when only when there is more than one form in the body

Perhaps this is a bit like when one doesn't use braces in some spots in C when one can get away with it...increasing the chances of errors later if one forgets to make an appropriate modification when an additional line / form might need to be added? :)

CosmicToast:
Right. I see `(when ...)` as an `(if (do ...))` rather than a `(if ...)`.
Now that I check, this is somewhat unsurprising, as it is a macro that compiles to exactly that.

primo-ppcg:
> I've been wondering why `if` over `when` sometimes (assuming no "else"-ish things).

Because it creates an unnecessary scoping. A micro-optimization at best, but measurable.

sogaiu:
I guess it would seem that the specific situation would matter.

If measurements suggest it would help, that makes sense.  Otherwise, it doesn't seem like a clear winner to me.  Particularly in languages without source step debugging where the expectation appears to be modifying the source when investigating things.  I think I'd prefer to use `when` the way I've been doing.

primo-ppcg:
> So in the case of this PR, may be it makes sense based on measurements.

Actually, I doesn't seem to make a difference anymore. Perhaps the recent compiler optimizations have made the point moot. We're free to use `when` without feeling guilty :laughing:

sogaiu:
> We're free to use when without feeling guilty

Ah, thanks for sharing this info.

I guess what's tricky is that with folks like you optimizing things (thanks BTW!), it might be a never-ending game of figuring out what to write...oh, may be that's a feature? :)

bakpakin:
Just for reference `when` vs `if` _never_ impacted performance. Old code may have just used `if` instead of `when` because there used to be an else statement, or it is 2 characters shorted.

Anyway, LGTM.

PR #1257, opened 2023-08-18
Update `any?`, `every?`

primo-ppcg:
Updates `any?` and `every?` to be exact functional analogues to `or` and `and`.

- `any?` returns the final value if all preceding values are falsey, rather than `nil`.
- `every?` returns the final value if all preceding values are truthy, rather than `true`.

I believe this makes their definitions more coherent. `any?` would sometimes return an element from the collection (if it were truthy), and sometimes `nil`. `every?` would sometimes return an element from the collection (if it were falsey), and sometimes `true`. After, each will always return an element from the collection.

As their meaning will not have changed when used as a predicate, I believe these changes can be made with minimal fallout. `every?` in particular is much more useful this way. Because of the spatial separation in the documentation, I have duplicated the docstrings from `or` and `and`.

For the interested, the implementation is approximately 10% faster for each.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1424,20 +1424,21 @@
     (fn [& r] (f ;more ;r))))
 
 (defn every?
-  ``Returns true if each value in `ind` is truthy, otherwise returns the first
-  falsey value.``
+  ``Evaluates to the last element of `ind` if all preceding elements are truthy,
+  otherwise evaluates to the first falsey argument.``
   [ind]
   (var res true)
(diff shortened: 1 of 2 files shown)


pepe (approved):
LGTM

sogaiu:
It wasn't really in a real-world usage, but I did make [this change](https://github.com/sogaiu/janet-ref/commit/57d92e43e0612abeb209431aa59a7725d264ac4b) after this PR got merged.

PR #1258, opened 2023-08-18
Update `each` keys before body

### LINKED ISSUE #1220: Wrong results when mutating table inside eachk / eachp
```janet
(let [t @{1 true 2 true 3 true 4 true}]
  (eachk k t
    (when (even? k)
      (put t k nil)))
  t)
# outputs:  @{1 true 3 true 4 true}
# expected: @{1 true 3 true}
```

Using `(each k (keys t))` instead of `(eachk k t)` produces the correct result.
Same applies to `(each [k v] (pairs t))` instead of `(eachp [k v] t)`.

primo-ppcg:
Fixes #1220

Changes `each-template` to update the iteration key after bindings and before the body, rather than after the body. As the iteration key is not exposed to the body, I believe this has no consequence, other than to resolve the linked issue.
```janet
(let [t @{1 true 2 true 3 true 4 true}]
  (eachk k t
    (when (even? k)
      (put t k nil)))
  t)
```
master:
```janet
@{1 true 3 true 4 true}
```
branch:
```janet
@{1 true 3 true}
```

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -436,8 +436,8 @@
               :each ~(,in ,ds ,k)
               :keys k
               :pairs ~[,k (,in ,ds ,k)]))
-         ,;body
-         (set ,k (,next ,ds ,k))))))
+         (set ,k (,next ,ds ,k))
+         ,;body))))
 
 (defn- iterate-template
   [binding expr body]


sogaiu:
Tried the changes out and ran some projects' tests (including spork's) with no issues :+1:

pepe (approved):
LGTM

PR #1277, opened 2023-08-28
Promote (comment) to special form

iacore:
JANET_SLOT_SPLICED + empty array is used here.

The implementation is splice of empty array. It kind of works. When used at top-level, it returns `@[]`. (it's only visible in REPL though)

Is there a way to return `JanetSlot` that represents nothing?

Related: #1272

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -164,10 +164,6 @@
   [sym val]
   ~(def ,sym (if (= nil ,sym) ,val ,sym)))
 
-(defmacro comment
-  "Ignores the body of the comment."
-  [&])
-
 (defmacro if-not
   "Shorthand for `(if (not condition) else then)`."
   [condition then &opt else]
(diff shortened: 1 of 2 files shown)


CosmicToast:
To clarify the effects, since I quickly tested the patch, noting none of this is surprising to me:
1. In the top-level, this returns `@[]`, but does not error. I would say this is fine, since it's in the top-level it's meaningless. `(comment test) # => @[]`
2. In a function-call, it's equivalent to nothing: `(defn x [a b] (+ a b)) (x 1 (comment test) 2) # => 3`
3. In a macro-call, it is unevaluated, and thus remains as-is: `(defmacro x [& body] (pp body)) (x 1 (comment test) 3) # => (1 (comment hm) 3)`

This is ultimately a breaking change (you can no longer use `(comment ...)` to mean nil), but is notably better than the `ignore` macro suggestion, as it does not error in the top-level.
Perhaps it make make sense to keep the current `comment` while reusing this logic under a different name?

bakpakin:
This isn't going to be merged for a number of reasons. This does not belong in the compiler and should be part of the parser.

iacore:
Well, this is a proof of concept. I don't know how to add it to the parser.

Issue #1275, opened 2023-08-28
Make loop `:range` more consistent with `range`


primo-ppcg:
Looking at the asymmetry between these forms:
```janet
> (range 10)
@[0 1 2 3 4 5 6 7 8 9]
```
```janet
> (loop [i :range [10]] (pp i))
10
11
12
13
14
15
...
```
That the second produces an infinite loop is a mistake I confess to having made more than once. The `:range` verb could be modified fairly easily to allow a single term:
https://github.com/primo-ppcg/janet/commit/b8e3482619e659f985230756341a9c851dd2c843
Which compiles to the same code as the two or three term forms:
```janet
(pp ((disasm |(loop [i :range [10]])) :bytecode))
(pp ((disasm |(loop [i :range [0 10]])) :bytecode))
(pp ((disasm |(loop [i :range [0 10 1]])) :bytecode))
# ->
@[(ldi 0 0) (ldi 1 10) (lt 2 0 1) (jmpno 2 3) (addim 0 0 1) (jmp -3) (retn)]
@[(ldi 0 0) (ldi 1 10) (lt 2 0 1) (jmpno 2 3) (addim 0 0 1) (jmp -3) (retn)]
@[(ldi 0 0) (ldi 1 10) (lt 2 0 1) (jmpno 2 3) (addim 0 0 1) (jmp -3) (retn)]
```
This change would also affect `:range-to`, `:down`, and `:down-to`.

---

### Other suggestions
A single term form could also be allowed without needing to be a tuple:
```janet
(loop [cnt :down 10]
  (pp cnt))
```
```
10
9
8
7
6
5
4
3
2
1
```
This can be achieved by modifying a single line:
```diff
@@ /src/boot/boot.janet -416,4 +416,4 @@
 (defn- check-indexed [x]
   (if (indexed? x)
     x
-    (error (string "expected tuple for range, got " x))))
+    [x]))
```
I'm not sure if that's a good idea, just throwing it out there.

It would also be _really nice_ if `:range` could also allow a negative step, as `range` does. I don't know if this could be done without adding some run-time comparisons to the macro expansion, which is definitely undesirable, but if it could I think it would be a good idea. Perhaps the current behavior could be preserved as `:up` and `:up-to`.

pepe:
I like the first modification. About the other suggestion I am not sure 😊.

primo-ppcg:
> I also felt a negative step would have been nice on occasion.

For sure. The problem is the following:
```janet
[...]

PR #1278, opened 2023-08-29
Allow one-term `:range` and `:down` forms

primo-ppcg:
Related: #1275

Makes `:range` more predictable, in particular that:
```janet
(loop [i :range [10]] body)
```
is equivalent to:
```janet
(loop [i :range [0 10] body)
```
rather than producing an infinite loop.

This is consistent with the behavior that the following forms:
```janet
(range 10)
(range 0 10)
```
are also equivalent.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -419,9 +419,11 @@
     (error (string "expected tuple for range, got " x))))
 
 (defn- range-template
-  [binding object rest op comparison]
+  [binding object kind rest op comparison]
   (let [[start stop step] (check-indexed object)]
-    (for-template binding start stop (or step 1) comparison op [rest])))
+    (case kind
+      :range (for-template binding (if stop start 0) (or stop start) (or step 1) comparison op [rest])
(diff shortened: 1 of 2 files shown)


sogaiu:
Tried it out with the usual set of tests and no issues were detected :+1: 

...and some manual trials (edited for readability):

```
$ ./build/janet
Janet 1.30.0-2f43cb84 linux/x64/gcc - '(doc)' for help

repl:1:> (seq [i :range [0 10]] i)
@[0 1 2 3 4 5 6 7 8 9]

repl:2:> (seq [i :range [10]] i)
@[0 1 2 3 4 5 6 7 8 9]

repl:3:> (seq [i :range-to [10]] i)
@[0 1 2 3 4 5 6 7 8 9 10]

repl:4:> (seq [i :down [10]] i)
@[10 9 8 7 6 5 4 3 2 1]

repl:5:> (seq [i :down-to [10]] i)
@[10 9 8 7 6 5 4 3 2 1 0]
```

Having tried out the `down` ones, I did find myself pausing for a moment about their meanings.  I understand about the symmetry angle, just giving some impressions here :)

PR #1282, opened 2023-09-06
Remove old parser code

sogaiu:
This PR contains the removal of a line that looks obsolete.

I think the oldest version of the line in question is [here](https://github.com/janet-lang/janet/blob/05318669540ed8d02f2a2b841333290b61b12ccd/core/parse.c#L406).  AFAICT the code has evolved since that time in such a way such that the line no longer appears necessary.

I ran various tests without detecting problems and failed to turn up any issues working at the REPL.

Perhaps I've missed something though (^^;

### DIFF
--- a/src/core/parse.c
+++ b/src/core/parse.c
@@ -612,7 +612,6 @@ static int atsign(JanetParser *p, JanetParseState *state, uint8_t c) {
             break;
     }
     pushstate(p, tokenchar, PFLAG_TOKEN);
-    push_buf(p, '@'); /* Push the leading at-sign that was dropped */
     return 0;
 }
 


sogaiu:
Actually, it looks like there is a difference in behavior:

```
$ ./build/janet
Janet 1.30.0-037e8288 linux/x64/gcc - '(doc)' for help
repl:1:> (def @ 1)
1
repl:2:> @
repl:3:0: parse error: symbol literal cannot start with a digit
```

VS

```
$ janet
Janet 1.30.0-51c0cf97 linux/x64/gcc - '(doc)' for help
repl:1:> (def @ 1)
1
repl:2:> @
1
```

sogaiu:
There are [some tests to check that symbols with `@` in them work](https://github.com/janet-lang/janet/blob/a13aeaf9557250639018802baceeefb78b834846/test/suite-parse.janet#L110-L117), but they somehow didn't catch the incorrect change made in this PR.

PR #1287, opened 2023-09-12
Document array/push variadic argument

pepe:
Like we discussed in #1284. Just minimal change.


sogaiu:
Thanks for spotting it, suggesting the change, and making the PR :+1:

PR #1289, opened 2023-09-13
Add `:unless` loop modifier

primo-ppcg:
Related: #1286

- Adds `:unless` loop modifier
```janet
> (seq [i :range [0 10] :unless (odd? i)] i)
@[0 2 4 6 8]
```

PR #1296, opened 2023-09-21
add :unless modifier to (doc loop)

Andriamanitra:
`:unless` was added in 1.31.0 (#1289) but never documented

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -592,6 +592,9 @@
   * `:when condition` -- only evaluates the current loop body when `condition`
     is true.
 
+  * `:unless condition` -- only evaluates the current loop body when `condition`
+    is false.
+
   The `loop` macro always evaluates to nil.
   ```
   [head & body]


pepe (approved):
LGTM

sogaiu:
On a side note, there appear to be docstrings where there are phrases such as:

* "when `condition` is true"
* "when `condition` is false"
* "if `expression` is falsey"
* "if `expression` is truthy"

In this particular docstring, there appear to be a mix (e.g. true and truthy).

Do we want to be using `truthy` and `falsey` in some of these cases where `true` and `false` are being used?

bakpakin:
`truthy` is more precise here since any non-nil, non-false value will behave as `true`.

Andriamanitra:
> `truthy`/`falsey` is more precise here since any non-nil, non-false value will behave as `true`, and `nil` will work as `false` (like most conditions).

That's a good point – I changed both `when` and `unless` to say truthy/falsey.

Issue #1312, opened 2023-10-19
`default-peg-grammar` lacks `*` and `+` versions of `:A`, `:D`, `:H`, `:S`, and `:W`


sogaiu:
Unlike `:a`, `:d`, `:h`, `:s`, and `:w`, the "capitalized" versions (negations - e.g. `:A`) in `default-peg-grammar` lack `*` and `+` versions:

https://github.com/janet-lang/janet/blob/cc5beda0d26c0b91fb28540291e950392cf59fc4/src/boot/boot.janet#L2333-L2355

I suspect that on some occasions some of us have missed `:S+` or similar.

May be it makes sense to "complete" the set?

bakpakin:
Looks fine to me

sogaiu:
Thanks for the merge!

PR #1313, opened 2023-10-19
Add more + and * keywords to default-peg-grammar

sogaiu:
This PR is an attempt to address #1312.

Also, the order of some existing rows of `default-peg-grammar` was changed with the intention of easing future maintenance.


pepe (approved):
LGTM

PR #1325, opened 2023-11-10
net/ev: Cleaned up unused NetStateConnect, fixed janet_async_end() ev refcount

zevv:
Not sure if this makes proper sense, needs some review.

This cleans up the unused `NetStateConnect` state in the connect state machine, but not allocating any state to pass to `janet_async_start()` hung my janet at cleanup time; I believe the refcount dec should not depend on state of `evdata`


bakpakin:
LGTM as it has passed CI, so it is working on windows

PR #1328, opened 2023-11-14
Revert "Optimize nil conditions for while and if"

### LINKED ISSUE #1327: (default) not working
I don't know how to name the bug (issue title). It is very fundamental.

```
(def x "A")
#(default x "B")
(def x (if (= nil x) "B" x))
# this works
# (def x (if (pp (= nil x)) "B" x))
(pp x)
```

What I see as output: `B`
What I should see: `A`

iacore:
This seems to fix #1327 (the example in that issue works correctly now)

### DIFF
--- a/src/core/specials.c
+++ b/src/core/specials.c
@@ -530,32 +530,6 @@ static JanetSlot janetc_def(JanetFopts opts, int32_t argn, const Janet *argv) {
     return ret;
 }
 
-/* Check if a form matches the pattern (= nil _) or (not= nil _) */
-static int janetc_check_nil_form(JanetFopts opts, Janet x, Janet *capture, uint32_t fun_tag) {
-    if (!janet_checktype(x, JANET_TUPLE)) return 0;
-    JanetTuple tup = janet_unwrap_tuple(x);
-    if (3 != janet_tuple_length(tup)) return 0;
-    Janet op1 = tup[0];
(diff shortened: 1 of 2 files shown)


sogaiu:
It worked here too (and the tests for `make test` passed as well).

sogaiu:
Nice to have a regression test :+1:

bakpakin:
Ideally we could keep the improvements from the original change without just blanket reverting them

sogaiu:
@primo-ppcg Sorry to bug you (^^;

Do you happen to have some spare cycles to take a look at this?

bakpakin:
I've got it covered.

The issue was with the constant conditional calculation. When `x` was constant in `(if (not= x nil) ...)` , the true and false code would be swapped.

primo-ppcg:
> I've got it covered.

I apologize for the regression. The update https://github.com/janet-lang/janet/commit/56f33f514b4a29a449869ca2899f73d34b292ef7 will fail to optimize code like `(if (not= x nil) ...)` and will only optimize code like `(if (,not= x nil) ...)` within a macro, which is mildly unfortunate. This doesn't appear to have been the source of the error (but rather a failure to update previously existing logic), unless there's another issue with it that I'm unaware of.

iacore:
`not=` in `(if (not= x nil) ...)` could mean something else, I think

Issue #1342, opened 2023-12-25
segfault can result if `net/server`'s handler has incorrect arity


sogaiu:
Following code currently segfaults here:

```janet
(def host "127.0.0.1")
(def port "10765")

(net/server 
  host port
  # XXX
  (fn handler [] :greetings))

(os/sleep 1)

(net/connect host port)
```

Not sure if the following is an appropriate fix, but behavior seems better here with:

```diff
diff --git a/src/core/net.c b/src/core/net.c
index ce8e7d56..e51efa64 100644
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -821,6 +821,7 @@ JANET_CORE_FN(cfun_stream_accept_loop,
     JanetStream *stream = janet_getabstract(argv, 0, &janet_stream_type);
     janet_stream_flags(stream, JANET_STREAM_ACCEPTABLE | JANET_STREAM_SOCKET);
     JanetFunction *fun = janet_getfunction(argv, 1);
+    if (fun->def->arity != 1) janet_panicf("handler arity != 1");
     janet_sched_accept(stream, fun);
 }
```

PR #1343, opened 2023-12-25
Address #1342 - check handler arity

sogaiu:
This is an attempt to address #1342.


pepe (approved):
LGTM

bakpakin:
This is overly restrictive and disallows valid use. Instead we should check if a function's minimum arity is at least 1

PR #1359, opened 2024-01-14
Add buffer/push-* sized int and float

pnelson:
I suspect many binary protocols need to encode/decode floats to/from big/little endian. Here is a first draft, should it be included in the standard library.

I am not tied to the `binary` module as I'm aware there exists `buffer/*` functions that are similar and it could possibly be bundled into the existing `buffer` module. I was concerned about the naming because `buffer/push-word` already exists and the signature is different with no control over endianness. I thought it might be confusing to add `buffer/push-uint32` in that case - but maybe not.

Additionally, `binary/write-float32` and `binary/write-float64` could potentially be removed if we implement something like `math/float32-bits` and `math/float64-bits` respectively for constructing sized `uint`s from the underlying bits and then lean on `binary/write-uint32` and `binary/write-uint64` to write to the buffer.

I just wanted to start these discussions and fix any outstanding bugs that may be present.

### DIFF
--- a/src/core/buffer.c
+++ b/src/core/buffer.c
@@ -320,6 +320,143 @@ JANET_CORE_FN(cfun_buffer_chars,
     return argv[0];
 }
 
+static int should_reverse_bytes(const Janet *argv, int32_t argc) {
+    JanetKeyword order_kw = janet_getkeyword(argv, argc);
+    if (!janet_cstrcmp(order_kw, "le")) {
+#if JANET_BIG_ENDIAN
+        return 1;
+#endif
+    } else if (!janet_cstrcmp(order_kw, "be")) {
+#if JANET_LITTLE_ENDIAN
+        return 1;
+#endif
+    } else if (!janet_cstrcmp(order_kw, "native")) {
+        return 0;
+    } else {
+        janet_panicf("expected endianness :le, :be or :native, got %v", argv[1]);
+    }
+    return 0;
+}
+
+static void reverse_u32(uint8_t bytes[4]) {
+    uint8_t temp;
+    temp = bytes[3];
+    bytes[3] = bytes[0];
+    bytes[0] = temp;
+    temp = bytes[2];
+    bytes[2] = bytes[1];
+    bytes[1] = temp;
+}
+
+static void reverse_u64(uint8_t bytes[8]) {
+    uint8_t temp;
+    temp = bytes[7];
+    bytes[7] = bytes[0];
+    bytes[0] = temp;
+    temp = bytes[6];
+    bytes[6] = bytes[1];
+    bytes[1] = temp;
+    temp = bytes[5];
+    bytes[5] = bytes[2];
+    bytes[2] = temp;
+    temp = bytes[4];
+    bytes[4] = bytes[3];
+    bytes[3] = temp;
+}
+
+JANET_CORE_FN(cfun_buffer_push_uint16,
+              "(buffer/push-uint16 buffer order data)",
+              "Push a 16 bit unsigned integer data onto the end of the buffer. "
+              "Returns the modified buffer.") {
+    janet_fixarity(argc, 3);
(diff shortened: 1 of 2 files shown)


src/core/binary.c
```diff
@@ -0,0 +1,201 @@
...
+    JanetBuffer *buffer = janet_getbuffer(argv, 0);
+
+    int reverse = binary_should_reverse(argv, 1);
+
+    union {
+        uint16_t data;
+        uint8_t bytes[2];
+    } u;
+
+    u.data = (uint16_t) janet_getinteger(argv, 2);
```
pnelson, line 86: I was thinking this should be `janet_getuinteger` which exists below `janet_getinteger` in `src/core/capi.c` but I was getting these warnings/errors on `make valtest`:

```
cc -O2 -g -std=c99 -Wall -Wextra -Isrc/include -Isrc/conf -fvisibility=hidden -fPIC -c build/c/janet.c -o build/janet.o
src/core/binary.c: In function ‘cfun_binary_write_uint16’:
src/core/binary.c:86:25: warning: implicit declaration of function ‘janet_getuinteger’; did you mean ‘janet_getinteger’? [-Wimplicit-function-declaration]
   86 |     u.data = (uint16_t) janet_getuinteger(argv, 2);
      |                         ^~~~~~~~~~~~~~~~~
      |                         janet_getinteger
src/core/capi.c: At top level:
src/core/capi.c:296:10: error: conflicting types for ‘janet_getuinteger’; have ‘uint32_t(const Janet *, int32_t)’ {aka ‘unsigned int(const Janet *, int)’}
  296 | uint32_t janet_getuinteger(const Janet *argv, int32_t n) {
      |          ^~~~~~~~~~~~~~~~~
src/core/binary.c:86:25: note: previous implicit declaration of ‘janet_getuinteger’ with type ‘int()’
   86 |     u.data = (uint16_t) janet_getuinteger(argv, 2);
      |                         ^~~~~~~~~~~~~~~~~
make: *** [Makefile:223: build/janet.o] Error 1
```

src/core/binary.c
```diff
@@ -0,0 +1,201 @@
...
+    JanetBuffer *buffer = janet_getbuffer(argv, 0);
+
+    int reverse = binary_should_reverse(argv, 1);
+
+    union {
+        uint64_t data;
+        uint8_t bytes[8];
+    } u;
+
+    u.data = (uint64_t) janet_getuinteger64(argv, 2);
```
pnelson, line 135: Maybe there is a historical reason for this but I found that I couldn't write an 8 byte `uint64_t`. This function checks for the unsigned integer range using `JANET_INTMAX_DOUBLE` which is `9007199254740992.0` or hex `20000000000000` (7 bytes).

pnelson (commented):
Self-review comments.

bakpakin:
This should almost certainly go in the buffer module, but otherwise seems reasonable.

The ffi module and peg module have some of this functionality but with other limitations. I would prefer if this were just folded into the buffer module as "binary" isn't particularly descriptive for a module name.

src/core/buffer.c
```diff
@@ -320,6 +320,143 @@ JANET_CORE_FN(cfun_buffer_chars,
...
+    JanetKeyword order_kw = janet_getkeyword(argv, argc);
+    if (!janet_cstrcmp(order_kw, "le")) {
+#if JANET_BIG_ENDIAN
+        return 1;
+#endif
+    } else if (!janet_cstrcmp(order_kw, "be")) {
+#if JANET_LITTLE_ENDIAN
+        return 1;
+#endif
+    } else if (!janet_cstrcmp(order_kw, "native")) {
```
pnelson, line 333: `int/to-bytes` used `nil` to represent native endianness. I felt `:native` was more descriptive but I would understand the desire to maintain consistency with what already exists.

sogaiu: Another path to consistency might be for `int/to-bytes` to also support `:native`?

PR #1371, opened 2024-01-26
Fix typo in destructuring

pepe:


sogaiu:
LGTM, but the original text was more fun (^^;

pepe:
Indeed! Closing as fun ruinage :-D.

I must have looked at it the hundredth time, yet I saw it for the first time today.

PR #1389, opened 2024-02-02
Tweak fiber/last-value docstring

sogaiu:
This PR contains a minor docstring tweak to `fiber/last-value`.  It just changes the signature of the function so that it has an argument.


pepe (approved):
LGTM

Issue #1386, opened 2024-01-31
Document how `os/spawn` should be cleaned up or handle zombie processes...


amano-kenji:
For now, I discovered a few tricks for `os/spawn` as below.
```janet
(try
  (with [proc (os/spawn ["prog" "arg1"] :p {:out :pipe})]
    (ev/read (proc :out) :all nil 5))
  ([_] nil))
```
```janet
(try
  (with [proc (os/spawn ["prog" "arg1"] :p {:out :pipe})]
    (ev/with-deadline 5
      (os/proc-wait proc)))
  ([_] nil))
```

If `os/spawn`ed proc is not closed with `with`, the spawned process ends up devouring a CPU core. This behavior is either intended or a bug. If this behavior is intended, `(doc os/spawn)` should mention that `os/spawn`ed processes should be closed with `:close`, `with`, or `os/proc-close`.

To prevent zombie processes from blocking `ev/read` or `os/proc-wait`, I must specify a timeout through `ev/read` or `ev/with-deadline`. This is not documented somewhere. Ideally, a way to deal with zombie processes should be documented in the janet core.

pepe:
I remember there was a flag for not collecting processes, turning them into zombies. Also, a quick search showed that there is a configuration for that.

sogaiu:
Perhaps the following is not what you were thinking of, but this bit from `os/execute`'s docstring seems related:

```
    * :d - Don't try and terminate the process on garbage 
      collection (allow spawning zombies).
```

Since `os/spawn`'s docstring mentions:

```
    Execute a program on the system and return a handle to the process. 
    Otherwise, takes the same arguments as `os/execute`. 
```

it would seem the intention is for `:d` to be usable for `os/spawn` as well.

Further, from the docstrings, it doesn't seem like `:d` is in effect by default, so if zombies can be spawned with the code examples above, the current behavior and docstrings don't seem to be in agreement.

---

On a related note, [these lines](https://github.com/janet-lang/janet/blob/9c14c09962c43f11bb4eb4538904f81386f16b15/src/core/os.c#L541-L563) seemed to be about garbage collection and dealing with zombies:

```c
static int janet_proc_gc(void *p, size_t s) {
    (void) s;
    JanetProc *proc = (JanetProc *) p;
#ifdef JANET_WINDOWS
    if (!(proc->flags & JANET_PROC_CLOSED)) {
        if (!(proc->flags & JANET_PROC_ALLOW_ZOMBIE)) {
[...]

PR #1394, opened 2024-02-07
Improve documentation on subprocess API

amano-kenji:
It closes https://github.com/janet-lang/janet/issues/1386

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -618,7 +618,11 @@ os_proc_wait_impl(JanetProc *proc) {
 
 JANET_CORE_FN(os_proc_wait,
               "(os/proc-wait proc)",
-              "Suspend the current fiber until the subprocess completes. Returns the subprocess return code.") {
+              "Suspend the current fiber until the subprocess completes. Returns the subprocess return code. "
+              "os/proc-wait cannot be called twice on the same process. If `ev/with-deadline` cancels `os/proc-wait` "
+              "with an error or os/proc-wait is cancelled with any error caused by anything else, os/proc-wait still "
+              "finishes in the background. Only after os/proc-wait finishes, a process is cleaned up by the operating "
+              "system. Thus, a process becomes a zombie process if os/proc-wait is not called.") {
     janet_fixarity(argc, 1);
     JanetProc *proc = janet_getabstract(argv, 0, &ProcAT);
 #ifdef JANET_EV
@@ -729,10 +733,11 @@ static int get_signal_kw(const Janet *argv, int32_t n) {
 JANET_CORE_FN(os_proc_kill,
               "(os/proc-kill proc &opt wait signal)",
               "Kill a subprocess by sending SIGKILL to it on posix systems, or by closing the process "
-              "handle on windows. If `wait` is truthy, will wait for the process to finish and "
-              "returns the exit code. Otherwise, returns `proc`. If signal is specified send it instead."
-              "Signal keywords are named after their C counterparts but in lowercase with the leading "
-              "`SIG` stripped. Signals are ignored on windows.") {
+              "handle on windows. If os/proc-wait already finished for proc, os/proc-kill raises an error. After "
+              "sending signal to proc, if `wait` is truthy, will wait for the process to finish and return the exit "
+              "code by calling os/proc-wait. Otherwise, returns `proc`. If signal is specified, send it instead. "
+              "Signal keywords are named after their C counterparts but in lowercase with the leading `SIG` stripped. "
+              "Signals are ignored on windows.") {
     janet_arity(argc, 1, 3);
     JanetProc *proc = janet_getabstract(argv, 0, &ProcAT);
     if (proc->flags & JANET_PROC_WAITED) {
@@ -771,8 +776,9 @@ JANET_CORE_FN(os_proc_kill,
 
 JANET_CORE_FN(os_proc_close,
               "(os/proc-close proc)",
-              "Wait on a process if it has not been waited on, and close pipes created by `os/spawn` "
-              "if they have not been closed. Returns nil.") {
+              "Close pipes created by `os/spawn` if they have not been closed. Then, if os/proc-wait was not already "
+              "called on proc, os/proc-wait is called on it, and it returns the exit code returned by os/proc-wait. "
+              "Otherwise, returns nil.") {
     janet_fixarity(argc, 1);
     JanetProc *proc = janet_getabstract(argv, 0, &ProcAT);
 #ifdef JANET_EV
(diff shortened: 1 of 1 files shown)


amano-kenji:
By the way, `os/proc-kill` actually calls `os/proc-wait`. It is just the C version of `os/proc-wait`.

I changed the documentation. Now, `os/proc-kill` doesn't mention it returns the exit code if wait is truthy.

amano-kenji:
I found a little problem. This is os/proc-wait

```c
#ifdef JANET_EV
    os_proc_wait_impl(proc);
    return janet_wrap_nil();
#else
    return os_proc_wait_impl(proc);
#endif
```

Although ev is enabled, it still returns the exit code. Somehow, `janet_wrap_nil` returns exit code.....

I just tested this code.

```
(def proc (os/spawn ["sleep" "10000000"] :p))

(ev/sleep 1)

(printf "exit code %n" (os/proc-kill proc true :term))
```

The exit code was returned from `os/proc-kill` although ev was enabled. Thus, I'm going to revert the commit.

amano-kenji:
To avoid mentioning `os/proc-wait` in `os/proc-kill` without causing confusion, janet would need to ship with a document about subprocess management, and subprocess functions will need to refer users to the document for details.

Right now, janet doesn't ship with `info janet` or anything like that.

Emacs documentation is heavily hyperlinked. I want doc function to refer to relevant documentation. Right now, it is `os/proc-wait`.

sogaiu:
As I've mentioned elsewhere, I think the expectation with Janet is to read the website docs along with the docstrings (though the docstrings live among the website docs as well (for most releases)).

I'll give an example of something that isn't really in the docstrings nor does it seem practical to put it in some docstring.  How long-strings work has no obvious place in docstrings, nor does syntax for buffers, etc.  If we start to think about it, we may come up with other sorts of examples as well, but I'm not going to go into that at this time.  A point of this paragraph is that docstrings are not practical as the sole method of understanding a programming language.  Having something else as well is way more helpful than not.

The website is available to many users over the Internet and it also exists in a repository that can be cloned if one wants to view those docs locally [1].

Thus I disagree with:

> To avoid mentioning os/proc-wait in os/proc-kill without causing confusion, janet would need to ship with a document about subprocess management, and subprocess functions will need to refer users to the document for details.

I believe we're in the process of [working on subprocess management docs for the website](https://github.com/janet-lang/janet-lang.org/issues/215).  I think it would be better to figure out what goes there before settling on what goes in this PR.

Having said that, I think clarifying the information in this PR could be helpful in formulating subprocess management documentation.

---

[1] Possibly that's a bit more challenging for Windows folks because of the use of `make`, though I think it may still be possible with a bit of work.  If the current state is a significant issue, perhaps a PR to try to address that would be accepted.

amano-kenji:
The fact that janet doesn't document thoroughly doesn't mean it's the right way. The fact that something has bone done in a certain way doesn't mean it can't be improved.

Emacs has complete hyperlinked documentation online and offline.

If I look up function documentation in emacs, I get hyperlinks to relevant details. If I press Ctrl+h K, then I am transferred to info manual section on the function.

Whatever we end up doing, users reading `(doc os/xxx)` should be referred to relevant documentation. Referring users to online documentation is not great.

sogaiu:
I agree I was mistaken about `os/proc-wait`.

However:

> Somehow, janet_wrap_nil returns exit code.....

This is also incorrect.

What appears to happen is that `os_proc_wait_impl(proc);` does not end up returning in this case, it does not appear to get beyond [this line](https://github.com/janet-lang/janet/blob/431ecd3d1a4caabc66b62f63c2f83ece2f74e9f9/src/core/os.c#L598):

```c
    janet_await();
```

I think you need to read what I wrote more carefully though.

[I wrote above](https://github.com/janet-lang/janet/pull/1394#issuecomment-1932196896):

> What I'm doing is to execute small samples of Janet code as well as compare the prose with what I see in the C implementations. I may also end up tracing execution with gdb / rr at some point.

In this case I had not yet traced execution yet due to time constraints, but I think the `os/proc-wait` discussion shows that I should change that "may" to a "will".  It's good that you caught the mistaken conclusion.  Thanks.

At present I intend to go through the docstrings in this PR with the slightly amended (`s/may/will/`) plan in hand to understand them well enough to inform drafting a page on subprocess management.  Once the page on subprocess management has taken shape I think it makes sense to revisit the docstrings, possibly to be followed with some iterating over both.

sogaiu:
> The fact that janet API doesn't document things thoroughly doesn't mean it's the right way. The fact that something has bone done in a certain way doesn't mean it can't be improved.

I don't disagree with this, but I don't think there is anyone who is stepping up to do this yet.

I am working with things as they are, I likely won't object to adjusting if things change, but I doubt they will change any time soon.

It's not soley a matter of making the change either, the new arrangement should also match the needs of the main maintainer.  It has to be doable for him to look after in the long term because the rest of us come and go.

amano-kenji:
Even if the website has a section on subprocess management, there is no guarantee that people will read it or start there.

Many people may start using functions directly. Then, I can imagine them coming to janet matrix channel to ask about `os/proc-wait` and `os/proc-kill` because `(doc os/proc-kill)` doesn't mention peculiarities of `os/proc-wait` such as not being able to wait twice on a process and `os/proc-wait` finishing in the background even if it is cancelled by `ev/with-deadline`.

I had to reverse-engineer those details for hours.... If I'm stuck in my rut after starting with `os/spawn`, it's very likely that the subprocess section on the website will not occur to my mind. So, even if the section is on the website, I'm still likely to bang my head on the wall for hours and be stressed out.

That's what usually has happened to me even if there is documentation because janet internal doc doesn't refer people to background information.

So, janet internal doc needs to refer people to relevant documentation. Hyperlinks make documentation cohesive. If you skip referring people to documentation, then documentation loses cohesiveness.

Documents need to link to each other.

If someone has to reverse-engineer something or wonder where the relevant details are, then the documentation is not good enough. Emacs solved that issue with hyperlinked documentation.

sogaiu:
As I wrote above, I suggest you put thoughts about offline documentation in another issue.

[As bakpakin has mentioned elsewhere, this is not a chatroom.  It is an issue tracker.](https://github.com/janet-lang/janet/issues/1386#issuecomment-1922655204)

Please try to stay on the topic for each issue.  Not doing so too much makes it difficult for current readers but also future readers who may be trying to investigate an issue or understand the context of some decision.  You are causing clutter that may adversely affect the time, energy, and resources of people trying to improve Janet (which I also believe you are genuinely trying to do).

amano-kenji:
Okay.

sogaiu:
Thank you for your cooperation, I appreciate it!

Issue #1349, opened 2024-01-04
`ev/deadline` and `ev/with-deadline` should be documented better.


amano-kenji:
* In `(ev/with-deadline deadline & body)`, `deadline` should become `sec`.
* In `(ev/deadline sec &opt tocancel tocheck)`, I don't know the difference between `tocancel` and `tocheck`. If deadline is reached, what exactly happens to `tocheck` and `tocancel`?
* `ev/deadline` throws an error if deadline is reached. This behavior is not documented.

sogaiu:
Not an answer, but may be some [relevant code](https://github.com/janet-lang/janet/blob/9142f38cbceb72e7d2d8a12846d2c22c2322fc34/src/core/ev.c#L1294-L1310).

Some further bits:

* `tocancel` apparently defaults to `fiber/root` (`janet_vm.root_fiber`), and in the above `.c` code corresponds to `to.fiber`
* `tocheck` apparently defaults to `fiber/current`
(`janet_vm.fiber`), and in the above `.c` code corresponds to `to.curr_fiber`

sogaiu:
Ok, I've looked into this a bit more and have some results to share.  (Note: I'm not done, this is a status update.)

I'll start with sample code:

```janet
(def check-wait 1.1)
(def cancel-wait (* 2 check-wait))
(def deadline (* 0.5 check-wait))

(var tocancel-fib nil)
(var tocheck-fib nil)

(set tocancel-fib
  (ev/go
    (fiber/new
      (fn []
        (print "tocancel: started")
        (set tocheck-fib
             (fiber/new
               (fn []
                 (print "tocheck: started")
                 (printf "tocheck: waiting: %n sec" check-wait)
                 (ev/sleep check-wait)
                 (print "tocheck: ended"))))
        (resume tocheck-fib)
        (printf "tocancel: waiting: %n sec" cancel-wait)
        (ev/sleep cancel-wait)
        (print "tocancel: ended")))))

# yield so tocancel-fib can start
(ev/sleep 0)

(ev/deadline deadline tocancel-fib tocheck-fib)
```

Output looked like:

```
tocancel: started
tocheck: started
tocheck: waiting: 1.1 sec
error: deadline expired
  in ev/sleep [src/core/ev.c] on line 2938
  in <anonymous> [ev-deadline.janet] on line 18, column 18
  in <anonymous> [ev-deadline.janet] on line 20, column 9
```

[...]

PR #1410, opened 2024-02-20
Doc tweaks for ev/deadline and ev/with-deadline

sogaiu:
This is an attempt to address #1349.

Background can be seen in that issue, and a summary is available at [this comment](https://github.com/janet-lang/janet/issues/1349#issuecomment-1954103405).

Thanks to @amano-kenji and @llmII for discussions and investigations.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3736,12 +3736,20 @@
     ~(,ev/thread (fn _spawn-thread [&] ,;body) nil :n))
 
   (defmacro ev/with-deadline
-    `Run a body of code with a deadline, such that if the code does not complete before
-    the deadline is up, it will be canceled.`
-    [deadline & body]
+    ``
+    Create a fiber to execute `body`, schedule the event loop to cancel
+    the task (root fiber) associated with `body`'s fiber, and start
+    `body`'s fiber by resuming it.
+
+    The event loop will try to cancel the root fiber if `body`'s fiber
+    has not completed after at least `sec` seconds.
+
+    `sec` is a number that can have a fractional part.
+    ``
+    [sec & body]
     (with-syms [f]
       ~(let [,f (coro ,;body)]
-         (,ev/deadline ,deadline nil ,f)
+         (,ev/deadline ,sec nil ,f)
          (,resume ,f))))
 
   (defn- cancel-all [chan fibers reason]
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -2952,10 +2952,15 @@ JANET_CORE_FN(cfun_ev_sleep,
 
 JANET_CORE_FN(cfun_ev_deadline,
               "(ev/deadline sec &opt tocancel tocheck)",
-              "Set a deadline for a fiber `tocheck`. If `tocheck` is not finished after `sec` seconds, "
-              "`tocancel` will be canceled as with `ev/cancel`. "
-              "If `tocancel` and `tocheck` are not given, they default to `(fiber/root)` and "
-              "`(fiber/current)` respectively. Returns `tocancel`.") {
+              "Schedules the event loop to try to cancel the `tocancel` "
+              "task as with `ev/cancel`. After `sec` seconds, the event "
+              "loop will attempt cancellation of `tocancel` if the "
+              "`tocheck` fiber is resumable. `sec` is a number that can "
+              "have a fractional part. `tocancel` defaults to "
+              "`(fiber/root)`, but if specified, must be a task (root "
+              "fiber). `tocheck` defaults to `(fiber/current)`, but if "
+              "specified, should be a fiber. Returns `tocancel` "
+              "immediately.") {
     janet_arity(argc, 1, 3);
     double sec = janet_getnumber(argv, 0);
     JanetFiber *tocancel = janet_optfiber(argv, argc, 1, janet_vm.root_fiber);


pepe (approved):
LGTM

amano-kenji:
It is okay, but it doesn't explain the relationship between `tocheck` and `tocancel`.

pepe:
@amano-kenji, I would like to ask you to edit your last comment to remove vulgarism, please.

llmII:
The only relationship between `tocheck` and `tocancel` that is definitive is within `ev/with-deadline`. With `ev/deadline` it is an optional but likely a typical association that the task `tocancel` started the fiber `tocheck` and is interested in receiving an error rather than waiting on `tocheck` to complete, and possible could respond by `cancel`ing `tocheck`. I'm not sure how or if the typical use case should be in the doc-string because it is possible to have something less typical. It might be that site documentation could benefit from outlining the typical usage, but I believe that if one is using the typical usage, they could likely just use `ev/with-deadline`. Which would of sorts,  roundabouts, lead towards that one probably is using `ev/deadline` specifically for a special usage and thus documenting the typical usage as with `ev/with-deadline` would fail to make sense to do in the docstring?

src/boot/boot.janet
```diff
@@ -3736,12 +3736,13 @@
     ~(,ev/thread (fn _spawn-thread [&] ,;body) nil :n))
 
   (defmacro ev/with-deadline
-    `Run a body of code with a deadline, such that if the code does not complete before
-    the deadline is up, it will be canceled.`
-    [deadline & body]
+    ``Run a body of code with a deadline of `sec` seconds, such that if
+    the code does not complete before the deadline is up, it will be
```
llmII, line 3740: I'm unsure if this is true, I thought it was `tocancel` that got canceled. The body of code goes into `tocheck`. I'd think that `tocheck` isn't canceled here at all, unless the parent chooses to catch the deadline error and then `cancel` `tocheck`?

sogaiu:
> It is okay, but it doesn't explain the relationship between tocheck and tocancel. It doesn't document the art of choosing the right fiber for tocancel.

AFAIU, there isn't necessarily a relationship between `tochecck` and `tocancel`.

IIUC, establishing one is entirely up to the programmer.  There doesn't need to be one for this function operate.

src/boot/boot.janet
sogaiu: The default value of `tocancel` is `(fiber/root)`.

sogaiu: The following code:

```janet
(let [fib (coro
            (pp [:fib (fiber/current)])
            (pp [:lineage (debug/lineage (fiber/root))])
            (ev/sleep 0.2)
            (yield :hi))]
  (def tocancel (ev/deadline 0.1 nil fib))
  (pp [:tocancel tocancel])
  (resume fib))
```

gives the output (here anyway):

```
(:tocancel <fiber 0x5652813D8EC0>)
(:fib <fiber 0x5652814228A0>)
(:lineage @[<fiber 0x5652813D8EC0> <fiber 0x565281383900> <fiber 0x565281418480> <fiber 0x565281421770> <fiber 0x5652814228A0>])
error: deadline expired
  in ev/sleep [src/core/ev.c] on line 2945
  in <anonymous> [<anonymous>] on line 22, column 13
  in _thunk [<anonymous>] (tailcall) on line 26, column 3
```

sogaiu: So the above was meant to be somewhat close to a macroexpansion of `ev/with-deadline`.

For a use of `ev/with-deadline`:

```janet
(ev/with-deadline 0.01
                  (ev/sleep 1)
                  :hi)
```

gives (here anyway):

```
error: deadline expired
  in ev/sleep [src/core/ev.c] on line 2945
  in <anonymous> [repl] on line 2, column 19
```

I think this is because when `tocancel` is specified as `nil`, the value of `(fiber/root)` is used for `tocancel` instead.

amano-kenji:
I want to deliver precise emotions. I'm aiming for precise communication. If I didn't use fuck, I can't deliver my emotions precisely. I don't know how to deliver it other ways, yet.

amano-kenji:
If I was the author of this pull request, I would add more background information to the docstring for `ev/deadline`.

I would at least document how cancelling `tocancel` usually leads to cancellation of `tocheck` as well. If I don't understand the art of choosing fibers for `ev/deadline`, I wouldn't feel comfortable about using it.

llmII:
@amano-kenji wrote:

> I would at least document how cancelling `tocancel` usually leads to cancellation of `tocheck` as well, but `tocheck` doesn't have to be cancelled along with `tocancel` in some cases. If I don't understand the art of choosing fibers for `ev/deadline`, I wouldn't feel comfortable about using it. People who aren't as OCD as I am would probably just avoid using `ev/deadline` because they don't know what `tocancel` is supposed to be.

The problem would be that that isn't necessarily true. One can contrive ways in which this would not be the case. Say you have a task (A) and a task (B). B knows about A and knows A has an interest in the cancellations of fibers it runs under a deadline. `ev/deadline` allows for more uses than the typical, though `ev/with-deadline` outlines the typical usage and though I find it hard to have a valid use otherwise, I'm sure it will be imagined, and documentation stating the typical should probably say something more along the lines of "Typically used via `ev/with-deadline`, and not called manually".

> > What the hell is root fiber? How is it related to current fiber in any way? It doesn't make sense at all.
> 
> That's not good enough. This will lead to confusion or reverse-engineering. Perhaps, people can actually freeze `ev/deadline` for ever with a wrong fiber for `tocancel`.
> 
> If the background information is too verbose, then `ev/deadline` can refer people to an offline document.

Documenting what is a Fiber, and a Task, and a Root Fiber in every single function using them is not a good idea. That's way too verbose. It shouldn't be surprising or unexpected for one to glean those terms from referencing the language documents.

amano-kenji:
I'd likely mention something like

> The root fiber is a task in the event loop. Cancelling the root fiber leads to cancellation of the current fiber running in the task. If you choose anything other than the root fiber, then cancelling it should lead to cancellation of the current fiber.

amano-kenji:
I would likely mention something like

> The root fiber is a task in the event loop. Cancelling the task leads to cancellation of `tocheck` running in the task. If anything other than the root fiber is given as `tocancel`, cancelling `tocancel` leads to cancellation of `tocheck`. Otherwise, `ev/deadline` may freeze for ever.

llmII:
> I would likely mention something like
> 
> > Cancelling `tocancel` must lead to cancellation of `tocheck`. Otherwise, `ev/deadline` may freeze for ever. The current fiber is cancelled when the root fiber is cancelled.
> 
> This is not too verbose. You can probably condense it down to
> 
> > Cancelling `tocancel` must lead to cancellation of `tocheck`. Otherwise, `ev/deadline` may freeze for ever.
> 
> After reading this, people will realize cancelling the root fiber leads to cancellation of the current fiber.

If anything other than the root fiber of `tocheck` is given as `tocancel` then `tocheck` might not be canceled. Cancelling the root fiber of `tocheck` is no guarantee that `tocheck` will be canceled. If `tocheck` `protect`s all it's calls into the event loop it will remain un-errored (the fiber created by protect receives the error thrown by calling into the event loop via `ev/*` functions).

amano-kenji:
You can freeze `ev/deadline` unknowingly?

llmII:
It *should* be possible. The likelihood that someone will without purposefully mucking around in `ev/deadline` is minimal. It's also why one should limit their error handlers to the minimum body of code needing to run within an error handler (so that the error propagates upwards).  I think some of the code I displayed in my review actually illustrated that the body of `ev/with-deadline` can continue to run even after the root fiber of that body which is put into a fiber is cancelled. `tocheck` is just never directly cancelled by `ev/deadline`.

amano-kenji:
I read the definition of `ev/with-deadline`.

```janet
(defmacro ev/with-deadline
    `Run a body of code with a deadline, such that if the code does not complete before
    the deadline is up, it will be canceled.`
    [deadline & body]
    (with-syms [f]
      ~(let [,f (coro ,;body)]
         (,ev/deadline ,deadline nil ,f)
         (,resume ,f))))
```
I wrote test.janet
```
(def fiber (coro
             (try
               (forever (ev/sleep 1))
               ([err f]
                (eprint "fiber cancelled")
                (propagate err f)))))

(def task (ev/spawn
            (try
              (forever
                (ev/sleep 1))
              # ignore errors
              ([err f]))))

(pp (ev/deadline 5 task fiber))
(print "ev/deadline returned...")
(resume fiber)
```
Running this script results in
```
$ janet test.janet
<fiber 0x561235B2B600>
ev/deadline returned...
```
Thus, `ev/deadline` returns `tocancel` immeidately. It doesn't block. `ev/with-deadline` cancels `coro` by cancelling the root fiber.

llmII:
Borrowing from my review:

This also shows the body is still alive...

``` janet
(pp
  (protect
    (ev/with-deadline 0.01
      (protect (ev/sleep 2)) # the call to `ev/sleep` receives the error...
      (pp (fiber/status (fiber/current)))
      (pp [:fib (fiber/current)])
      (pp [:lineage (debug/lineage (fiber/root))])
      (flush)
      (yield :hi))))
```

With results:
``` janet
:alive
(:fib <fiber 0x06EB603B8050>)
(:lineage @[<fiber 0x06EB6026CED0> <fiber 0x06EB603B7BF0> <fiber 0x06EB603B7FE0> <fiber 0x06EB603B8050>])
(true :hi)
nil
```

And
``` janet
(let [fib (coro
            (pp [:fib (fiber/current)])
            (pp [:lineage (debug/lineage (fiber/root))])
            (ev/sleep 0.2)
            (yield :hi))]
  (pp [:tocancel (ev/deadline 0.1 nil fib)])
  (pp (protect (ev/sleep 5)))
  (print "Cancellation has occurred.")
  (resume fib))
```

Has results:
```
:tocancel <fiber 0x06EB6026CED0>)
(false "deadline expired")
Cancellation has occurred.
(:fib <fiber 0x06EB6026DD40>)
(:lineage @[<fiber 0x06EB6026CED0> <fiber 0x06EB6026DA30> <fiber 0x06EB6026DD40>])
:hi
```

llmII:
@amano-kenji wrote:

> I think `ev/deadline` should mention that it doesn't cancel `tocheck` and returns `tocancel` immediately without waiting for anything.

It already mentions in both the new documentation and the old that it is `tocancel` that gets canceled. It takes a lot of assumption to arrive at the conclusion that `tocheck` is actually canceled by `ev/deadline`. 

The current documentation for `ev/with-deadline` is wrong though, and the tweaked documentation doesn't highlight this fact either. I only realized while reviewing that I was automatically interpreting the word "it" as "root fiber" and that the root fiber was `(fiber/root)`.

amano-kenji:
I think `ev/deadline` should mention that it returns `tocancel` immediately.

amano-kenji:
The current docstring for `ev/deadline` says

> Set a deadline for a fiber `tocheck`.

deadline is a bad choice of work for `tocheck`. This word makes me think `tocheck` is cancelled beyond the deadline.

Deadline for `tocheck` means `tocheck` is dead...!!

llmII:
That's going to argue what words mean or imply. I'm not sure if there is a better selection of a name.

In life, one often continues and completes work after a deadline has failed to be met, but it means they did the work for nothing (it's unaccepted as past due). This is exactly what happens to `tocheck`. It continues it's work unless told otherwise even after it failed to meet the deadline.

amano-kenji:
> If you don't finish task A, I will cancel your mom, but I won't do anything to task A.

amano-kenji:
A better docstring would be

> Wait fiber `tocheck` for `sec` seconds. If it doesn't finish in `sec` seconds, `tocancel` will be cancelled as with `ev/cancel`, but nothing happens to `tocheck` directly.

amano-kenji:
> Set a deadline for a fiber `tocheck`

This wording is still confusing. This made me think `tocheck` dies....

I'd rather write

> Set a deadline for `tocancel`. If `tocheck` doesn't finish in `sec` seconds, `tocancel` is cancelled as with `ev/cancel`.

llmII:
> > Set a deadline for a fiber `tocheck`
> 
> This wording is still confusing. This made me think `tocheck` dies....
> 
> I'd rather write
> 
> > Set a deadline for `tocancel`. If `tocheck` doesn't finish in `sec` seconds, `tocancel` is cancelled as with `ev/cancel`.

That would be wrong. The thing checked for liveliness beyond a deadline expiration is the thing which has a deadline set upon it. `tocancel` does not get checked for liveliness.

> I may even add
> 
> > Cancelling `tocancel` can lead to cancellation of `tocheck` if `tocancel` is the root fiber for `tocheck`.
> 
> Deadline for something means death for the thing in plain english. The documentation should make it clear that death directly comes to `tocancel` only.

Adding an ambiguous truth is likely less useful than adding nothing and letting it remain true that "there is nothing documented about anything being done to `tocheck` so we cannot assume anything is done against it".

amano-kenji:
Then, this is better.

> Wait fiber `tocheck` for sec seconds. If it doesn't finish in sec seconds, `tocancel` will be cancelled as with ev/cancel, but `ev/deadline` doesn't do anything directly to tocheck.

llmII:
> Then, this is better.
> 
> > Wait fiber `tocheck` for sec seconds. If it doesn't finish in sec seconds, `tocancel` will be cancelled as with ev/cancel, but `ev/deadline` doesn't do anything directly to tocheck.

Wait is untrue, there is no waiting, according to what you said earlier, `ev/deadline` returns immediately.

amano-kenji:
How about this?

> If `tocheck` doesn't finish in sec seconds, `tocancel` will be cancelled as with ev/cancel, but `ev/deadline` doesn't do anything directly to `tocheck`.

amano-kenji:
> a misunderstanding of the term deadline

It is not a misunderstanding. Deadline by itself doesn't explicitly mention which things die after deadline.

In matters of death, we should explicitly mention which things are killed directly.

llmII:
A deadline implies the death of validity of work, not the death of that doing the work.

amano-kenji:
A deadline has many meanings. It's not just one thing. It can mean a different things to a different person.

Unless you want to define deadline in `ev/deadline`, you need to let people know that `tocheck` is not cancelled directly.

llmII:
> A deadline has many meanings. It's not just one thing. It can mean a different things to a different person.
> 
> One of definitions on dictionary.com says
> 
> > a boundary around a military prison beyond which a prisoner could not venture without risk of being shot by the guards.
> 
> This is the definition I thought of. Deadline for prisoner A means death for prisoner A.

The validity of the work of life expired beyond the boundary, yes. The work is still being done minutes beyond the deadline in that case (until true death occurs, depending upon where one is shot, and it was a risk, not a guarantee).

amano-kenji:
So, you just want to accept the possibilty that other people will misinterpret the docstring?  That's esoteric.

llmII:
I'm not against adding that "nothing is explicitly done to `tocheck` beyond checking it's liveliness" (and wording may differ from mine present).

I am wary of removing the terminology "deadline" from the `docstring` due to possible misinterpretation. Failure to interpret the meaning of the word, even from amongst the context of the other documentation, is very unlikely, unless one makes assumptions. Assuming a deadline implies death has no bearings in reality outside of where a very specific and non-general definition is used (which means someone's picking the definition they like, vs something more general). In almost all cases I've seen deadline used as a term it implied the death of validity of work (and this is me summing things up). The first, and nearly most applicable definition on dictionary.com is as follows:

```
noun

    the time by which something must be finished or submitted; 
    the latest time for finishing something: a five o'clock deadline.
```

It nowhere implies relevance towards anything but the validity of the work, nor does it say that one doesn't continue to work on it past deadline.

amano-kenji:
> Assuming a deadline implies death has no bearings in reality

The docstring for `ev/deadline` mentions it already tries to kill `tocancel` with `ev/cancel`. Perhaps, bakpakin forgot to mention that it also tries to kill `tocheck` with `ev/cancel`. There is no way to know whether someone forgot to mention something.

llmII:
I said:
> I'm not against adding that "nothing is explicitly done to `tocheck` beyond checking it's liveliness" (and wording may differ from mine present).

> > Assuming a deadline implies death has no bearings in reality
> 
> The docstring for `ev/deadline` mentions it already tries to kill `tocancel` with `ev/cancel`. Perhaps, bakpakin forgot to mention that it also tries to kill `tocheck` with `ev/cancel`. There is no way to know whether someone forgot to mention something.

Which would solve your "what happens to `tocheck` conundrum".  The key is to not say "nothing happens to `tocheck`" since that could be untrue, and to note that `ev/deadline` does not do anything to `tocheck` *if* it must be noted.

amano-kenji:
> nothing is explicitly done to tocheck beyond checking it's liveliness

Yes, I would like that.

sogaiu:
Still processing the content above (^^;

For the moment, my leaning is that perhaps both docstrings might be further improved.

I think `ev/with-deadline`'s docstring is not quite correct (as pointed out by llmII).

I can see some benefit in trying to further clarify `ev/deadline`'s docstring.

src/boot/boot.janet
amano-kenji: The root fiber is trying to resume a fiber constructed by `ev/with-deadline`. If you cancel the root fiber while it's trying to resume a fiber, an error is sent to the fiber through the root fiber.

If the `ev/with-deadline` fiber propagates the error upward, the root fiber is also cancelled.... If the fiber ignores the error, then nothing really happens...

So, it's correct to say that the block of code inside `ev/with-deadline` is cancelled with an error. If you ignore the error, the cancellation stops.

amano-kenji (changes requested):

src/boot/boot.janet
amano-kenji: I suggest adding something like

> If the body of code inside `ev/with-deadline` discards the error caused by `ev/with-deadline`, `ev/with-deadline` doesn't stop it.

sogaiu:
Ok, based on discussions with @amano-kenji and @llmII, I've put together [the following candidate for `ev/deadline`](https://github.com/sogaiu/janet-fiber-and-event-loop-notes/blob/96727a3d58423f7401ea9f2c8ad15d989f146114/doc/ev_deadline.md#information):

---

`(ev/deadline sec &opt tocancel tocheck)`

Schedules the event loop to try to cancel the `tocancel` task.

The cancellation will be attempted by the event loop if the `tocheck` fiber is resumable after `sec` seconds.

`sec` is a number that can have a fractional part.

`tocancel` defaults to `(fiber/root)`, but if specified, must be a task (root fiber).

`tocheck` defaults to `(fiber/current)`, but if specified, is a fiber.

Returns `tocancel` immediately.

amano-kenji:
That is not confusing to me.

amano-kenji:
Mention that `ev/with-deadline` may try and fail to cancel the body of code inside it if the body discards the error.

llmII:
> Mention that `ev/with-deadline` may try and fail to cancel the body of code inside it if the body discards the error.

That's incorrect, the body of code within `ev/with-deadline` is the `tocheck` fiber in `ev/deadline`. No cancellation is attempted on `tocheck` whatsoever (otherwise, the code doing cancellation would have to figure out if it needs to call `ev/cancel` since it's a root fiber, or `cancel` since it's just a fiber).

llmII:
> `tocheck` defaults to `(fiber/current)`, but if specified, should be a fiber.

I'll note that his is true, as a root fiber is a fiber. Any kind of fiber is fine for `tocheck` I believe. One could theoretically do something like the following and it'll work.

``` janet
(ev/spawn
  (ev/deadline 5)
  (pp [:fiber-current (fiber/current) :fiber-root (fiber/root)])
  (print "here")
  (ev/sleep 10))
```

In this case, the current fiber, and the root fiber are one in the same, and is the value for `tocheck` and `tocancel`.

sogaiu:
> I'll note that his is true, as a root fiber is a fiber. Any kind of fiber is fine for tocheck I believe. 

Yes, the intent of the text was may be not so apparent (^^;

Perhaps:

> `tocheck` defaults to `(fiber/current)`, but if specified, can be any type of fiber.

is an improvement?

Another possibility might be:

>  `tocheck` defaults to `(fiber/current)`, but if specified, should be a fiber.  Note that `tocheck` does not need to be a task or root fiber.

Where are my better terms when I need them?

llmII:
@sogaiu 

I prefer the first. The other is more explicit but probably more verbose than necessary I'd think.

amano-kenji:
Because `ev/with-deadline` resumes the body inside the root fiber, the error is passed down to the body through the root fiber. So, `ev/with-deadline` is trying to kill the body.

llmII:
If you say that, then you have to say `ev/deadline` is trying to kill `tocheck`. You're ignoring the macro's expansion and the semantics behind what is true for `ev/deadline` remains true for `ev/with-deadline`. No action is taken against the body except to check it's liveliness. 

If you look at the example code, the fiber created by the body is clearly alive. Who knows if all of it's calls into the event loop are protected either because it directly does so, or finds it necessary to do so because it calls into code that can error for various reasons, or calls functions that interact with the event loop but protect it from the error generated thereby?

Nothing is done to the body because the body is `tocheck` in `ev/deadline` which `ev/with-deadline` expands to underneath the hood, also clearly demonstrated by the code above.

amano-kenji:
`ev/with-deadline` still gave an order to kill the body through the event loop and the root fiber.

llmII:
Again, untrue, it orders the value of `(fiber/root)` to be canceled. That's all. Anything happening due to that is coincidence at best and not guaranteed whatsoever. Saying a function (or in this case macro) does something it clearly doesn't is to lie to the consumers of the function (or in this case, macro).

llmII:
The root fiber doesn't deliver the error. Only if the body calls into the event loop will it get the error. That's only if it's calls into the event loop are unprotected. There is no guarantee that it makes a call into the event loop after the deadline either.

A user may intend that `ev/with-deadline` cancel the body, but it doesn't ever. Otherwise it'd be calling `cancel` instead of `ev/cancel`.

amano-kenji:
You have any better docstring for `ev/with-deadline`?

llmII:
The docstring in my review outlines exactly what the function does. Nothing else. It does not lie to the user about what the function does.

    ``Run a body of code with a deadline of `sec` seconds, such that if
    the code does not complete before the deadline is up, the value of
    `(fiber/root)` will be canceled. `sec` is a number that can have a 
    fractional part.``

Perhaps it can be improved, but it should not invent fictions about what it actually does.

amano-kenji:
That's confusing as hell......

sogaiu:
I'm still working on `ev/with-deadline`.

[These](https://github.com/sogaiu/janet-fiber-and-event-loop-notes/blob/b091f22bd5aaedd47301888239809f7d541300d7/doc/ev_with-deadline.md#information) are very much in draft state, but FWIW.

sogaiu:
Please stop with the use of profanity.

amano-kenji:
Fine... But, that docstring doesn't explain the intention behind `ev/with-deadline`.

sogaiu:
Which docstring are you referring to?

I don't necessarily disagree.

llmII:
@amano-kenji 
It is known that obviating what exactly is getting canceled with `ev/with-deadline` is difficult, to the degree I nop'd and went with just say exactly what happens. The only other way I could figure out to outline what's getting canceled would mean the docstring would contain documentation on the event loop.

@sogaiu 
Looking at [those](https://github.com/sogaiu/janet-fiber-and-event-loop-notes/blob/b091f22bd5aaedd47301888239809f7d541300d7/doc/ev_with-deadline.md#information) 
I so far like the first one the best.

amano-kenji:
> the value of `(fiber/root)` will be canceled

This sounds esoteric and confusing. It would throw me off balance if I haven't been in the discussion for the last few days.

I would write

> The task(the root fiber) running the body of `ev/with-deadline` will be cancelled.

If the body of `ev/with-deadline` blocks the whole task with CPU intensive operations, then what happens?

sogaiu:
> If the body of ev/with-deadline blocks the whole task with CPU intensive operations, then what happens?

I didn't test, but IIUC, the [website docs say this about the event loop](https://janet-lang.org/docs/event_loop.html#Blocking-the-Event-Loop):

> When using the event loop, it is important to be aware that CPU bound loops will block all other tasks on the event loop.

amano-kenji:
So, reception of the cancellation error will be delayed after deadline if the body of `ev/with-daedline` stalls the task.

sogaiu:
As I mentioned, I have not tested.

llmII:
> I would write
> 
> > The task(the root fiber) running the body of `ev/with-deadline` will be cancelled.

That's closer to the wording @sogaiu has in what he linked which I'm fine with.

> If the body of `ev/with-deadline` blocks the whole task with CPU intensive operations, then what happens?

It does exactly that, it blocks the whole task... It's why we can't say that `ev/with-deadline` kills the body because the body can be ill-behaved. It could even perhaps interact with some event loop from another project where it ends up not actually being cpu intensive but since it never calls into the event loop it actually successfully completes.

amano-kenji:
> The value of `(fiber/root)`

This has a disorienting effect on my psyche. This is a riddle.

sogaiu:
Re: `(fiber/root)`

> This sounds esoteric and confusing.

I think this is not reasonable characterization for a few reasons:

* It specifically mentions something you can look up the docstring for
* `fiber/root` is explicitly mentioned in the website docs the event loop in [this section](https://janet-lang.org/docs/event_loop.html#An-Overview-of-the-Event-Loop)

`ev/*` callables have to do with the event loop and the docstrings for `ev/deadline` and `ev/with-deadline` should be able to assume a certain amount of familiarity with the event loop.

---

If you are thrown off-balance, perhaps you should consider whether there are other factors that are contributing to that as well.  Are you calm right now?  Or are you caught up in emotion?

amano-kenji:
I read the event loop twice. Perhaps, the documentation doesn't inculcate the concept of root fiber enough.

sogaiu:
Glad you read them.

I am thinking to suggest the addition of a glossary to the website docs.  You can see the beginning of a draft [here](https://github.com/sogaiu/janet-fiber-and-event-loop-notes/blob/b091f22bd5aaedd47301888239809f7d541300d7/README.md#glossary).

amano-kenji:
I think https://janet-lang.org/docs/event_loop.html doesn't account for the fact that humans don't remember concepts they pass by just once.

sogaiu:
Your observations are not lost on me.  It's part of why I work on docstrings and examples and have been doing so over the past few years.

amano-kenji:
Because humans just forget words they pass by briefly, docstring should reinforce the concept of root fiber...

sogaiu:
I don't agree with necessarily though.

sogaiu:
I'll remind you that bakpakin has stated that this is an issue tracker and not a chat room.

llmII (approved):
LGTM!

amano-kenji:
mpv-shot0001

src/core/ev.c
```diff
@@ -2952,10 +2952,15 @@ JANET_CORE_FN(cfun_ev_sleep,
...
 JANET_CORE_FN(cfun_ev_deadline,
               "(ev/deadline sec &opt tocancel tocheck)",
-              "Set a deadline for a fiber `tocheck`. If `tocheck` is not finished after `sec` seconds, "
-              "`tocancel` will be canceled as with `ev/cancel`. "
-              "If `tocancel` and `tocheck` are not given, they default to `(fiber/root)` and "
-              "`(fiber/current)` respectively. Returns `tocancel`.") {
+              "Schedules the event loop to try to cancel the `tocancel` "
+              "task as with `ev/cancel`.  After `sec` seconds, the event "
+              "loop will attempt cancellation of `tocancel` if the "
+              "`tocheck` fiber is resumable.  `sec` is a number that can "
```
amano-kenji, line 2958: > resumable.  `sec`

should become

> resumable. `sec`

sogaiu: Thanks for noticing.

I think I found a third one too (^^;

Since https://github.com/janet-lang/janet/commit/5de889419ff26b710b706958bf99e180d084f564 seems to suggest that lengthier docstrings might become an issue, I've removed the extra spaces.

I suspect there may be other cases in other docstrings.

PR #1414, opened 2024-02-23
Fix documentation of peg/replace

MaxGyver83:
It looks like this piece of text was overseen while renaming `repl` to `subst` in commit 485099fd6e9e88fdc7521fff7e54a6d7023e9871.


sogaiu:
Yes, that seems correct.

LGTM

pepe (approved):
LGTM

PR #1432, opened 2024-03-23
Add additional format options for os/clock

wishdev:
Allows for more flexibility in the returned value from os/clock


sogaiu:
Some minor things about the proposed docstring:

For this bit of text in the docstring:

> {seconds, nanoseconds}

May be it's better to not use curly braces as that's what's used for the literal syntax of structs.

Perhaps square brackets would work better?

Also, slight typo here?

> - :int: Return the number of seconds as am integer

May be "an" is what was intended (not "am"), though may be there is some [humor value](https://www.youtube.com/watch?v=d0LaT6qVRpg)...

wishdev:
Fixed, thanks for the review!

sogaiu:
Nice to have the tests :)

I tried the branch here manually successfully and the tests passed too.

PR #1445, opened 2024-05-10
Add ffi/defbind-alias

wishdev:
I would like the ability to alias an FFI method in order to maintain the naming convention of the underlying shared library

As an example I would like to be able to do the following

```
(ffi/defbind-alias tb_poll_event _tb_poll_event :int [event :ptr])
(defn tb_poll_event
  []
  (def event (ffi/write tb_event (tb_event/new)))
  (_tb_poll_event event)
  (zipcoll tb_event_keys (ffi/read tb_event event)))
```

The shared library method is called tb_poll_event but it requires a janet "wrapper" because it returns a struct pointer.


bakpakin:
Makes sense, LGTM. I usually try to keep a one-to-one mapping for FFI but this would makes sense.

PR #1457, opened 2024-06-15
Fix buffer push uint max

pnelson:
I was using the newer `buffer/push-uint[16|32|64]` functions and ran up against issues with upper limits.

```
error: bad slot #2, expected 32 bit signed integer, got 3927367768
  in buffer/push-uint32 [src/core/buffer.c] on line 388
```

This fixes the issue. I think it is fair that negative numbers here raise bad slot errors so I modified those tests to test the upper bounds instead.

Well, `buffer/push-uint16` anything `0xFFFF <= x <= 0xFFFFFFF` will push `"\xff\xff"` without error until it exceeds `uint32_t`. We could add a similar `janet_getuinteger16`, `janet_checkuint16` , and `janet_checkuint16range` if that's an issue. Might look better to rename the 32-bit functions for consistency as well in that case so I thought I'd get feedback first if this is enough or what path we should take.

### DIFF
--- a/src/core/buffer.c
+++ b/src/core/buffer.c
@@ -375,7 +375,7 @@ JANET_CORE_FN(cfun_buffer_push_uint16,
         uint16_t data;
         uint8_t bytes[2];
     } u;
-    u.data = (uint16_t) janet_getinteger(argv, 2);
+    u.data = (uint16_t) janet_getuinteger(argv, 2);
     if (reverse) {
         uint8_t temp = u.bytes[1];
         u.bytes[1] = u.bytes[0];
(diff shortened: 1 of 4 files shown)


bakpakin:
The old behavior was certainly confusing. I also thing the behavior of pushing the uint16 integers is wrong, too - we should fix that.

bakpakin:
That said, I should point out that the FFI module has a much more general implementation of pushing bytes to buffers.

For example,

```janet
(ffi/write :uint16 100 @"") # -> @"d\0"
(ffi/read :uint16 @"d\0") # -> 100
```

pnelson:
Thanks for the quick turnaround! Control over endianness was required for my use case but I didn't see a way to do so with the FFI module.

PR #1459, opened 2024-06-16
Tweak some docstrings

sogaiu:
This PR tweaks a couple of docstrings for some dynamic variables.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -2746,8 +2746,8 @@
 (defn- check-project-relative [x] (if (string/has-prefix? "/" x) x))
 
 (defdyn *module-cache* "Dynamic binding for overriding `module/cache`")
-(defdyn *module-paths* "Dynamic binding for overriding `module/cache`")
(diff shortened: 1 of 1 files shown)


sogaiu:
Ah sorry, I see there might be some other docstrings that might improve with a bit of adjustment.

I'll list some other things I've noticed below.

---

* `bundle/install` mentions `bundle-name` but it's not apparent how this gets passed in via arguments (I think it's via `config`).  Also, there is no mention of `path` or `config` in the docstring prose.
* `eval` and `eval-string` have docstrings that don't mention the new ability to pass `env`.  May be the docstrings can be modified to account for this.
* `ev/with-wlock`'s docstring says "read access", but may be this should be "write access"?
* `import` doesn't mention the newly added `:only`
option.  `:only` is mentioned in `merge-module` so may be those docs can be pointed at or the content copy-modified appropriately.

bakpakin:
All reasonable changes, these are bad enough I should probably make a patch release. Thanks @sogaiu

PR #1471, opened 2024-07-03
Add ctrl + left/right arrow support to the REPL

Darazaki:
Hi! Made a small patch for a recurring annoyance I had with the REPL. Only tested on Linux w/ WezTerm & Konsole but should work on any modern terminal

Thank you for making/maintaining such a nice programming language :)

---

Commit Description:

>Ctrl + left/right arrow would simply input "5D"/"5C" into the REPL which was useless and confusing
>
>With this change, it instead goes to the previous/next word which is usually expected in readline-like interfaces


sogaiu:
Tried it out in Void Linux.

Works nicely :)

PR #1472, opened 2024-07-08
Add default output path for --compile (-c)

### LINKED ISSUE #1451: `janet -c source` with missing output files should give informative error message
This is my first issue; apologies if I get style or substance wrong. I absolutely adore Janet.

Desired behaviour (if not exactly expected): 
```
$ janet -c source.janet 
> Error: expected output image path`
```

Actual behaviour: 
```
$ janet -c source.janet
> error: expected integer key for array in range [0, 2), got 2
  in c-switch [boot.janet] (tailcall) on line 4066, column 18
  in cli-main [boot.janet] on line 4102, column 13
```
I feel like giving an internal stack trace on a user error like this is not ideal.

Alternative, possibly desired behaviour, is to use the source filename for the new image file:
```
$ janet -c source.janet
> Successfully builds source.jimage
```

Andriamanitra:
Changes behavior of `-c` flag such that:
* `janet -c foo.janet` writes to `foo.jimage`
* `janet -c foo` writes to `foo.jimage`

Previously these would cause a confusing error due to out of bounds access to the args array (see #1451).

The default is only used if there are no more arguments after the input file – I didn't add any special handling for edge cases like `janet -c foo.janet -h` (which uses '-h' as the output file).

closes #1451

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -4475,7 +4475,13 @@
      "c" (fn c-switch [i &]
            (def path (in args (+ i 1)))
            (def e (dofile path))
-           (spit (in args (+ i 2)) (make-image e))
+           (def output-path
+             (if (< (+ i 2) (length args))
+               (in args (+ i 2))
+               (string
+                 (if (string/has-suffix? ".janet" path) (string/slice path 0 -7) path)
+                 ".jimage")))
+           (spit output-path (make-image e))
            (set no-file false)
            3)
      "-" (fn [&] (set handleopts false) 1)


sogaiu:
Gave the changes a try.

Both invocations seemed to work :+1: 

When a file named `foo` did not exist I got an error like:

```
error: could not find file foo
  in dofile [boot.janet] on line 2966, column 5
  in c-switch [boot.janet] (tail call) on line 4477, column 19
  in cli-main [boot.janet] on line 4520, column 13
```

which seemed helpful.

Issue #1470, opened 2024-06-28
Warning/error on seq with no body


sarna:
It'd be nice if Janet warned you (threw an error, printed a warning, ...) when you don't provide the body for `seq`, for example:
```janet
(seq [i :range [0 5]])
```
Right now it will happily return `@[nil nil nil nil nil]`. I can't think of a reason why a user wanted to write it this way.

The same thing happens with `loop`, not sure about other looping constructs.

sogaiu:
I guess if one wanted 5 nils, it would be clearer to express it like:

```janet
(seq [i :range [0 5]] nil)
```

sogaiu:
@sarna I think [this commit](https://github.com/janet-lang/janet/commit/368b891499b5b89c1864bea688189c54acd8333e) may have a relevant change:

```
$ janet
Janet 1.35.2-342a29c7 linux/x64/gcc - '(doc)' for help
repl:1:> (seq [i :range [0 5]])
repl:1:1: compile warning (normal): empty loop body
@[nil nil nil nil nil]
```

sogaiu:
Although the change introduced makes sense for `seq`, I'm not sure it does for `loop`.

I found an example [here](https://github.com/MikeBeller/janet-benchmarksgame/blob/40d5d50dfae34857270eae828d1d441d82b8a60e/knucleotide/knucleotide2.janet#L6-L7) where `loop` is being used for side-effects:

```janet
    (loop [line :iterate (file/read infile :line)
           :until (string/has-prefix? ">THREE " line)])
```

sogaiu:
FWIW, in the above case, the warning (that shows up on stderr) that appears for an empty-body `loop` broke a "driver" program that was parsing stderr.  It had expectations that only certain output would show up on stderr (i.e. that coming from progams being executed).

I think in this case, the warning should actually not happen.

PR #1476, opened 2024-07-18
Do not warn when loop body is empty

sogaiu:
This PR removes the warning in `loop` for an empty body.

Although it is nice to check for an empty body in `seq`, `catseq`, `tabseq`, and `generate`, in the case of `loop`, there are legitimate uses for having an empty body.

An example from the wild and some discussion can be seen [here](https://github.com/janet-lang/janet/issues/1470#issuecomment-2232688409).

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -652,7 +652,6 @@
   The `loop` macro always evaluates to nil.
   ```
   [head & body]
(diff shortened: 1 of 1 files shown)


pepe (approved):
It makes sense to me and LGTM.

Issue #1484, opened 2024-08-11
Awkwardness of `dest` argument for `bundle/add-bin`?


sogaiu:
I'm trying out `bundle/add-bin` and found that I ended up expressing a call with a `dest` argument like this:

```janet
(defn install
  [manifest &]
  (bundle/add-bin manifest 
                  "git-some-janets.janet"
                  # dest
                  "bin/git-some-janets"))
```

Note the prefix of `bin/` for the `dest` argument.

I think it would be nicer to be able to say:

```janet
(defn install
  [manifest &]
  (bundle/add-bin manifest 
                  "git-some-janets.janet"
                  # dest
                  "git-some-janets"))
```

If this is non-problematic in theory, is it too late to change `bundle/add-bin`'s behavior?

---

On a side note, it looks like `bin` and `bundle` are "special" directories under `(dyn *syspath*)`, i.e. probably it's best if people don't create a project that stores their library code in those directories?

There already seem to be some other "special" directories like `.cache`, `.manifests`, etc. in my environment.

I wonder whether there could be a `.bin` and `.bundle` instead of `bin` and `bundle`...thoughts?

_Update_: Spelling things out a bit differently, adopting a convention like `.<name>` for all "special" directories might make it easier to avoid accidental use by 3rd parties, be more consistent with the pre-bundle setup, and makes it easier to programmatically operate on "special" vs "non-special" directories under `(dyn *syspath*)`.  May be there are some downsides too (possibly breaking some code?) I'm not thinking of.

bakpakin:
I'm ok with the change for the default of the `dest` argument for `add-bin`, but not sure if the hidden directories are really needed. Files in the `bin` directory are meant to be run and you absolutely _can_ put files in there. The various files in the `bundle` directory are just Janet modules that you can import. The hidden files are from JPM and are things that you either should need to look at or shouldn't mess with.

sogaiu:
TLDR; Thanks for your thoughts, perhaps as you seem to be hinting at, it may be unlikely that problems will arise with the current arrangement.

---

I'm going to spell things out a bit more below(, partly for my own understanding).

[...]

PR #1485, opened 2024-08-15
Prefix bin+sep to add-bin dest argument

sogaiu:
This PR is an attempt to address #1484.

With this change, one should be able to use `bundle/add-bin` like this:

```janet
(bundle/add-bin manifest
               "git-some-janets.janet"
               # dest argument
               "git-some-janets")
```

instead of having to express the same intent like this:

```janet
(bundle/add-bin manifest
                "git-some-janets.janet"
                # dest argument
                "bin/git-some-janets")
```

That is, if the optional `dest` argument is specified to `bundle/add-bin`, it will have `bin` and an appropriate separator prefixed to it before being passed to `install/add-file`.

I tested the new version with `spork` and one of my own repositories and it seems to work.


pepe (approved):
LGTM

PR #1513, opened 2024-10-16
Mention nth and only-tags in changelog

sogaiu:
This PR adds mentioning of the new `nth` and `only-tags` PEG specials to the unreleased section of the changelog.


pepe (approved):
LGTM

PR #1514, opened 2024-10-20
Cast arguments to `int32_t` before passing to `janet_formatb` with `%d` format specifier

ArtSin:
`s->line` and `s->column` in `delim_error` are `size_t`, which is typically 64-bit, but `va_arg` in `janet_formatbv` reads `int32_t` for `%d`.


bakpakin:
Thanks, the CI failure look spurious here.

PR #1532, opened 2024-12-13
file/open: check if directory

### LINKED ISSUE #1530: include filename in `error: could not read file` message
I was building a project with `janet build` mistakenly instead of `jpm build` and got
```
error: could not read file
  in file/read [src/core/io.c] on line 193
  in chunks [boot.janet] on line 3009, column 24
  in run-context [boot.janet] on line 2653, column 12
  in dofile [boot.janet] on line 3032, column 5
  in cli-main [boot.janet] on line 4645, column 17
```
as a result.

This isn't particularly helpful and it would be neat if the message included the name of the file it tried to read.

Looks like the error comes from https://github.com/janet-lang/janet/blob/83e8aab2897aa2982db48668e85757eef680ea5e/src/core/io.c#L188 which only has access to a `JanetFile`, which doesn't include the filename, only a `FILE*`: https://github.com/janet-lang/janet/blob/83e8aab2897aa2982db48668e85757eef680ea5e/src/include/janet.h#L1248-L1251

So possible solutions could be
1. store the filename the file was opened by in `JanetFile`
2. attempt to find the filename associated to the file descriptor. (could be `/proc/self/fd/NNN` or `F_GETPATH` or `F_GETPATH` or `GetFileInformationByHandleEx` on windows? would need to investigate)
if this is in scope for janet

strangepete:
Adds fstat directory test after fopen, which can return non-NULL when passed a directory name on Linux.

Tested on Ubuntu 22.04 (+valgrind), & Windows 10

Should close #1530

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -31,6 +31,7 @@
 
 #ifndef JANET_WINDOWS
 #include <fcntl.h>
+#include <sys/stat.h>
 #include <sys/wait.h>
 #include <unistd.h>
 #endif
@@ -164,6 +165,14 @@ JANET_CORE_FN(cfun_io_fopen,
     }
     FILE *f = fopen((const char *)fname, (const char *)fmode);
     if (f != NULL) {
+#ifndef JANET_WINDOWS
+        struct stat st;
+        fstat(f->_fileno, &st);
+        if (S_ISDIR(st.st_mode)) {
+            fclose(f);
+            janet_panicf("cannot open directory: %s", fname);
+        }
+#endif
         size_t bufsize = janet_optsize(argv, argc, 2, BUFSIZ);
         if (bufsize != BUFSIZ) {
             int result = setvbuf(f, NULL, bufsize ? _IOFBF : _IONBF, bufsize);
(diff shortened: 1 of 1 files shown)


strangepete:
As an afterthought, it's not great that this introduces a different message for the same error on multiple platforms (ie, Windows says 'could not find file')

jakobhellermann:
I can confirm that this does improve the error message on linux
```
error: cannot open directory: build
  in file/open [src/core/io.c] on line 153
  in dofile [boot.janet] on line 3002, column 12
  in cli-main [boot.janet] on line 4645, column 17
```

bakpakin:
While I like the error message, something bothers me about making an extra system call every time we open a file. That said, python does the same thing, and there is always `os/open` for less abstracted IO.

strangepete:
I agree, though looking at the field it seems everyone is fine doing this, which is crazy to me...open/fopen give multiple options to ensure a target is **definitely** a directory; I'm very curious what the use case is to 'read' a directory in binary...

Alternatively, as Jakob originally suggested just save the file name for use later if an error pops up. I assume nobody can open enough files to make the memory usage matter?

PR #1541, opened 2024-12-31
Brief: Add Arm64 .msi support on Windows

rwtolbert:
Small update to add Windows on Arm64 support. This does require changing the (os/arch) value from 'aarch64' to 'arm64' just for Windows on Arm.

Also requires the latest version 3.14 of the WiX toolset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ Alternatively, install the package directly with `pkgin install janet`.
 
 To build an `.msi` installer executable, in addition to the above steps, you will have to:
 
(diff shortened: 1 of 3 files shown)


bakpakin:
Looks good to me, but is there any real need to change `:aarch64` to `:arm64` for windows only? afaik they are the same thing, with aarch64 being the more "official" name for the ISA.

rwtolbert:
just pushed an update that removes the change in (os/arch). 

WiX uses "arm64" and Python uses "arm64". Clang and Rust use "aarch64" so there is no clear answer.

PR #1568, opened 2025-03-10
Update Makefile and meson to build with msys2 gcc and clang

rwtolbert:
This updates the Makefile and meson.build to allow building with both gcc and clang on MSYS2/x86_64 and clang on MSYS2/ARM64.


bakpakin:
Thanks for not forgetting to update meson as well. LGTM.

PR #1582, opened 2025-04-09
Fix absence of pthread_cancel in Android

pyrmont:
This PR fixes a build error on Android because `pthread_cancel` is [not implemented](https://android.googlesource.com/platform/bionic/+/refs/heads/main/docs/status.md#posix) in Bionic, the library Google uses for Android that implements parts of C like `pthread` (thanks to @sogaiu for bringing this to my attention). This PR uses signals for Android as a workaround.


bakpakin:
Looks good to me. As far as pthread_cancel, this is why Janet doesn't really support canceling threads in the first place, as it is difficult to coordinate with threads and make sure that clean up code gets run, such as decrementing refcounts.

Issue #809, opened 2021-09-17
`(defn fun* [] ())` when displayed with `doc` shows as `(fun)` instead of `(fun*)`


llmII:
As per title, markdown issue?

In an REPL do:
```
(defn fun* [] ())
(doc fun*)
```

Output is:
```
Janet 1.17.1-468a31f5 freebsd/x64 - '(doc)' for help
repl:1:> (defn fun* [] ())
<function fun*>
repl:2:> (doc fun*)


    function
    repl on line 1, column 1

    (fun)


nil
repl:3:> 
```

bakpakin:
Yes, `*` is considered to be a delimiter for italics in markdown, and the doc formatter does not do any look ahead for parsing. We probably should consider removing the first line from markdown semantics when printing or something similar.

PR #1590, opened 2025-04-16
Expose docstring parse tree via `doc-parse`

pyrmont:
This PR exposes the parse tree generated as part of Janet’s docstring formatting. To achieve this, it adds the `doc-parse` function to `boot.janet` and more cleanly separates the parsing and rendering phases that were previously both in `doc-format`. It also improves Markdown parsing and adds a test suite.

This PR has benefitted greatly from work that @sogaiu has been doing in parallel to improve Janet's documentation.

## Background

Janet supports a subset of Markdown[^1] in its docstrings. This support is manifested via the `doc-format` function. This allows a user to write an unordered list in a docstring and have the docstring correctly wrap the items in the list when displaying the docstring in the REPL:

```text
repl:1:> (doc match)


    macro
    boot.janet on line 1888, column 1

    (match x & cases)

    Pattern matching. Match an expression x against any number of
    cases. Each case is a pattern to match against, followed by an
    expression to evaluate to if that case is matched. Legal patterns
    are:

    * symbol -- a pattern that is a symbol will match anything, binding
      x's value to that symbol.

    * array or bracket tuple -- an array or bracket tuple will match
      only if all of its elements match the corresponding elements in
      x. Use & rest at the end of an array or bracketed tuple to bind
      all remaining values to rest.

    * table or struct -- a table or struct will match if all values
      match with the corresponding values in x.

    <...>


```

`doc-format` allows some customisation but to a limited degree.

## Rationale

There are two scenarios where exposing the parse tree would be useful to me:

1. **Website API documentation.** The Janet website uses the docstrings for functions in Core and Spork. To display these docstrings, it renders the docstring as plaintext and then wraps these in HTML’s `<pre>` tag. While this is fine for most desktop environments, on smaller-width screens (like phones), wrapping ‘breaks’ if the width of the display is less than the width at which the docstring was rendered to plaintext. I have been discussing this with @sogaiu and if the parse tree were accessible, we think it would be relatively trivial to transform this into the Mendoza template language so that the text included more semantic information in the markup (e.g. lists and code blocks).

2. **REPL tooling for text editors**. I am (very slowly) working on an extension for the Conjure Neovim plug-in that would provide a richer user experience by connecting to an upgraded REPL environment. One of the most common uses for this kind of tool is to look-up documentation for a function. This is possible at the moment but the docstring which is returned is in plaintext and requires parsing to display in an elegant way.

## Implementation
 
This PR adds a new function, `doc-parse`:

```janet
(doc-parse str)
```

If a user calls `doc-parse` with a string, the return value is the parse tree that used to be created by functions in `doc-format`.

The current implementation of `doc-format` creates a parse tree using a single pass. This pass analyses each line, first identifying whether it is a block element (a list, a code block) and then tokenising each line into a sequence of _word tokens_. A _word token_ is a 2-element tuple consisting of (1) a string of bytes (including punctuation that should be continguous with that word, such as `.`) and (2) the number of bytes in the string. During each tokenisation step, delimiters for inline elements are inserted into the word token. While parsing and rendering is mostly separated, this action mixes an aspect of rendering into parsing.

This PR avoids this mixing by introducing the concept of a _facet token_[^2]. A _facet token_ is a 2- or 3-element tuple that consist of (1) a keyword (representing the kind of facet), (2) an integer (representing the position at which the facet applies to the following word token) and, for opening facets only, (3) a string (representing the delimiter used in the original source). Facet tokens come before the word tokens to which they apply. It is possible to have multiple facet tokens apply to one word token.

As an example,

```janet
(doc-parse “*foo*”)
```

would produce

```janet
@[((:italics 0 "*") (:italics 3) ("foo" 3))]
```

The two facet tokens representing where the facet begins and where the facet ends occur before the word token to which they apply.

In addition to this larger change, the PR also does the following:

1. **Supports tight/loose lists**. Markdown distinguishes between tight lists (where list elements occur on consecutive lines) and loose lists (where list elements are separated by blank lines). This PR adds a metadata element to the list array to record this information.

2. **Supports nested blocks inside lists**. Markdown allows blocks to be nested inside lists. This PR supports this for both tight and loose lists. This would allow the docstring for [`pm/resolve-bundle`](https://github.com/janet-lang/spork/blob/2d4638cfb843589485dde75ab79c1eda7ccd1f46/spork/pm.janet#L71-L75) in Spork to render as it seems like it might have been intended (once the necessary indentation is applied):

   ````janet
   (defn resolve-bundle
     ```
     Convert any bundle string/table to the normalized table form. `bundle` can be any of the following forms:

     * A short name that indicates a package from the package listing.
     * A URL or path to a git repository
     * A URL or path to a .tar.gz archive
     * A string of 2 parts separated by "::" - {type}::{path-or-url}
     * A string of 3 parts separated by "::" - {type}::{path-or-url}::{tag}
     * A table or struct with the following keys:

     * `:url` or `:repo` - the URL or path of the git repository or of the .tar.gz file. Required.
     * `:tag`, `:sha`, `:commit`, or `:ref` - The revision to checkout from version control. Optional.
     * `:type` - The dependency type, either `:git`, `:tar`, or `:file`. The default is `:git`. Optional.
     ```
     [bundle]
   [...])
   ````

3. **Recovers from incorrectly categorised punctuation.** The current implementation does not recover if punctuation is treated as Markdown syntax incorrectly. This necessitates a hack (see #809)  so that function names that include `*` are not treated as Markdown syntax. This PR includes a simple algorithm that restores punctuation characters that are unmatched at the time a block element is finished being parsed which removes the need for the hack.

4. **Hides backticks**. While delimiting certain text in prose as being ‘code’ can aid in readability, I think this argument is less persuasive in typically monospaced environments like command-line REPLs. For this reason, this PR hides the backticks used in docstrings to delimit text as code.

5. **Adds a test suite**. To protect against breaking changes, this PR adds a test suite for docstring parsing and rendering.

## Limitations

This PR adds another function to the Core API. It is also necessarily more complex than the current implementation.

[^1]: Technically, Janet is not a strict subset of Markdown. The `_` delimiter is recognised as distinct from `*` delimiter. In Markdown, text delimited by these two characters is semantically equivalent. This is not the case in Janet.

[^2]: The term ‘facet’ is borrowed from a similar concept in the [AT Protocol](https://github.com/bluesky-social/atproto/blob/231eb4826512c413fbe0165fc47baf11c8f0695d/lexicons/app/bsky/richtext/facet.json).


pyrmont:
I've got an idea to improve the robustness of parsing of the inline elements so have converted to draft while I work on that.

bakpakin:
Thanks for the updates here! The doc formatting and parsing code is always a bit tedious since it should work even with PEGs disabled. I don't see any issue with this other than the facet token idea seeming strange to me, and the given example doesn't sit right. Shouldn't it be something like `@[(:italics "*") "foo" (:italics-end "*")]` so that it can be easily printed to HTML, a terminal, markdown, etc. by iterating over the tokens? If we are going through the effort to parse things, we may as well make it as easy to print/translate as possible. This seems like the primary reason to export parsed documentation in the first place.

pyrmont:
@bakpakin I've been interrupted by a few things in real life. I'll close this for now as I'm not sure when I'll have time to make these changes but I do intend to come back to it :) (Hopefully in the next two weeks but not sure.)

sogaiu:
It has been a while since this PR was closed, but due to recent discussions I thought it might be worth revisting.

I'm trying the latest code in https://github.com/janet-lang/janet/pull/1590/commits/3b5964441d901ce925656b973332153a8a15eab8 and wondering a bit about the following results I'm seeing:

```janet
(doc-parse `**Example**`)
# =>
@[[[:bold 0 "**"] [:bold 7] ["Example" 7]]]

(doc-parse ``*emphasis text*``)
# =>
@[[[:italics 0 "*"] ["emphasis" 8] [:italics 4] ["text" 4]]]

(doc-parse "`(a-fn arg-0 arg-1)`")
# =>
@[[[:code 0 "`"] ["(a-fn" 5] ["arg-0" 5] [:code 6] ["arg-1)" 6]]]
```

Are the latter two intended?

Looking at [some of the provided tests](https://github.com/janet-lang/janet/pull/1590/changes#diff-e18ceb278c243d372c409f608c75f565b925f9638e81f5e3e558949b85480fcaR55-R59) (translated to match the style above):

```janet
(doc-parse "foo _bar_ **baz qux**")
# =>
@[[["foo" 3]
   [:underline 0 "_"] [:underline 3] ["bar" 3]
   [:bold 0 "**"] ["baz" 3] [:bold 3] ["qux" 3]]]]

(doc-parse "foo **_bar_ baz qux**")
# =>
@[[["foo" 3]
   [:bold 0 "**"] [:underline 0 "_"] [:underline 3] ["bar" 3] ["baz" 3]
   [:bold 3] ["qux" 3]]]]
```

I guess the answer is "yes" and what are shown above are examples of:

> Facet tokens come before the word tokens to which they apply. It is possible to have multiple facet tokens apply to one word token.

in action.

Trying to ascertain if that is different from:

> A facet token comes before the single word token to which it applies. It is possible to have multiple facet tokens apply to one word token.

So for the `"foo _bar_ **baz qux**"` example:

```janet
(doc-parse "foo _bar_ **baz qux**")
# =>
@[[["foo" 3]
   [:underline 0 "_"] [:underline 3] ["bar" 3]
   [:bold 0 "**"] ["baz" 3] [:bold 3] ["qux" 3]]]]
```

the face token `[:bold 0 "**"]` applies to `["baz" 3]`, but not to `["qux" 3]`, while `[:bold 3]` does apply to `["qux" 3]`.

i.e. the "stuff in between" is unadorned with facet tokens in some sense.  Perhaps that's clearer in this example:

```janet
(doc-parse "`(a-fn arg-0 arg-1)`")
# =>
@[[[:code 0 "`"] ["(a-fn" 5] ["arg-0" 5] [:code 6] ["arg-1)" 6]]]
```

``[:code 0 "`"]`` applies to `["(a-fn" 5]` but not to `["arg-0" 5]`, while `[:code 6]` applies to `["arg-1)" 6]`.  The "stuff in between" here is `["arg-0" 5]` and there are no facet tokens that apply to it.

---

I realize that there was [an alternative idea](https://github.com/janet-lang/janet/pull/1590#issuecomment-2814270918) [1] suggested to use something like:

```janet
@[[:italics "*"] "foo" [:italics-end "*"]]
```

instead of:

```janet
@[[[:italics 0 "*"] [:italics 3] ["foo" 3]]]
```

I think notation like `<i>hello my friends</i>` (an analog of the alternative idea) suggests that "italics" applies to all of the "contained" content.  The facet token idea is in some sense "lower level", it doesn't seem to be directly implying anything about "all of the stuff between", but rather signaling things about the beginning and ending tokens.

If the `doc-parse` examples above are intentional, I think another reason to consider the alternative is that the alternative might come across as more familiar.

---

Looking over the content of what ended up in the proposed version of `doc-format`, I get the sense that the facet token idea might mean less changes to the existing code (compared to if the alternative idea were to be implemented) as it looks like the main changes (to `doc-format`) were:

* break out (and modify) `doc-parse`
* change `emit-node` to accomodate output from `doc-parse`

All the other bits in `doc-format`, i.e. other `emit-*` functions and the "main loop" look to be unchanged.

Just a guess though.

---

[1] I wonder if the `"*"` portion of `[:italics-end "*"]` is needed.  Can we get by with just `[:italics-end]`?  For that matter, do we need `-end`?  May be `[:italics]` by itself could work?  

May be that's doable if the "opening" token always has at least one attribute and the "closing" token always has no attributes.

sogaiu:
So if something like:

```janet
@[[:italics "*"] "foo" [:italics-end "*"]]
```

is [acceptable](https://github.com/janet-lang/janet/pull/1590#issuecomment-2814270918), then why not go for something hiccup-like, e.g.

```janet
@[[:italics @{:delim "*"} ["foo"]]]
```

or similar?

Is it that a hiccup-like option might be too costly because of the dictionaries...and/or may be the rendering cost might be needlessly expensive?

Or may be it would be ok?

---

I guess since construction of what parsing outputs involves looking at all of the characters anyway, keeping the lengths of things, e.g.:

```janet
["bar" 3]
```

allows us to look up a length instead of compute it?

sogaiu:
Further consideration has turned up the following sort of case:

```janet
(doc-parse "`hello`, there")
# =>
@[[[:code 0 "`"] [:code 5] ["hello," 6] ["there" 5]]]
```

The facet token (and original?) approach seems to make indicating where the closing delimiter should end up within a word token straight-forward (e.g. after the "o" but before the "," in the example above). That is, it's a matter of specifying a smaller number for `:code` than for `"hello,"` (e.g. here 5 instead of 6) .

I'm not sure how one should handle this via the alternative idea:

```janet
@[[:code "`"] "forms," [:code-end "`"]]
```

How do we indicate that the closing delimiter should be placed before the end of the text (and not at or after), i.e. after the last letter but before the trailing punctuation (so here, between the "s" and the ",") [1]?

May be parse as something like this?

```janet
@[[:span]
    [:code "`"] "forms" [:code-end]
    ","
  [:span-end]]
```

---

[1] I've phrase things above in terms of "punctuation", but AFAIU, there's no explicit handling of punctuation, but rather [the end of a word token is determined based on if there is a following space or newline character](https://github.com/janet-lang/janet/blob/bef6cb5d741308bf95b6b8ada425b1b533a74ad2/src/boot/boot.janet#L3452).  So implicitly in many cases, that ends up defining where the next word token might start I suppose.

PR #1601, opened 2025-06-14
Use strnlen when checking for null byte

### LINKED ISSUE #1600: "string contains zeros" when formatting buffer
When formatting a buffer using `%s` and either width or precision, I often get the error:

```
error: string contains zeros
```

Some example code that lets me reproduce this reliably, but not 100% of the time:

```
(printf "%1s" @"")
(printf "%.1s" @"")
```

If I omit both width and precision, I never see the error. I've been able to reproduce this error on Linux (Alpine and Debian Docker images), but not on MacOS.

edsrzf:
Buffers are not always null-terminated, so using `strlen` here could cause reads past the end of the buffer and cause Janet to incorrectly report that the string contains zeros because the lengths don't match.

There may be other instances to fix, but this fixes #1600.

### DIFF
--- a/src/core/pp.c
+++ b/src/core/pp.c
@@ -1066,7 +1066,7 @@ void janet_buffer_format(
                     if (form[2] == '\0')
                         janet_buffer_push_bytes(b, s, l);
                     else {
-                        if (l != (int32_t) strlen((const char *) s))
+                        if (l != (int32_t) strnlen((const char *) s, l))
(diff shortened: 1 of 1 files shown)


bakpakin:
Looks good. Most places where we convert buffers to strings, we should be calling `janet_getcbytes` or similar to ensure that final 0 byte of padding.

Issue #950, opened 2022-04-14
more informative error message if used def instead of defn


uvtc:
If the user accidentally types `def` rather than `defn`:

~~~janet
(def foo
  [x]
  (print (* x x)))

(foo 3)
~~~

The error message is:

~~~
$ ./foo.janet 
./foo.janet:6:10: compile error: unknown symbol x
~~~

I think it would be a nice feature if the compiler was able to suggest that the user may have meant to type `defn`.

I notice that if I change it to:

~~~janet
(var x 5)

(def foo
  [x]
  (print (* x x)))

(foo 3)
~~~

the error message is now

~~~
$ ./foo.janet 
./foo.janet:5:1: compile error: could not add metadata to binding
~~~

sogaiu:
I wonder if this could work well as part of a linter.

uvtc:
I see linter linting was brought up in #644 .

uvtc:
Whoops. Linting docs are at https://janet-lang.org/docs/linting.html .

sogaiu:
I was thinking of an external linter program.

IIUC @pepe has [some bits](https://git.sr.ht/~pepe/jakoune/tree/master/item/jakoune/lint.janet) to do some kinds of linting.

A quick way to get an idea of what it currently does is to examine the calls to `emit`, e.g.
```
(emit ;pos "warning" "`tracev`")
```
let's one know that there is at least one `tracev` form in one's program (which perhaps one might not want to commit).

sogaiu:
@uvtc I tried [b4ec168](https://github.com/janet-lang/janet/commit/b4ec16840112871a0384deda9708b8184e2d355f):
```
$ janet
Janet 1.21.2-b4ec1684 linux/x64 - '(doc)' for help
repl:1:> (def foo
repl:2:(>   [x]
repl:3:(>   (print (* x x)))
repl:1:1: compile error: unexpected form - did you intend to use defn?
```

uvtc:
Thanks!

sogaiu:
Just wanted to add that this has helped me on multiple occasions :)

PR #1605, opened 2025-06-24
Patch try to accept (try body ([] catch-body))

### LINKED ISSUE #1603: Weird error message when trying to compile (try 1 (error))
```
❯ janet
Janet 1.38.0-d96e5848 linux/x64/gcc - '(doc)' for help
repl:1:> (try 1 (error))
repl:1:1: compile error: unexpected type in destructuring, got 101
```

The error message is weird. (got 101)

iacore:
Fixes #1603 

Also makes `(try 1 ([] (error "hi")))` valid.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -290,22 +290,6 @@
   (array/concat accum body)
   (tuple/slice accum 0))
 
-(defmacro try
-  ``Try something and catch errors. `body` is any expression,
-  and `catch` should be a form, the first element of which is a tuple. This tuple
-  should contain a binding for errors and an optional binding for
-  the fiber wrapping the body. Returns the result of `body` if no error,
-  or the result of `catch` if an error.``
-  [body catch]
-  (let [[[err fib]] catch
-        f (gensym)
-        r (gensym)]
-    ~(let [,f (,fiber/new (fn :try [] ,body) :ie)
-           ,r (,resume ,f)]
-       (if (,= (,fiber/status ,f) :error)
-         (do (def ,err ,r) ,(if fib ~(def ,fib ,f)) ,;(tuple/slice catch 1))
-         ,r))))
-
 (defmacro protect
   `Evaluate expressions, while capturing any errors. Evaluates to a tuple
   of two elements. The first element is true if successful, false if an
@@ -351,6 +335,23 @@
                  (tuple 'do (tuple 'def $fi fi)
                         (tuple 'if $fi $fi ret))))))
   ret)
+
+(defmacro try
+  ``Try something and catch errors. `body` is any expression,
+  and `catch` should be a form, the first element of which is a tuple. This tuple
+  should contain a binding for errors and an optional binding for
+  the fiber wrapping the body. Returns the result of `body` if no error,
+  or the result of `catch` if an error.``
+  [body catch]
+  (assert (and (not (empty? catch)) (indexed? (catch 0))) "the first element of `catch` must be a tuple or array")
+  (let [[err fib] (catch 0)
+        r (or err (gensym))
+        f (or fib (gensym))]
+    ~(let [,f (,fiber/new (fn :try [] ,body) :ie)
+           ,r (,resume ,f)]
+       (if (,= (,fiber/status ,f) :error)
+         (do ,;(tuple/slice catch 1))
+         ,r))))
 
 (defmacro with-syms
   "Evaluates `body` with each symbol in `syms` bound to a generated, unique symbol."


pepe:
Perhaps I don't understand it entirely, but what is the purpose of `(try 1 ([] (error "hi")))`? Frankly, I do not see any value in the change.

iacore:
I use it to change the error message. For example, `(try (arr 0) ([] (error "array empty"))`

sogaiu:
I may be missing some of what's going on in this discussion (^^; -- but below are some related thoughts.

For `try`, IIUC, usually a name is chosen for the error portion of the catch if we're going to use it:

```janet
(try
  (error "hey")
  ([e]
    (print "hi:" e)))
```

If we don't want to have to choose a name (because we aren't going to use the information) we can do:

```janet
(try
  (error "hey")
  ([_]
    (print "hi")))
```

but:

```janet
(try
  (error "hey")
  ([]
    (print "hi")))
```

seems ok to me too.

---

There's another part of the PR which is to improve the error message, which IIUC is what #1603 was originally mostly(?) about:

> The error message is weird. (got 101)

For the reported case, I did find the error message to be puzzling.

pepe:
For the usage, I see _ totally fine, even better to be honest.

 For the error, that is true, but then there are  more places, where to guard it. I mean more macros. Do we want to have these guards everywhere? But I am ok with that, not the issue I had on my mind. Sorry for not being clear. 

Not my call anyway, just my opinion.

sogaiu:
> For the usage, I see _ totally fine, even better to be honest.

Yeah, I am ok with using `_` as well.

Just curious why you think using `_` is better.  It's not obvious to me and I'd like to see if I'm missing something :)

> For the error, that is true, but then there are more places, where to guard it. I mean more macros. Do we want to have these guards everywhere?

It's a good question.

On a somewhat related note, I wasn't so keen on [this](https://github.com/janet-lang/janet/issues/950) initially, but in that case it turned out later that I benefited from it repeatedly (still do occasionally!), so I am not so clear on when it's better :sweat_smile:

May be for the `def` / `defn` case, it's fairly likely to happen whereas it's not as clear in this case?

pepe:
It is better for me, cause it signals to me that something is omitted. TBH I love the `[&]` the most. 

For `def` and `defn`, I agree very much. And also, I understand the need for this guard. It is just that I am not sure about asserts in macros. I guess we can build a debate about linting in this context?

sogaiu:
Thanks for sharing.

It seems like the proposed change gives behavior that's similar to using `[&]` -- though may be I misunderstood?

```
(defn f [& args] :smile)

(f)
# =>
:smile: 

(f 1)
# =>
:smile

(f 1 2)
# =>
:smile
```

May be you were saying that there is something positive about the proposed behavior?

---

Taking a closer look at `try`...the way it is currently, it requires the catch part to have at least one element:

```janet
[e]
```

I presume we all know that it can have two though:

```janet
[e fib]
```

I assume that one reason it is this way is because one doesn't always want to use the associated fiber.  If that's the main reasoning, it seems reasonable to me to also make `e` optional.  That seems consistent in a way.  I guess that would be like `[&opt e fib]` - close to `[&]`, but would error with 3 or more arguments?.

Alternatively, one could have always required two elements in the tuple, i.e. `[e fib]`.  That also feels like it has a similar type of consistency to me.  If one didn't want either argument, one could express the catch part of the try as `[_ _]`...though using `_` more than once seems somewhat uncommon.

With the way things are currently, `[e]` doesn't explicitly signal that the associated fiber is being omitted.  So may be similar to `[e &opt fib]`?  In terms of having to remember things, I think this is the one that I'd be most likely to have trouble with :sweat_smile: 

In terms of not breaking things with existing code, allowing `[]` seems ok to me (though I do wonder how common that use case is likely to be -- perhaps when quickly writing code it's more common?).

---

I'm even less sure about `assert` use in macros :sweat_smile:  I went through `boot.janet` though and I do see a few macros that use `assert` (`defdyn` and may be `ffi/defbind-alias`?).  Not many though!

Regarding linting, do you think it would be practical for the checking to be done via flycheck?  I guess not everyone uses that -- though I am saved by it quite frequently.

---

On a related note, `try` lacks examples at the website docs.  Having some examples there possibly could help for future occasions, but I guess one would have to think to look to benefit.

iacore:
> On a related note, `try` lacks examples at the website docs. Having some examples there possibly could help for future occasions, but I guess one would have to think to look to benefit.

I would like to see example in `(doc try)`. Reading the usage in natural language ... need to wrap my head around that.

> Regarding linting, do you think it would be practical for the checking to be done via flycheck? I guess not everyone uses that -- though I am saved by it quite frequently.

flycheck for EMACS? Not sure how to use that.

pepe:
https://janet-lang.org/docs/linting.html for the reference.

sogaiu:
> I would like to see example in (doc try).

This has been touched on [elsewhere](https://github.com/janet-lang/janet/pull/1555), but I believe it's pretty much a policy to not include examples in the docstrings.

However, we've been adding examples to [this page](https://janet-lang.org/api/index.html).  (To get an idea of what this can look like, please see [this portion for `all`](https://janet-lang.org/api/index.html#all).)

We can add some examples for `try` there.

On a side note, earlier this year a number of us [made an effort to increase the coverage](https://github.com/sogaiu/janet-examples/blob/9428798e12ed6a4a4e88ff805f06969f8dd93331/doc/janetuary-examples-2025.md).  There is still a lot that could be added for sure.

> Reading the usage in natural language ... need to wrap my head around that.

FWIW, there is some prose on the website for `try` in [this section](https://janet-lang.org/docs/fibers/error_handling.html#Two-error-handling-forms-built-on-fibers).

Hmm, it's interesting that the prose there seems to be describing what might be considered the behavior being proposed in this PR:

> if any error is raised, optionally bind the error and the raising fiber in its second clause.

sogaiu:
> flycheck for EMACS? Not sure how to use that.

Ah sorry. `flycheck` happens to also be the name of the feature in `janet` available via `-k` for using the built-in linting that pepe linked to.

For editor support, I've had some success getting it to work with Emacs and Neovim [1], and I think pepe had something for kakoune.  May be other people have had success with other editors too.

---

[1] For Emacs, there is at least [flycheck-janet](https://github.com/sogaiu/flycheck-janet) and for Neovim using [nvim-lint](https://github.com/mfussenegger/nvim-lint) with the following in `init.vim` seems to work here:

```vim
lua << EOF
-- manually triggering -> :lua require("lint").try_lint()
require('lint').linters_by_ft = {
  janet = {'janet'},
}
EOF

lua << EOF
-- TextChangedI might be worth trying at some point
vim.api.nvim_create_autocmd({ "BufRead", "BufWritePost", "InsertLeave" }, {
  callback = function()
    require("lint").try_lint()
  end,
})
EOF

```

sogaiu:
I've made a PR with some initial examples [here](https://github.com/janet-lang/janet-lang.org/pull/300).

Below is how they can end up looking in a web browser on [the aforementioned page](https://janet-lang.org/api/index.html).

Image

iacore:
Nice! Are we going to accept this patch or not?

sogaiu:
Ultimately, I think it's up to @bakpakin.

bakpakin:
LGTM, esp. the addition of the improved error message.

Issue #1614, opened 2025-07-17
net/server use with `:datagram` and error output


sogaiu:
As reported [here](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/Creating.20a.20UDP.20server.20using.20net.2Fserver.20always.20fails/with/528834582), if `janet` is passed a file with the following code:

```janet
(defn handler [conn]
  (print conn)
  (net/close conn))

(net/server "0.0.0.0" 8080 handler :datagram false)
```

The following error output has been reported by a few folks:

```
error: bad stream, expected server socket
  in net/accept-loop [src/core/net.c] on line 828
  in <anonymous> [boot.janet] on line 3883, column 21
```

This seems to be caused by `net/accept-loop` [here](https://github.com/janet-lang/janet/blob/ddc122958be34a18edd754743851536b5a16a3cc/src/boot/boot.janet#L3926).

It seems that `net/accept-loop` isn't meant to work with UDP?

For reference, `net/server` is [currently like](https://github.com/janet-lang/janet/blob/ddc122958be34a18edd754743851536b5a16a3cc/src/boot/boot.janet#L3920-L3927):

```janet
(compwhen (dyn 'net/listen)
  (defn net/server
    "Start a server asynchronously with `net/listen` and `net/accept-loop`. Returns the new server stream."
    [host port &opt handler type no-reuse]
    (def s (net/listen host port type no-reuse))
    (if handler
      (ev/go (fn [] (net/accept-loop s handler))))
    s))
```

---

Perhaps one resolution is for `net/server` to avoid:

```janet
   (if handler
      (ev/go (fn [] (net/accept-loop s handler))))
```

when `type` is `:datagram`.

...and may be the docstring can be adjusted appropriately too.

sogaiu:
Another alternative mentioned in the discussion linked to earlier is to avoid handling `:datagram` in `net/server`.

This might be done by either of:

1. Removing `net/server`'s `type` parameter, -OR-
2. Erroring if the `type` parameter is neither `nil` nor `:stream` (for a more backward compatible alternative)

sogaiu:
https://github.com/janet-lang/janet/pull/1615 was merged so I'm going to close this issue.

Thanks @ifreund!

PR #1615, opened 2025-07-17
net/server: improve error for truthy handler and type :datagram

ifreund:
Since it is invalid to call accept on a datagram socket, net/server always errors if handler is truthy and type is :datagram.

Tweak the implementation to give a better error message in this case.

References: https://github.com/janet-lang/janet/issues/1614

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3919,8 +3919,14 @@
 
 (compwhen (dyn 'net/listen)
   (defn net/server
-    "Start a server asynchronously with `net/listen` and `net/accept-loop`. Returns the new server stream."
+    ``
+    Starts a server with `net/listen`. Runs `net/accept-loop` asynchronously if
+    `handler` is set and `type` is `:stream` (the default). It is invalid to set
+    `handler` if `type` is `:datagram`. Returns the new server stream.
+    ``
     [host port &opt handler type no-reuse]
+    (assert (not (and (= type :datagram) handler))
+            "handler not supported for :datagram servers")
     (def s (net/listen host port type no-reuse))
     (if handler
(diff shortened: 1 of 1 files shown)


sogaiu:
Thanks for the PR!

If `type` is `nil`, `net/listen` behaves as if `type` is `:stream`, right?

Do we want to handle things a bit differently in the `case` for `type` being `nil` and `handler` is not `nil`?

Right now it looks to me like the `case` form might just skip doing anything with `handler` if `type` is `nil`:

```janet
(case type
  :stream (if handler
            (ev/go (fn [] (net/accept-loop s handler))))
  :datagram (if handler
              (error "handler not supported for :datagram servers")))
```

May be we can use `(default type :stream)` or something earlier in the code?  Although I suppose that relies on `net/listen`'s current behavior...may be that's not a problem.

ifreund:
Good catch @sogaiu! I think the `(default type :stream)` approach is fine. It doesn't even really rely on the behavior of `net/listen` to function as documented, the only downside is that it could become inconsistent with `net/listen` if the default type there is changed. That doesn't seem likely to ever happen though in my humble opinion :)

ifreund:
The newest revision avoids using `default` and also avoids binding a socket at all if the arguments are invalid.

sogaiu:
Just a quick note.

Don't know if it would be an improvement, but perhaps an alternative to this expression:

```janet
    (if (and (= type :datagram) handler)
      (error "handler not supported for :datagram servers"))
```

might be to use `assert`?

Will think on it a bit more.

ifreund:
I think I do like the assert a bit more, perhaps this is finally the final version of this patch :)

sogaiu:
LGTM

pepe (approved):
LGTM

Issue #1620, opened 2025-08-11
net/send-to fails when using net/connect with :datagram


tttuuu888:
Hi All,

When writing a UDP client in Janet, using a stream created by net/connect with the net/send-to function results in a "bad stream, expected datagram socket" error.

Problematic Code:
```
# Server
(def server-stream (net/listen "127.0.0.1" 8001 :datagram))
(def buf (buffer/new 1024))
(net/recv-from server-stream 1024 buf)
(print (string "Recv: " buf))
```

```
# Client
(def server-address (net/address "127.0.0.1" 8001 :datagram))
(def client-stream (net/connect "127.0.0.1" 8001 :datagram))
(net/send-to client-stream server-address "Test message")
```

Error message on Client net/send-to function:
```
Janet 1.38.0-0aee7765 linux/x64/gcc - '(doc)' for help
repl:1:> <core/socket-address 0x64E73B4870B0>
repl:2:> <core/stream [fd=7]>
repl:3:> error: bad stream, expected datagram socket
  in net/send-to [src/core/net.c] on line 927
  in thunk [repl] (tail call) on line 3, column 1
```

Thank you.

PR #1621, opened 2025-08-11
Fix: Correctly flag UDP streams in net/connect #1620

tttuuu888:
This change adds the `JANET_STREAM_UDPSERVER` flag to `:datagram` streams created by `net/connect`.
Previously, these streams were missing the flag, causing functions like `net/send-to` to fail with a "bad stream" error.

Thank you.


bakpakin:
Thanks @tttuuu888 , this looks like a good improvement. I suppose although a UDP connection isn't really a "server", the `net/send-to` and `net/recv-from` would work fine in this case.

Issue #1622, opened 2025-08-13
Request for Lower-Level net socket API for flexible port handling


tttuuu888:
Hi @bakpakin  and all,

I found that it seems impossible to implement a TFTP client using the basic Janet net APIs.

The `net/connect` function binds a port to the socket upon connection, which makes it difficult to implement the TFTP protocol's behavior where the server's response needs to be received on a different, temporary port.

Therefore, to implement a protocol that requires flexible port allocation, like TFTP, a lower-level API such as `net/socket` that creates an unbound socket seems necessary.

I have attached simplified C examples of a TFTP server/client. 

[example.zip](https://github.com/user-attachments/files/21757700/example.zip)

The core of these examples is as follows:

Server:
```
// Create socket and bind to port 8080
listen_sock = socket(AF_INET, SOCK_DGRAM, 0);
bind(listen_sock, ...);

// Receive message from client
recvfrom(listen_sock, buffer, BUF_SIZE, 0, ...);

// Create a new socket without bind() to send the response
new_sock = socket(AF_INET, SOCK_DGRAM, 0);
sendto(new_sock, "world", ..., client_addr, ...);
```

Client:
```
// Create a socket (without binding)
client_sock = socket(AF_INET, SOCK_DGRAM, 0);

// Send initial request to server on port 8080 (an ephemeral port is automatically assigned)
sendto(client_sock, "hello", ..., server_addr, ...);

// Receive the server's response from a new port
recvfrom(client_sock, buffer, BUF_SIZE, 0, ..., &from_addr, ...);
// from_addr.sin_port will contain the ephemeral port number
```

The log when the example runs successfully is shown below.

Server:
```
Server listening on 127.0.0.1:8080
Server received: hello from port 54379
Server sent: world from a new socket
```

Client:
```
Client socket created (unbound)
Client sent: hello
Client received: world from port 60914
```
As I understand it, this action seems difficult to implement with `net/connect`.
I would appreciate for any feedback.

Thank you.

bakpakin:
I think this is addressed now, @tttuuu888 reopen this if anything else is needed.

PR #1623, opened 2025-08-13
Add `net/socket` for unbound socket creation for #1622

tttuuu888:
This PR add a new core function, `net/socket`, to create an unbound network socket. 
I believe this is useful for handling flexible port assignments, as seen in protocols like TFTP.

And the following are Janet server/client codes that implement the server/client examples attached to issue #1622, using the `net/socket` API added in this PR.

Server:
```
# 1. Create and bind a socket
(def listen-sock (net/listen "127.0.0.1" 8080 :datagram))
(print "Server listening on 127.0.0.1:8080")

# 2. Receive the client's message
(def buff (buffer/new 1024))
(def recv-info (net/recv-from listen-sock 1024 buff))
(print "Server received: " buff " from port " ((net/address-unpack recv-info) 1))

# 3. Send "world" message from a new socket
(def world-msg "world")
(def new-sock (net/socket :datagram))
(net/send-to new-sock recv-info world-msg)
(print "Server sent: " world-msg)

(net/close listen-sock)
(net/close new-sock)
```

Client:
```
# 1. Create a socket (without binding)
(def client-sock (net/socket :datagram))
(def server-addr (net/address "127.0.0.1" 8080))
(print "Client socket created for server at 127.0.0.1:8080")

# 2. Send "hello" message to the server
(def hello-msg "hello")
(net/send-to client-sock server-addr hello-msg)
(print "Client sent: " hello-msg)

# 3. Receive the server's response
(def buff (buffer/new 1024))
(def recv-info (net/recv-from client-sock 1024 buff))
(print "Client received: " hello-msg " from port " ((net/address-unpack recv-info) 1))
(net/close client-sock)
```

I have confirmed that the C server - Janet client and Janet server - C client combinations also work correctly.

Thank you in advance for the review.
If `net/socket` is unnecessary and there's another way to implement the behavior described in issue #1622, I'd appreciate it if you could share that with me.

Thank you.

### DIFF
--- a/src/core/net.c
+++ b/src/core/net.c
@@ -587,6 +587,56 @@ JANET_CORE_FN(cfun_net_connect,
     net_sched_connect(stream);
 }
 
+JANET_CORE_FN(cfun_net_socket,
+              "(net/socket &opt type)",
+              "Creates a new unbound socket. Type is an optional keyword, "
+              "either a :stream (usually tcp), or :datagram (usually udp). The default is :stream.") {
+    janet_arity(argc, 0, 1);
+
+    int socktype = janet_get_sockettype(argv, argc, 0);
+
+    /* Create socket */
+    JSock sfd = JSOCKDEFAULT;
+    struct addrinfo *ai = NULL;
+    struct addrinfo hints;
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = socktype;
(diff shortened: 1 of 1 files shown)


bakpakin:
The code looks good and the justification reasonable - thanks again, @tttuuu888 .

It does seem to me that we will probably want a "bind" function for these new unbound sockets, although I'm not sure of the exact use case and that would be a separate issue.

PR #1624, opened 2025-08-17
Fix `LDFLAGS` Usage in Makefile

nlordell:
This PR fixes what appears to be a typo `LDFLAGS` written with an additional `_` in the Makefile for setting the default linker flags for `libjanet`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ HOSTAR?=$(AR)
 # Symbols are (optionally) removed later, keep -g as default!
 CFLAGS?=-O2 -g
 LDFLAGS?=-rdynamic
-LIBJANET_LDFLAGS?=$(LD_FLAGS)
+LIBJANET_LDFLAGS?=$(LDFLAGS)
 RUN:=$(RUN)
 
 


sogaiu:
I've looked into this a bit and haven't found anything definitive, but FWIW there are hits for both `LD_FLAGS` and `LDFLAGS` at various locations.

Not having much luck with determining whether both are valid for some environments and/or one has been deprecated.

May keep looking.

bakpakin:
LDFLAGS is the right one, LD_FLAGS is clearly a typo. Good catch, @nlordell .

sogaiu:
On a peripheral note, there do seem to be places where `LD_FLAGS` is mentioned / used in the wild.  Possibly some or all of them are typos...Below are a few examples:

* https://gotify.net/docs/build
* https://ftp.eecs.berkeley.edu/~bh/61a-pages/Scheme/source/src/stk/Tcl/configure.in
* https://gitedu.hesge.ch/adrian.spycher/yoctos/-/blob/dcca70e8c869b50204aac8448c1234b04740ccf3/Makefile

There are far more hits for `LDFLAGS` in the gcc and llvm-project project repositories though :)

PR #1626, opened 2025-08-18
Use `CLOCKS_PER_SEC` instead of `CLOCKS_PER_SECOND`

nlordell:
The POSIX standard defines that `clock(3)` returns a `clock_t` as a number of clock ticks in `CLOCKS_PER_SEC` and not `CLOCKS_PER_SECOND`, [source](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html).


nlordell:
Note that this is superseded by #1627 (which also includes the `s/CLOCKS_PER_SECOND/CLOCKS_PER_SEC/g` fix).

bakpakin:
oops, thanks :+1:

PR #1655, opened 2025-10-11
Add option to build with the Zig build system

jpaquim:
A work in progress for now, but the basic build, bootstrap, and testing process seems to be working well.

I still want to properly review and port over the rest of the Makefile as well.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -256,6 +256,7 @@ pub fn build(b: *std.Build) void {
     const run_tests_step = b.step("run-tests", "Run Janet test suite scripts");
     for (test_scripts) |test_script| {
         const run_test_cmd = b.addRunArtifact(janet_exe);
+        run_test_cmd.step.dependOn(b.getInstallStep());
         run_test_cmd.addFileArg(b.path(test_script));
         run_test_cmd.expectExitCode(0);
         run_test_cmd.stdio = .inherit;
@@ -265,6 +266,7 @@ pub fn build(b: *std.Build) void {
     const check_examples_step = b.step("examples", "Check if example scripts compile");
     for (example_scripts) |example_script| {
         const check_example_cmd = b.addRunArtifact(janet_exe);
+        check_example_cmd.step.dependOn(b.getInstallStep());
         check_example_cmd.addArg("-k");
         check_example_cmd.addFileArg(b.path(example_script));
         check_example_cmd.stdio = .inherit;


pepe:
I'm not a C expert, unfortunately, but I would like to be able to at least try Zig more with Janet

sogaiu:
> port over the rest of the Makefile as well.

May be known but FWIW there is also [this file](https://github.com/janet-lang/janet/blob/db7f741dad064fbf392eb116f88590af02dc7139/meson.build) for building with `meson`.

albheim:
Nice, actually had a go at the same thing a month or two ago just for fun and to learn the zig build system a bit better. I think yours is more complete than mine though.

bakpakin:
I have no problem with this except that we would need to maintain this, and it really seems more like a way to test the zig compiler tool-chain rather than adding extra functionality to Janet. Also, I assume "version" in build.zig.zon should be the janet version, not 0.0.0. 

All in all, nice work but a bit out of scope, no?

McSinyx:
IMHO this would add functionality if build.zig generates janetconf.h, otherwise depending on this Zig package would be of no difference from dynamically linking to Janet built in other ways.

albheim:
You can always add it to https://github.com/AllYourCodebase/

tw4452852:
I think the most valuable benift of this is cross-compiling (including different variants of libc implementations, i.e. glibc and musl). Don't need other cross-compiling toolchains, only one zig executable is enough.

amano-kenji:
What problem does it solve?

sogaiu:
As I understand it, if the cross-compilation capabilities of Zig survive the various changes involved with getting away from using LLVM, then cross-compilation will (continue to) be aided.

May be we can avoid having to use MSVC on Windows?  In turn, possibly CI may become simpler.

It might also be that eventually, working with Zig's build system would make maintenance and extension better (of build system bits) in various ways compared to having to work with `make`, `gcc` / `clang`, Visual Studio, etc.

Part of that might be similar to benefits one might have from using `meson`.

---

On a related note, generally, the cross-compilation capability is quite handy if one wants to create executables for various platforms without having to arrange for virtual machines or access to appropriate operating systems environments (or cross-compilation toolchains as mentioned above).

It might make providing one's software to others in executable forms on a variety of platforms much less work.

---

However, Zig is not yet stable and I'm not sure if the cross-compilation capabilities will be maintained when stability is reached.

bakpakin:
Closing this as out of scope for now.

No problem with alternative build systems, provided we have a good reason, such as improved safety, performance, more targets, etc. I have vetoed similar requests for autotools-based builds in the past.

PR #1659, opened 2025-10-25
fix (try) macro hygiene

ianthehenry:
The following code worked before Janet 1.39.0, but now errors:

```janet
repl:1:> (try (print "something") ([_ _] (print "oh no")))
something
error: bad slot #0, expected fiber, got nil
  in fiber/status [src/core/fiber.c] on line 596
  in thunk [repl] (tail call) on line 1, column 1
```

#1605 introduced a nicer way to ignore the "arguments" in the catch clause, but in doing so broke this pattern.

I realize that's equivalent to `(try (print "something") ([_] (print "oh no")))`, which still works, but I was used to writing the above to make it less likely that I would forget about the fiber argument if I changed code to something that wanted to re-raise. So upgrading to Janet 1.39.0 broke some code in the wild.

### DIFF
--- a/test/suite-boot.janet
+++ b/test/suite-boot.janet
@@ -1023,4 +1023,11 @@
 (assert (deep-not= @{:key1 "value1" [@"key2"] @"value2"}
                    @{:key1 "value1" [@"key2"] @"value2"}) "deep= mutable keys")
 
+# different try overloads
+(assert (= (try (error :error) ([] :caught)) :caught))
+(assert (= (try (error :error) ([e] e)) :error))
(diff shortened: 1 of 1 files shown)


sogaiu:
Thanks for the report and the fix (and sorry we broke things!).

Would you mind adding a test to help prevent regressions?

sogaiu:
Thanks for the tests!

sogaiu:
FWIW, using https://github.com/janet-lang/janet/pull/1659/commits/0c34033b72d506789e8d197b861538ba677f9ef5 I ran tests for most of my projects as well as `spork`'s and they all turned out fine :+1:

PR #1660, opened 2025-10-25
Clarify error message for missing bundle script during bundle installation

pyrmont:
> [!NOTE]  
> This post has been edited from the original. Comments prior to [this comment](https://github.com/janet-lang/janet/pull/1660#issuecomment-3449097262) refer to the earlier version of this post.

This PR changes the error message produced during bundle installation if a bundle directory does not contain a **bundle script** (i.e. `./bundle.janet` or `./bundle/init.janet`).
 
To assist with testing, it also adds a new `assert-error-value` macro to `test/helpers.janet`.

## Background

If a user attempts to install a bundle that does not contain a bundle script, the `bundle/install` function can fail with an error message that does not clearly explain that the error is that this file is missing.

The error will occur in `get-bundle-module` and look like this:

```

error: could not find module @syspath/bundle/example:
    <syspath>/bundle/example.jimage
    <syspath>/bundle/example.janet
    <syspath>/bundle/example/init.janet
    <syspath>/bundle/example.so
  in require-1 [boot.janet] (tail call) on line 3126, column 20            
  in defer [boot.janet] on line 4200, column 7                             
  in get-bundle-module [boot.janet] on line 4192, column 5
  in edefer [boot.janet] on line 4351, column 19
  in bundle/install [boot.janet] (tail call) on line 4339, column 5
  [...]
```

## Implementation
 
This PR changes the error message in `get-bundle-module` to specifically refer to the absence of a bundle script. There are some related clean-ups of the `bundle/install` logic (notably, `initfile` is changed to `bscript` to avoid visual confusion with `infofile`).

To assist with testing, this PR also adds an `assert-error-value` macro to `test/helpers.janet` so that tests can check that the error they are expecting is raised as opposed to a different kind of error.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -4222,7 +4222,9 @@
       (put new-env *syspath* fixed-syspath)
       (with-env new-env
         (put new-env :bundle-dir (bundle-dir bundle-name)) # get the syspath right
-        (require (string "@syspath/bundle/" bundle-name)))))
+        (try
+          (require (string "@syspath/bundle/" bundle-name))
+          ([_] (error "bundle must contain bundle.janet or bundle/init.janet"))))))
 
   (defn- do-hook
     [module bundle-name hook & args]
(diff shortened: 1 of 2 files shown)


sogaiu:
I might be misunderstanding, but the original code looks like it went to some trouble to allow the possibility of an optional info file.

I wonder if there is a use for that kind of arrangement.

Guesses anyone?

sogaiu:
Perhaps a minor point but I find `infofile-*` and `initfile-*` to be pretty easy to confuse in the context of the code in question.

I wonder if it's worth doing one of:

* inserting a hyphen (e.g. `infofile-*` -> `info-file-*`)
* removing `file` (e.g. `infofile-src1` -> `info-src`)
* truncating `file` to `f` (e.g. `infofile-src1` -> `infof-src1`)

or some other helpful thing :)

pyrmont:
After further review, I realised that it's almost certainly intentional that you can call `bundle/install` for a bundle that does not include an info file. I've reverted the changes that required this file to exist and instead changed the error message thrown in `get-bundle-module` to more helpfully explain the problem. I've rewritten the OP to reflect this new version.

PR #1663, opened 2025-11-05
Keep dobytes' error string

McSinyx:
The reason for failure would be more useful than the most recently evaluated value.


bakpakin:
LGTM, thanks @McSinyx.

McSinyx:
Thanks, and what is the ETA for the next release?

bakpakin:
Not immediately, hopefully before December though. I am slowing down release cadence to as needed, there is no longer need to release monthly.

Issue #1675, opened 2025-12-01
Windows FFI function call with stack-based arguments raises fatal error if MSVC runtime stack checks (/RTCs) are enabled


tsujamin:
When MSVC and the non-standard `/RTCs` flag is used to compile Janet, [enabling runtime stack checks](https://learn.microsoft.com/en-us/cpp/build/reference/rtc-run-time-error-checks), FFI function calls with stack-based arguments fail with a fatal stack corruption error.

## Cause 
The root cause appears to be in `janet_ffi_win64()`, specifically:
- When `/RTCs` is specified in the compilation flags, `alloca()` is replaced by `_RTC_AllocaHelper()` which **prepends** an `_RTC_ALLOCA_NODE` structure to the stack buffer used to pass stack-based FFI arguments.
  <img width="987" height="355" alt="Image" src="https://github.com/user-attachments/assets/43f1ad11-8654-443b-bfda-4f494fa8e67a" />
  This causes stack-based FFI arguments to be misplaced at the time of the FFI call, with an `_RTC_ALLOCA_NODE` between the saved-reg area and stack arguments.
-  The `memmove()` to properly place stack-based arguments prior to the FFI call corrupts the `_RTC_ALLOCA_NODE` prepended to the arguments allocation, resulting in a fatal segfault. https://github.com/janet-lang/janet/blob/b9bc89a38ca44a806b60bbdb8f73d2227c268d85/src/core/ffi.c#L1383-L1391

## Example
When compiled with `/RTCs`, the following script can trigger the crash.
```
(ffi/context "bcrypt.dll")
(ffi/defbind BCryptGetProperty :u32 [hObject :u64 pszProperty :ptr pbOutput :ptr cbOutput :u32 pcbResult :ptr dwFlags :u32])

(BCryptGetProperty (int/u64 0) nil nil 0 nil 0)
```

This will either silently crash the interpreter, or raise an error such as the following.
<img width="402" height="361" alt="Image" src="https://github.com/user-attachments/assets/04db0374-f406-40e1-985c-71638f3d4f59" />

tsujamin:
Proposed fix in https://github.com/janet-lang/janet/pull/1676

tsujamin:
fix merged in #1676

PR #1676, opened 2025-12-01
Disable MSVC runtime stack checks in janet_ffi_win64

tsujamin:
The MSVC /RTCs argument instruments alloca() to prepend stack allocations with metadata used to check for memory corruption during the function prologue. This causes stack-based FFI arguments in janet_ffi_win64 to become mis-aligned, and also for the alloca()-prepended header to be corrupted leading to a fatal exception.

This change fixes the fatal error (https://github.com/janet-lang/janet/issues/1675) by disabling /RTCs during janet_ffi_win64 codegen.

### DIFF
--- a/src/core/ffi.c
+++ b/src/core/ffi.c
@@ -1344,6 +1344,15 @@ typedef double (win64_variant_f_ffif)(double, double, uint64_t, double);
 typedef double (win64_variant_f_fffi)(double, double, double, uint64_t);
 typedef double (win64_variant_f_ffff)(double, double, double, double);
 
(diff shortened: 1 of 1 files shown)


tsujamin:
Apologies for the fix being based on an older tag, I've got 1.38 vendored into another project. The bug and the fix repro correctly on current master.

bakpakin:
LGTM, thanks for the fix. I was wondering why MSVC was adding extra padding bytes to the stack when using alloca, and this illustrates one use of that extra space.

Issue #1674, opened 2025-11-30
Support splice in PEGs


sarna:
This came up on Zulip: [#help > Splicing PEG Match-time Captures?](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/Splicing.20PEG.20Match-time.20Captures.3F/with/560952159)

Copy-pasting the first post:

> Hi! Is it possible to splice back groups from match time captures (or any kind of group, really) into the output stream? Say I have something like this:
> 
> ```janet
> (peg/match ~{
>   :nn (number (some (+ :d+ (set "-."))))
>   :g3 (* :nn "," :nn "," :nn)
>   :main (* "$NMEA," (cmt :g3 ,(fn [x y z] [x y z (+ x y z)])) "," '(any 1))
> } "$NMEA,3,4,5,sork")
> ```
> Running the grammar on the example gives `@[(3 4 5 12) "sork"]`, which is _fine_, but what I really want is `@[3 4 5 12 "sork"]` (as if the `cmt` result was spliced into the result list.
> 
> Is there any way to do this? Thanks!

As of now it's not possible without diabolical hacks (which Ian demonstrated in the thread).

bakpakin:
I think this is a good idea, I also thought we could easily do this but it does seem to be glaringly missing. It is fairly simple to add a primitive instruction  that can do this with same interface as `cmt`.

sogaiu:
I've tried out `cms` and it seems to be working 👍 

---

Could we consider a generic `splice` peg special?

There is some further discussion [here](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/.E2.9C.94.20Splicing.20PEG.20Match-time.20Captures.3F/near/560950711) and [here](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/.E2.9C.94.20Splicing.20PEG.20Match-time.20Captures.3F/near/561069469) expressing some interest in that direction.  If you include yours truly, that's may be four people [1] at least.

---

I started looking into [`peg.c` around here](https://github.com/janet-lang/janet/blob/b9bc89a38ca44a806b60bbdb8f73d2227c268d85/src/core/peg.c#L624-L678) to see how this might be done.

[...]

PR #1677, opened 2025-12-02
Add `splice` peg special

sogaiu:
This PR is related to #1674 -- it may need more work [1].

Note that `cms` has already been added to address #1674 in f79e4d6249ecc25f460204f2bc509dd1ad9713f5.

The background is that although `cms` addressed the original situation (and can be applied to more situations), there was some discussion on Zulip that perhaps a `splice` peg special might be another way to address the issue.

See [this comment](https://github.com/janet-lang/janet/issues/1674#issuecomment-3600085538) and the following links for related discussions:

* [ianthehenry comments](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/.E2.9C.94.20Splicing.20PEG.20Match-time.20Captures.3F/near/560950711)
    > This seems like a very reasonable thing to add to the PEG evaluator -- splice is currently unused and it seems like it should work like this
* [oofoe comment 1](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/.E2.9C.94.20Splicing.20PEG.20Match-time.20Captures.3F/near/561069469)
    > is there any call for a generic splice operator (it appears that cms is function call only)?
* [oofoe comment 2](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/.E2.9C.94.20Splicing.20PEG.20Match-time.20Captures.3F/near/561157135)
    > Sure, but then (cms ... ,identity) then becomes an idiom, instead of an explicit affordance. Also an atomic splice operator could be used generically for anything, e.g. ... ;(group ...) ... as well as ... ;(cmt ... ,fnord) ...? (I realize that ; might interfere with other operations, please substitute whatever as appropriate.)

---

[1] ...or possibly there's something inherently problematic about it.

### DIFF
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -621,6 +621,35 @@ tail:
             return s->text_end;
         }
 
+        case RULE_SPLICE: {
+            uint32_t tag = rule[2];
+            int oldmode = s->mode;
+            CapState cs = cap_save(s);
+            s->mode = PEG_MODE_NORMAL;
+            down1(s);
+            const uint8_t *result = peg_rule(s, s->bytecode + rule[1], text);
+            up1(s);
+            s->mode = oldmode;
+            if (!result) return NULL;
+
+            int32_t num_captures = s->captures->count - cs.cap;
+            if (num_captures == 1) {
+                const Janet *elements = NULL;
+                int32_t len = 0;
+                if (janet_indexed_view(s->captures->data[s->captures->count - 1],
+                                       &elements, &len)) {
+                    // "drop" the captured single item by restoring
+                    cap_load_keept(s, cs);
+                    /* unpack and flatten capture */
+                    for (int32_t i = 0; i < len; i++) {
+                        pushcap(s, elements[i], tag);
+                    }
+                }
+            }
+
+            return result;
+        }
+
         case RULE_REPLACE:
         case RULE_MATCHSPLICE:
         case RULE_MATCHTIME: {
@@ -1169,6 +1198,9 @@ static void spec_group(Builder *b, int32_t argc, const Janet *argv) {
 static void spec_unref(Builder *b, int32_t argc, const Janet *argv) {
     spec_cap1(b, argc, argv, RULE_UNREF);
 }
+static void spec_splice(Builder *b, int32_t argc, const Janet *argv) {
+    spec_cap1(b, argc, argv, RULE_SPLICE);
+}
 
 static void spec_nth(Builder *b, int32_t argc, const Janet *argv) {
     peg_arity(b, argc, 2, 3);
@@ -1386,6 +1418,7 @@ static const SpecialPair peg_specials[] = {
     {"sequence", spec_sequence},
     {"set", spec_set},
     {"some", spec_some},
+    {"splice", spec_splice},
     {"split", spec_split},
     {"sub", spec_sub},
     {"thru", spec_thru},
@@ -1715,6 +1748,7 @@ static void *peg_unmarshal(JanetMarshalContext *ctx) {
             case RULE_GROUP:
             case RULE_CAPTURE:
             case RULE_UNREF:
+            case RULE_SPLICE:
                 /* [rule, tag] */
                 if (rule[1] >= blen) goto bad;
                 op_flags[rule[1]] |= 0x01;
--- a/src/include/janet.h
+++ b/src/include/janet.h
@@ -2204,6 +2204,7 @@ typedef enum {
     RULE_NTH,          /* [nth, rule, tag] */
     RULE_ONLY_TAGS,    /* [rule] */
     RULE_MATCHSPLICE,  /* [rule, constant, tag] */
+    RULE_SPLICE,       /* [rule, tag] */
 } JanetPegOpcode;
 
 typedef struct {
--- a/test/suite-peg.janet
+++ b/test/suite-peg.janet
@@ -845,4 +845,14 @@
       "abc"
       @[["b" "b" "b"]])
 
+(test "splice 1"
+      ~(splice (cmt (* 1 1 '1) ,|[$ $]))
+      "xyz"
+      @["z" "z"])
+
+(test "splice 2"
+      ~(group ;(cmt (* 1 1 '1) ,|[$ $]))
+      "xyz"
+      @[@["z" "z"]])
+
 (end-suite)


bakpakin:
There are a few reasons I didn't go with this to solve the original issue, although I think it is a reasonable addition.

* The first is that a rule may capture 0 or more captures, so it's unclear what happens when there are more than a single capture being spliced. A few ways to resolve this, mostly we just need to make a choice. I'm partial to the same behavior as cms, which is to simply push arguments that are not arrays/tuples as normal captures instead of discarding them. Also probably splice in _all_ captures, not just the last (or first?) -- Either than or update cms to behave the same as splice here.
* Splicing in captures will "reassign" tags. I don't know if this is really an issue, but was something I was thinking about.
* Capture match time is already the sort of generic escape hatch from PEG to code, so it made sense to just extend that to work with corner cases.

sogaiu:
Thanks for the comments!

Will be mulling them over.

I'll mark this PR as draft for the moment.

pepe:
> * Capture match time is already the sort of generic escape hatch from PEG to code, so it made sense to just extend that to work with corner cases.

That is the way I was always using to solve these problems.

sogaiu:
Just want to say upfront that I have no attachment to this PR.  I value that `janet` works well and would prefer not to add bits that jeopordize things.  Still, I'm interested in seeing whether the `splice` idea can be done well enough :)

---

Did some work trying to address the first bullet point:

> to simply push arguments that are not arrays/tuples as normal captures instead of discarding them. Also probably splice in all captures, not just the last (or first?)

To avoid overwriting existing captures, captures were copied elsewhere.  Not knowing how large of an array to allocate upfront so a minimum size (number of captures) was chosen and then relied on `janet_array_push` to grow accordingly.

I suppose twice the size of the number of captures might be a better choice or perhaps making an initial pass through to count things up is preferrable...

Didn't manage to think of a way to avoid the copying...may be there is a good way?

Also, replaced the two tests from before with more tests that I think cover the ground more.  A couple of them may help with checking for overwriting.

May be there's a nicer way, but FWIW:

```diff
diff --git a/src/core/peg.c b/src/core/peg.c
index e204e90a..cf7c67e0 100644
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -632,21 +632,27 @@ tail:
             s->mode = oldmode;
             if (!result) return NULL;
 
-            int32_t num_captures = s->captures->count - cs.cap;
-            if (num_captures == 1) {
+            /* save captures to avoid overwriting via pushcap below */
+            uint32_t num_captures = s->captures->count - cs.cap;
+            JanetArray *splice_caps = janet_array(num_captures); /* min size */
+            for (uint32_t i = 0; i < num_captures; i++) {
                 const Janet *elements = NULL;
                 int32_t len = 0;
-                if (janet_indexed_view(s->captures->data[s->captures->count - 1],
-                                       &elements, &len)) {
-                    // "drop" the captured single item by restoring
-                    cap_load_keept(s, cs);
-                    /* unpack and flatten capture */
-                    for (int32_t i = 0; i < len; i++) {
-                        pushcap(s, elements[i], tag);
+                if (janet_indexed_view(s->captures->data[i], &elements, &len)) {
+                    for (int32_t j = 0; j < len; j++) {
+                        janet_array_push(splice_caps, elements[j]);
                     }
+                } else {
+                    janet_array_push(splice_caps, s->captures->data[i]);
                 }
             }
 
+            /* restore, then push saved captures */
+            cap_load_keept(s, cs);
+            for (int32_t k = 0; k < splice_caps->count; k++) {
+                pushcap(s, splice_caps->data[k], tag);
+            }
+
             return result;
         }
 
diff --git a/test/suite-peg.janet b/test/suite-peg.janet
index e4572f2e..6cba35f8 100644
--- a/test/suite-peg.janet
+++ b/test/suite-peg.janet
@@ -845,14 +845,67 @@
       "abc"
       @[["b" "b" "b"]])
 
-(test "splice 1"
-      ~(splice (cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @["z" "z"])
-
-(test "splice 2"
-      ~(group ;(cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @[@["z" "z"]])
+# splice
+(test "basic splice"
+      ~(splice (group '1))
+      "abc"
+      @["a"])
+(test "compare with no splice for captured indexed type"
+      ~(group '1)
+      "abc"
+      @[@["a"]])
+(test "retain non-indexed type capture"
+      ~(splice '1)
+      "abc"
+      @["a"])
+(test "retain multiple non-indexed type captures"
+      ~(splice (* '1 '1))
+      "abc"
+      @["a" "b"])
+(test "splice indexed type built from multiple captures"
+      ~(splice (group (* '1 '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures"
+      ~(* (group '1) (group '1))
+      "abc"
+      @[@["a"] @["b"]])
+(test "splicing of multiple indexed type captures"
+      ~(splice (* (group '1) (group '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures - varying lengths"
+      ~(* (group '1) (group (* '1 '1)))
+      "abc"
+      @[@["a"] @["b" "c"]])
+(test "splice multiple indexed type captures - varying lengths"
+      ~(splice (* (group '1) (group (* '1 '1))))
+      "abc"
+      @["a" "b" "c"])
+(test "multiple indexed type captures - first having more elements"
+      ~(* (group (* '1 '1)) (group '1))
+      "abc"
+      @[@["a" "b"] @["c"]])
+# this tests that no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 1"
+      ~(splice (* (group (* '1 '1)) (group '1)))
+      "abc"
+      @["a" "b" "c"])
+# this also tests no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 2"
+      ~(splice (* (group (* '1 '1))
+                  (group '1)
+                  (group '1)))
+      "abcd"
+      @["a" "b" "c" "d"])
+(test "issue #1674"
+      ~{:nn (number (some (+ :d+ (set "-."))))
+        :g3 (* :nn "," :nn "," :nn)
+        :main (* "$NMEA,"
+                 ;(cmt :g3 ,(fn [x y z] [x y z (+ x y z)]))
+                 ","
+                 '(any 1))}
+      "$NMEA,3,4,5,sork"
+      @[3 4 5 12 "sork"])
 
 (end-suite)
```

sogaiu:
I have an idea about improving the approach taken in the diff in the [comment above](https://github.com/janet-lang/janet/pull/1677#issuecomment-3605969585):

* Make a single pass through initially to determine the number of elements needed for splicing.
* Start from the end of the currently captured things, splice "backwards".

I think this should reduce the amount of copying needed.

Will try to get this working.

sogaiu:
The following diff is an attempt to carry out the idea in the [comment above](https://github.com/janet-lang/janet/pull/1677#issuecomment-3609212251):

> * Make a single pass through initially to determine the number of elements needed for splicing.
> * Start from the end of the currently captured things, splice "backwards".

It doesn't use `pushcap` and directly manipulates `s->captures`...may be that's not such a great thing from a comprehension and maintenance perspective.

Also, no tagging is done currently as `pushcap` is not called and I haven't put anything else into place.

I've been trying to think of the point of tagging in the context of splicing and haven't yet arrived at a sensible perspective.  Since splicing can lead to multiple values showing up on the capture stack, it seems odd to tag all of those with the same tag as it appears that means one can only access the item that was "tagged last".

May be that's what this remark:

> Splicing in captures will "reassign" tags. I don't know if this is really an issue, but was something I was thinking about.

was getting at?

Would it make sense to just punt on the tagging?  

I put a second diff below the first one here that is for the signature `(splice patt)` instead of `(splite patt ?tag)`.

`(splite patt ?tag)`:

```diff
diff --git a/src/core/peg.c b/src/core/peg.c
index e204e90a..869e8b8c 100644
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -622,6 +622,7 @@ tail:
         }
 
         case RULE_SPLICE: {
+            /* currently ignoring tag */
             uint32_t tag = rule[2];
             int oldmode = s->mode;
             CapState cs = cap_save(s);
@@ -632,18 +633,36 @@ tail:
             s->mode = oldmode;
             if (!result) return NULL;
 
-            int32_t num_captures = s->captures->count - cs.cap;
-            if (num_captures == 1) {
+            uint32_t num_captures = s->captures->count - cs.cap;
+
+            /* count number of elements to result from splice */
+            int32_t num_elts = 0;
+            for (uint32_t i = 0; i < num_captures; i++) {
                 const Janet *elements = NULL;
                 int32_t len = 0;
-                if (janet_indexed_view(s->captures->data[s->captures->count - 1],
-                                       &elements, &len)) {
-                    // "drop" the captured single item by restoring
-                    cap_load_keept(s, cs);
-                    /* unpack and flatten capture */
-                    for (int32_t i = 0; i < len; i++) {
-                        pushcap(s, elements[i], tag);
+                if (janet_indexed_view(s->captures->data[i], &elements, &len)) {
+                    num_elts += len;
+                } else {
+                    num_elts++;
+                }
+            }
+
+            /* ensure there is enough space for splicing */
+            janet_array_setcount(s->captures, num_elts);
+
+            /* splice backwards from the end to avoid overwriting */
+            uint32_t i = num_elts - 1;
+            for (int32_t j = num_captures - 1; j >= 0; j--) {
+                const Janet *elements = NULL;
+                int32_t len = 0;
+                if (janet_indexed_view(s->captures->data[j], &elements, &len)) {
+                    for (int32_t k = len - 1; k >= 0; k--) {
+                        s->captures->data[i] = elements[k];
+                        i--;
                     }
+                } else {
+                    s->captures->data[i] = s->captures->data[j];
+                    i--;
                 }
             }
 
diff --git a/test/suite-peg.janet b/test/suite-peg.janet
index e4572f2e..6cba35f8 100644
--- a/test/suite-peg.janet
+++ b/test/suite-peg.janet
@@ -845,14 +845,67 @@
       "abc"
       @[["b" "b" "b"]])
 
-(test "splice 1"
-      ~(splice (cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @["z" "z"])
-
-(test "splice 2"
-      ~(group ;(cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @[@["z" "z"]])
+# splice
+(test "basic splice"
+      ~(splice (group '1))
+      "abc"
+      @["a"])
+(test "compare with no splice for captured indexed type"
+      ~(group '1)
+      "abc"
+      @[@["a"]])
+(test "retain non-indexed type capture"
+      ~(splice '1)
+      "abc"
+      @["a"])
+(test "retain multiple non-indexed type captures"
+      ~(splice (* '1 '1))
+      "abc"
+      @["a" "b"])
+(test "splice indexed type built from multiple captures"
+      ~(splice (group (* '1 '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures"
+      ~(* (group '1) (group '1))
+      "abc"
+      @[@["a"] @["b"]])
+(test "splicing of multiple indexed type captures"
+      ~(splice (* (group '1) (group '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures - varying lengths"
+      ~(* (group '1) (group (* '1 '1)))
+      "abc"
+      @[@["a"] @["b" "c"]])
+(test "splice multiple indexed type captures - varying lengths"
+      ~(splice (* (group '1) (group (* '1 '1))))
+      "abc"
+      @["a" "b" "c"])
+(test "multiple indexed type captures - first having more elements"
+      ~(* (group (* '1 '1)) (group '1))
+      "abc"
+      @[@["a" "b"] @["c"]])
+# this tests that no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 1"
+      ~(splice (* (group (* '1 '1)) (group '1)))
+      "abc"
+      @["a" "b" "c"])
+# this also tests no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 2"
+      ~(splice (* (group (* '1 '1))
+                  (group '1)
+                  (group '1)))
+      "abcd"
+      @["a" "b" "c" "d"])
+(test "issue #1674"
+      ~{:nn (number (some (+ :d+ (set "-."))))
+        :g3 (* :nn "," :nn "," :nn)
+        :main (* "$NMEA,"
+                 ;(cmt :g3 ,(fn [x y z] [x y z (+ x y z)]))
+                 ","
+                 '(any 1))}
+      "$NMEA,3,4,5,sork"
+      @[3 4 5 12 "sork"])
 
 (end-suite)
```

`(splice patt)`:

```diff
diff --git a/src/core/peg.c b/src/core/peg.c
index e204e90a..971fa9b8 100644
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -622,7 +622,6 @@ tail:
         }
 
         case RULE_SPLICE: {
-            uint32_t tag = rule[2];
             int oldmode = s->mode;
             CapState cs = cap_save(s);
             s->mode = PEG_MODE_NORMAL;
@@ -632,18 +631,36 @@ tail:
             s->mode = oldmode;
             if (!result) return NULL;
 
-            int32_t num_captures = s->captures->count - cs.cap;
-            if (num_captures == 1) {
+            uint32_t num_captures = s->captures->count - cs.cap;
+
+            /* count number of elements to result from splice */
+            int32_t num_elts = 0;
+            for (uint32_t i = 0; i < num_captures; i++) {
+                const Janet *elements = NULL;
+                int32_t len = 0;
+                if (janet_indexed_view(s->captures->data[i], &elements, &len)) {
+                    num_elts += len;
+                } else {
+                    num_elts++;
+                }
+            }
+
+            /* ensure there is enough space for splicing */
+            janet_array_setcount(s->captures, num_elts);
+
+            /* splice backwards from the end to avoid overwriting */
+            uint32_t i = num_elts - 1;
+            for (int32_t j = num_captures - 1; j >= 0; j--) {
                 const Janet *elements = NULL;
                 int32_t len = 0;
-                if (janet_indexed_view(s->captures->data[s->captures->count - 1],
-                                       &elements, &len)) {
-                    // "drop" the captured single item by restoring
-                    cap_load_keept(s, cs);
-                    /* unpack and flatten capture */
-                    for (int32_t i = 0; i < len; i++) {
-                        pushcap(s, elements[i], tag);
+                if (janet_indexed_view(s->captures->data[j], &elements, &len)) {
+                    for (int32_t k = len - 1; k >= 0; k--) {
+                        s->captures->data[i] = elements[k];
+                        i--;
                     }
+                } else {
+                    s->captures->data[i] = s->captures->data[j];
+                    i--;
                 }
             }
 
@@ -1176,6 +1193,9 @@ static void spec_drop(Builder *b, int32_t argc, const Janet *argv) {
 static void spec_only_tags(Builder *b, int32_t argc, const Janet *argv) {
     spec_onerule(b, argc, argv, RULE_ONLY_TAGS);
 }
+static void spec_splice(Builder *b, int32_t argc, const Janet *argv) {
+    spec_onerule(b, argc, argv, RULE_SPLICE);
+}
 
 /* Rule of the form [rule, tag] */
 static void spec_cap1(Builder *b, int32_t argc, const Janet *argv, uint32_t op) {
@@ -1198,9 +1218,6 @@ static void spec_group(Builder *b, int32_t argc, const Janet *argv) {
 static void spec_unref(Builder *b, int32_t argc, const Janet *argv) {
     spec_cap1(b, argc, argv, RULE_UNREF);
 }
-static void spec_splice(Builder *b, int32_t argc, const Janet *argv) {
-    spec_cap1(b, argc, argv, RULE_SPLICE);
-}
 
 static void spec_nth(Builder *b, int32_t argc, const Janet *argv) {
     peg_arity(b, argc, 2, 3);
@@ -1748,7 +1765,6 @@ static void *peg_unmarshal(JanetMarshalContext *ctx) {
             case RULE_GROUP:
             case RULE_CAPTURE:
             case RULE_UNREF:
-            case RULE_SPLICE:
                 /* [rule, tag] */
                 if (rule[1] >= blen) goto bad;
                 op_flags[rule[1]] |= 0x01;
@@ -1779,6 +1795,7 @@ static void *peg_unmarshal(JanetMarshalContext *ctx) {
             case RULE_NOT:
             case RULE_TO:
             case RULE_THRU:
+            case RULE_SPLICE:
                 /* [rule] */
                 if (rule[1] >= blen) goto bad;
                 op_flags[rule[1]] |= 0x01;
diff --git a/src/include/janet.h b/src/include/janet.h
index 228b2b07..5a15f852 100644
--- a/src/include/janet.h
+++ b/src/include/janet.h
@@ -2204,7 +2204,7 @@ typedef enum {
     RULE_NTH,          /* [nth, rule, tag] */
     RULE_ONLY_TAGS,    /* [rule] */
     RULE_MATCHSPLICE,  /* [rule, constant, tag] */
-    RULE_SPLICE,       /* [rule, tag] */
+    RULE_SPLICE,       /* [rule] */
 } JanetPegOpcode;
 
 typedef struct {
diff --git a/test/suite-peg.janet b/test/suite-peg.janet
index e4572f2e..6cba35f8 100644
--- a/test/suite-peg.janet
+++ b/test/suite-peg.janet
@@ -845,14 +845,67 @@
       "abc"
       @[["b" "b" "b"]])
 
-(test "splice 1"
-      ~(splice (cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @["z" "z"])
-
-(test "splice 2"
-      ~(group ;(cmt (* 1 1 '1) ,|[$ $]))
-      "xyz"
-      @[@["z" "z"]])
+# splice
+(test "basic splice"
+      ~(splice (group '1))
+      "abc"
+      @["a"])
+(test "compare with no splice for captured indexed type"
+      ~(group '1)
+      "abc"
+      @[@["a"]])
+(test "retain non-indexed type capture"
+      ~(splice '1)
+      "abc"
+      @["a"])
+(test "retain multiple non-indexed type captures"
+      ~(splice (* '1 '1))
+      "abc"
+      @["a" "b"])
+(test "splice indexed type built from multiple captures"
+      ~(splice (group (* '1 '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures"
+      ~(* (group '1) (group '1))
+      "abc"
+      @[@["a"] @["b"]])
+(test "splicing of multiple indexed type captures"
+      ~(splice (* (group '1) (group '1)))
+      "abc"
+      @["a" "b"])
+(test "multiple indexed type captures - varying lengths"
+      ~(* (group '1) (group (* '1 '1)))
+      "abc"
+      @[@["a"] @["b" "c"]])
+(test "splice multiple indexed type captures - varying lengths"
+      ~(splice (* (group '1) (group (* '1 '1))))
+      "abc"
+      @["a" "b" "c"])
+(test "multiple indexed type captures - first having more elements"
+      ~(* (group (* '1 '1)) (group '1))
+      "abc"
+      @[@["a" "b"] @["c"]])
+# this tests that no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 1"
+      ~(splice (* (group (* '1 '1)) (group '1)))
+      "abc"
+      @["a" "b" "c"])
+# this also tests no overwriting occurs
+(test "splice multiple indexed type captures - first having more elements 2"
+      ~(splice (* (group (* '1 '1))
+                  (group '1)
+                  (group '1)))
+      "abcd"
+      @["a" "b" "c" "d"])
+(test "issue #1674"
+      ~{:nn (number (some (+ :d+ (set "-."))))
+        :g3 (* :nn "," :nn "," :nn)
+        :main (* "$NMEA,"
+                 ;(cmt :g3 ,(fn [x y z] [x y z (+ x y z)]))
+                 ","
+                 '(any 1))}
+      "$NMEA,3,4,5,sork"
+      @[3 4 5 12 "sork"])
 
 (end-suite)
```

sogaiu:
I'm going to close this as I don't think `splice` fits the current peg system well for a few reasons:

1. Getting the already captured content spliced into approximately the same part of the capture stack is awkward to do efficiently (the best I came up with bypasses `pushcap` which seems bad for maintenance and comprehension).

2. I don't think splicing and the existing idea of providing a single tag per "captured content" fit very well.

Slightly more detail is in [this comment](https://github.com/janet-lang/janet/pull/1677#issuecomment-3610120507).

In summary, rather than force something that seems on the surface like it could be quite nice but isn't, I think that doing things through `cms` makes more sense as @pepe pointed out in [this comment](https://github.com/janet-lang/janet/pull/1677#issuecomment-3605743544).

That seems less misleading and an overall better choice to me (at least at this time).

PR #1678, opened 2025-12-03
shell: Prevent buggy moving zero column behavior

tw4452852:
According to https://unix.stackexchange.com/a/559331, moving zero column is implemented inconsistently between different terminal emulators.

In order to fix this inconsistency,
we avoid using this escape sequence entirely.

BTW, this fix is ported from https://github.com/janet-lang/spork/pull/262

Change-Id: I73d4252f8472c769f0cf98b6bbdf2b3d6a6a6964

### DIFF
--- a/src/mainclient/shell.c
+++ b/src/mainclient/shell.c
@@ -379,10 +379,12 @@ static void refresh(void) {
     janet_buffer_push_cstring(&b, gbl_prompt);
     janet_buffer_push_bytes(&b, (uint8_t *) _buf, _len);
     /* Erase to right */
-    janet_buffer_push_cstring(&b, "\x1b[0K");
+    janet_buffer_push_cstring(&b, "\x1b[0K\r");
     /* Move cursor to original position. */
-    snprintf(seq, 64, "\r\x1b[%dC", (int)(_pos + gbl_plen));
-    janet_buffer_push_cstring(&b, seq);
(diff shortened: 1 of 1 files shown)


sogaiu:
I thought the changes were working, but when I tried a few times, I noticed some odd behavior (not sure if it's the changes from this PR).

The way I verified was to set the dynamic variable `*repl-prompt*` to `(fn [_] "")`, enter an expression, and then type ctrl-a.

Before applying [878cc65](https://github.com/janet-lang/janet/pull/1678/commits/878cc65cd4ec602b115f9ba08750843f125460a9) (note that the cursor is not all the way to the left):

before-pr-column-zero

After applying [878cc65](https://github.com/janet-lang/janet/pull/1678/commits/878cc65cd4ec602b115f9ba08750843f125460a9):

after-pr-column-zero

Not sure what is going on here...

FWIW:

```
$ xfce4-terminal --version
xfce4-terminal 1.1.3 (Xfce 4.18)
```

tw4452852:
I know the reason. Will revise the patch later.

tw4452852:
@sogaiu I've fixed the bug in the previous version and verified on my side. You could give it a double confirm.

sogaiu:
@tw4452852 Tried it a few times and it seems to be working better :+1: 

6e883b89

PR #1679, opened 2025-12-10
Take 2: 9front port, using 9front's Native Porting/Posix Environment

pixelherodev:
There's a few things that need to be worked on for sure, but this is an initial working 9front port. Tested on amd64, and very similar to my second (unsubmitted) attempt on arm64 a few months ago, cleaned up from my first (submitted) attempt a year or two ago :)

### DIFF
--- a/src/core/pp.c
+++ b/src/core/pp.c
@@ -794,9 +794,7 @@ static const char *get_fmt_mapping(char c) {
         if (format_mappings[i].c == c)
             return format_mappings[i].mapping;
     }
-#ifndef JANET_PLAN9
     janet_assert(0, "bad format mapping");
-#endif
 }
 
 static const char *scanformat(
@@ -1025,10 +1023,19 @@ void janet_buffer_format(
             char form[MAX_FORMAT], item[MAX_ITEM];
             char width[3], precision[3];
             int nb = 0; /* number of bytes in added item */
+#ifdef JANET_PLAN9
+			if(*strfrmt == 'r'){
+				rerrstr(item, MAX_ITEM);
+				nb = strlen(item);
+			} else
+#endif
             if (++arg >= argc)
                 janet_panic("not enough values for format");
             strfrmt = scanformat(strfrmt, form, width, precision);
             switch (*strfrmt++) {
+#ifdef JANET_PLAN9
+				case 'r': break;
+#endif
                 case 'c': {
                     nb = snprintf(item, MAX_ITEM, form, (int)
                                   janet_getinteger(argv, arg));


pixelherodev:
This includes 9front's new atomics support, and other than that is mostly just disabling some things that don't work properly on 9front (threading, ev, etc). Some of the floating point tests are disabled as they give slightly different results on Plan 9 than the Janet implementation.

pixelherodev:
Tested using a light patch to NPE, at http://shithub.us/pix/npe/HEAD/info.html .

Instructions for building on plan9, that should work fully self-contained:

```
git/clone gits://shithub.us/pix/npe
cd npe
mk install
cd ..
git/clone https://github.com/pixelherodev/janet
cd janet
mkdir build/^(boot c core)
mk all
mk install
```

That should install Janet for the current architecture.

pixelherodev:
(For one last bit of context, previous attempt was Janet 1.26. This is the first time I've had a newer binary than that working, let alone one that's _almost_ good enough to upstream :D)

pixelherodev:
I think it's worth cleaning this up and merging as is [with some bugfixes], though I'm looking at what I need in order to remove JANET_REDUCED_OS...

bakpakin:
Overall, looks pretty good. Generally, fewer ifdefs the better for uncommon platforms, but I can't test on 9front.

bakpakin:
Removing the "TODO" comment will let it pass CI

bakpakin:
> This includes 9front's new atomics support, and other than that is mostly just disabling some things that don't work properly on 9front  (threading, ev, etc)

That's a shame, but makes sense. I think ev/ should be implementable with `poll()` or similar, but I don't know enough about the OS to know if that is the best way to do it. As for threading, is it feasible to do anything like that on 9Front? I am curious if that would even be feasible.

That said, @pixelherodev thanks for all the work here, a more minimal port is fine for many platforms. Happy to merge this as is

jackrusher:
Just poking my nose in to say I love to see this :)

pixelherodev:
> That's a shame, but makes sense.

Almost all of the features _can_ be implemented, but - well, Plan 9 was basically explicitly an attempt to do better than UNIX, and so it rejects a _lot_ of things that are viewed as _wrong_ in UNIX. So, implementing them is more like targeting NT than it is like targeting, say, a BSD. Plan9 almost makes a point of doing things differently and rejecting compatibility with what are seen as bad designs.

> Generally, fewer ifdefs the better for uncommon platforms, but I can't test on 9front.

I can add a plan9.c file or some such? TBH, it'd make some sense to me to have platform-specific code be isolated that way rather than by ifdefs in a single larger file, but I didn't want to try and change the structure of the codebase while doing a port :)

> I think ev/ should be implementable with poll() or similar

There is no `poll` :sweat:  It should be implementable anyways, but I'll need to look more closely at how the ev abstractions work in Janet in order to implement them properly.

> Removing the "TODO" comment will let it pass CI

Will do :)

> As for threading, is it feasible to do anything like that on 9Front?

Heh. Plan9's _native threading_ is basically Go. We have coroutines, threads, channels, etc, all in our libc, as well as functions for offloading I/O to dedicated worker procs.

The only reason it's not _here_ is because I need to figure out how to meld it to the Janet APIs :)

> a more minimal port is fine for many platforms

For now, sure ;)

One reason that this works so well for me on Plan9, vs on other platforms, is because most of the functionality I'd care about can be implemented in Janet. e.g. if I want to draw to the screen, I can just open files in /dev/draw and write to them. I've implemented directory scanning using slurp and some bit manipulation on the returned data :)

> Happy to merge this as is

:D

pixelherodev:
Okay, should be good now!

I also cleaned up the mkfile a bit so that making the directories isn't needed. Now, anyone on 9front who already has NPE installed can just run
```sh
git/clone https://github.com/janet-lang/janet
cd janet
mk install
```

pixelherodev:
> I can add a plan9.c file or some such? TBH, it'd make some sense to me to have platform-specific code be isolated that way rather than by ifdefs in a single larger file, but I didn't want to try and change the structure of the codebase while doing a port :)

~~Would you be open to a refactor that splits os.c into os.c (shared across all systems), os_windows.c, os_posix.c, os_plan9.c? It would get rid of the ifdef soup, but has its own tradeoffs, and I wasn't about to try doing something like that without talking about it first :)~~

Disregard, code would likely be a horrid mess that way. There's too much overlap even within functions.

pixelherodev:
I'm currently working on removing the REDUCED_OS config, at least. Had to add a flag for disabling locales, and also disable mode/stat on plan9, but other than that _most_ things JustWorkTM, with minor changes (e.g. non-reentrant gmtime, haven't tried working around that yetr).

pixelherodev:
> I'm currently working on removing the REDUCED_OS config, at least. Had to add a flag for disabling locales, and also disable mode/stat on plan9, but other than that _most_ things JustWorkTM, with minor changes (e.g. non-reentrant gmtime, haven't tried working around that yetr).

So, JANET_REDUCED_OS is removed :3 stat and modes all work now; just extended our POSIX emulation a bit to support those, and a few other things, rather than fill Janet with lines disabling them.

JANET_NO_PROCESSES is gone as well, and tested pretty well; I've been using os/posix-fork to implement a multithreaded chat client since there's no event loop yet.

bakpakin:
Alright, looks pretty good and ready to merge to me.

PR #1680, opened 2025-12-12
9front: expanded OS module support

pixelherodev:
This depends on changes on the 9front side, it's not ready yet, but it at least gets JANET_REDUCED_OS removed :)

Threading next, I think.


pixelherodev:
Just merging this into the other one. I've been adding more as I go anyways, I'll just do as much as I can there until it gets reviewed again :joy:

PR #1681, opened 2025-12-12
Apply :flycheck on def as well

tw4452852:
This is actually motivated by the investigation on https://github.com/janet-lang/spork/pull/264.

With this change, we could also allow user to do something like this:

```
(def v :flycheck (maybe-side-effect-func ...))
```

If they believe `maybe-side-effect-func` won't cause any harm during flycheck.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -4110,8 +4110,17 @@
     true))
 
 (defn- is-safe-def [thunk source env where]
-  (if (no-side-effects (last source))
-    (thunk)))
+  (if-let [ve (get env (source 1))
+           fc (get ve :flycheck)]
+    (cond
+      # Sometimes safe form
+      (function? fc)
+      (fc thunk source env where)
+      # Always safe form
+      fc
+      (thunk))
+    (if (no-side-effects (last source))
+      (thunk))))
 
 (defn- flycheck-importer
   [thunk source env where]


sogaiu:
I'm still familiarizing myself with janet's `flycheck` bits so likely others are better at reviewing this, but it seems like a nice idea.

Any hints about what `ve` is supposed to stand for?  May be "value" or "variable" entry?

I guess that `fc` is short for "flycheck".

---

On a side note, I think the [CI thing](https://github.com/janet-lang/janet/actions/runs/20152361677/job/57847760774?pr=1681) may be similar to what's described here: https://github.com/janet-lang/spork/pull/258.

I've made a separate PR for that [here](https://github.com/janet-lang/janet/pull/1682).

tw4452852:
> Any hints about what ve is supposed to stand for? May be "value" or "variable" entry?

I originally regard it as `variable's environment` , maybe `variable's metadata` is more appropriate? I'm so poor at naming...

> I guess that fc is short for "flycheck".

Yep.

sogaiu:
Thanks for the clarification.

FWIW, it seems like there is precedent for `entry` in `boot.janet`, e.g.:

* [in `macex1`](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/src/boot/boot.janet#L2229)
* [in `varfn`](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/src/boot/boot.janet#L2403-L2409)
* [in `module/value`](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/src/boot/boot.janet#L3001-L3005)
* [in `flycheck-evaluator`](https://github.com/janet-lang/janet/blob/efe66f207f94e194c25c76a312e16dcdf160c50b/src/boot/boot.janet#L4065-L4066)

bakpakin:
Looks pretty good to me, good for one off fixes to make things pass flychecking.

PR #1686, opened 2025-12-15
Use double-NULL byte for empty process environments on Windows

pyrmont:
While working on [Churlish](https://github.com/pyrmont/churlish) (a wrapper around the `curl` CLI utility), I noticed intermittent errors while using `os/spawn` on Windows. After further investigation, I believe the issue is that I was calling `os/spawn` with an effectively empty[^1] environment and that this was causing `CreateProcess` to occasionally fail due to a bug in `os_execute_env`.

The issue is that an empty environment block needs to be double-NULL terminated as described in more detail in [Microsoft's documentation](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa):

> Note that an ANSI environment block is terminated by two zero bytes: one for the last string, one more to terminate the block. A Unicode environment block is terminated by four zero bytes: two for the last string, two more to terminate the block.

I suspect that the error was intermittent because sometimes the memory being accessed happened to have a second NULL byte after the first. This was not guaranteed to be the case.

This commit adds an extra NULL byte if the environment is empty. It also adds a test.

[^1]: I was calling `os/spawn` with a struct that mapped `:in`, `:out` and `:err` to pipes. I mistakenly thought that this meant I needed to use `:pe` as the second argument to `os/spawn`. Because the struct contained no other values, it was empty for the purposes of `os_execute_env`.

### DIFF
--- a/test/suite-os.janet
+++ b/test/suite-os.janet
@@ -145,7 +145,10 @@
 
 # os/execute with empty environment
 # pr #1686
-(assert (= (if (= :mingw (os/which)) -1073741515 0) # MinGW can't find DLLs
+# native MinGW can't find system DLLs without PATH and so fails
+(assert (= (if (and (= :mingw (os/which))
+                    (nil? (os/stat "C:\\windows\\system32\\wineboot.exe")))
+             -1073741515 0)
            (os/execute [;run janet "-e" "(+ 1 2 3)"] :pe {}))
         "os/execute with minimal env")
 


pyrmont:
For future reference, binaries that are compiled with MinGW can fail on Windows (not WINE) when launched with `os/execute` (or `os/spawn`) because the absence of `PATH` causes system DLLs not to load. As best as I can tell, this is a MinGW-specific issue and results in error 0xC0000135 (exit code `-1073741515`). The test case checks if running MinGW on Windows and, only if that is the case, permits that exit code.

This is obviously not a great solution but I think it's the most reasonable approach. All the alternatives I could think of (e.g. silently adding `PATH` to the 'empty' environment when running in MinGW-compiled Janet, always panicking even though there might not be a problem) seemed worse.

bakpakin:
Looks good, that is a subtle issue. A missing PATH environment variable is a recipe for all sorts of issues, so I wouldn't worry too much about that failure mode

Issue #1383, opened 2024-01-31
`empty?` should be documented more clearly.


amano-kenji:
```
repl:1:> (doc empty?)


    function
    boot.janet on line 117, column 1

    (empty? xs)

    Check if xs is empty.
```
What? Define empty. I thought `nil` was empty.

kenaryn:
`nil` is derived from latin _nihil_ which approximately stands for "no thing". There is no thing (pun intended) like emptyness. Even the cosmic void is filled with reversed Kasimir force (formerly known as dark matter then more recently as dark energy).

I think the Janet doc means "**Check if xs has a null valence** ( from latin ne ullus, "not one")  if I were to use array theory terminology. I know Janet is not based in Array paradigm (unlike a language like Nial) so perhaps **Check if xs has a zero dimensionality** could be adopted in stead.

amano-kenji:
Define `xs`.

oneness:
`xs` in janet (in version 1.34 that I verified) means `iterables`. While I agree docs could be improved, I really like Janets error reporting, which is self explanatory.

sogaiu:
It is true that it's not spelled out what `xs` represents.

For reference, this seems to also be the case in the docstrings for the following as well: `%`, `%=`, `*`, `*=`, `+`, `-`, `/`, `<`, `<=`, `=`, `>=`, `array/insert`, `array/push`, `band`, `bor`, `buffer`, `buffer/push`, `buffer/push-at`, `buffer/push-byte`, `buffer/push-string`, `buffer/push-word`, `bxor`, `compare<`, `compare<=`, `compare=`, `compare>`, `compare>=`, `distinct`, `div`, `eprin`, `eprinf`, `eprint`, `eprintf`, `geomean`, `last`, `mean`, `mod`, `not=`, `prin`, `prinf`, `print`, `printf`, `product`, `string`, `sum`, `symbol`, `tuple/brackets`, `xprin`, `xprinf`, `xprint`, and `xprintf`.

sogaiu:
I've gone over the list above and based on each of the corresponding docstrings, I don't think it's unclear what `xs` means for the following:

* `array/insert`
* `array/push`
* `band`
* `bor`
* `buffer`
* `buffer/push`
* `buffer/push-at`
* `buffer/push-byte`
* `buffer/push-string`
* `buffer/push-word`
* `bxor`
* `compare<`
* `compare<=`
* `compare=`
* `compare>`
* `compare>=`
* `div`
[...]

PR #1696, opened 2026-01-19
Document empty? more clearly.

amano-kenji:
Now, (doc empty?) says empty? checks whether iterable is empty.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -114,7 +114,7 @@
 (defn true? "Check if x is true." [x] (= x true))
 (defn false? "Check if x is false." [x] (= x false))
 (defn nil? "Check if x is nil." [x] (= x nil))
-(defn empty? "Check if xs is empty." [xs] (= nil (next xs nil)))
+(defn empty? "Check if an iterable, `iter`, is empty." [iter] (= nil (next iter nil)))
 
 # For macros, we define an imcomplete odd? function that will be overriden.
 (defn odd? [x] (= 1 (mod x 2)))


amano-kenji:
It closes https://github.com/janet-lang/janet/issues/1383

sogaiu:
The proposed change at time of this writing is to substitute the string `iterable` for `xs` in:

```janet
(defn empty? "Check if xs is empty." [xs] (= nil (next xs nil)))
```

to arrive at:

```janet
(defn empty? "Check if iterable is empty." [iterable] (= nil (next iterable nil)))
```

Although I see that you are using `iterable` in [your `spork/generators` PR](https://github.com/janet-lang/spork/pull/272), the word `iterable` does not appear elsewhere in `boot.janet`.

It also does not appear to have a description on the website, whereas for example, `indexed`, `dictionary`, and `bytes` do [here](https://janet-lang.org/docs/data_structures/index.html).

For reference, the lone hit I see for `iterable` within the main content of the website is [here](https://janet-lang.org/docs/loop.html):

> The `next` function is the primary way to iterate in Janet, and is overloaded to support all iterable types.

This sort of hints at things but doesn't quite come out and say what an iterable type is (at least in my opinion).

On a related note, I only found [a single hit in janet's source for "iterable"](https://github.com/janet-lang/janet/blob/f021bb28394c5ce939aab1c3471d6fe0c3f7bbc3/src/core/value.c#L134).  The containing function is `janet_next_impl` (used by `janet_next`).  This function provides corroborating evidence (see the quote above) that what one can use `next` on might count as "iterable" and also spells out which types:

* dictionaries
  * table
  * struct
* bytes
  * string
  * keyword
  * symbol
  * buffer
* indexed
  * array
  * tuple
* other
  * abstract (depends on specific abstract)
  * fiber

As for what is not an iterable, it looks like that might include:

* nil
* boolean
* number
* abstract types with no `next` implementation
* functions
* cfunctions
* pointers

Some of the info above was obtained from [here](https://github.com/janet-lang/janet/blob/f021bb28394c5ce939aab1c3471d6fe0c3f7bbc3/src/include/janet.h#L493-L510).

---

Below are a few related thoughts:

1. I think janet-lang.org's main prose might benefit from spelling out a bit more about iterables, i.e. provide a definition and perhaps some examples.  Possibly it could make sense to add something [where `indexed`, `dictionary`, and `bytes` are defined - the data structures page](https://janet-lang.org/docs/data_structures/index.html).  Another idea is to modify the [looping page](https://janet-lang.org/docs/loop.html) and link to it from the aforementioned data structures page.
2. I think it could be that there are more locations in `boot.janet` that could benefit from clarifying whether something is an iterable.
3. The word "iterable" seems on the long side to me in the context of `boot.janet`'s argument names.  My sense is that `boot.janet` mostly seems to favor shorter argument names.  I wonder if `iter` would be an acceptable alternative.

For 1, I think we can work on this at the janet-lang.org repository.

For 2, I think it would be better to try to find other locations in `boot.janet` and consider a single PR that incorporates the changes over this PR.

For 3, it could also be that in the process of attending to 2, we might think of a shorter appropriate name than `iterable` if `iter` does not suffice.

amano-kenji:
Now, the argument name is `iter`, but the docstring still says iterable because I was confused as hell by `xs` as the argument name.

I think more core functions can have clearer docstring or clearer argument names, but in my memory, empty? was the only function that confused me.

We can document iterables somewhere in the core documentation, but even without any core documentation on iterables, I think iter and iterable are far easier to understand than `xs`. I don't think I'm the best person to document iterables in the core documentation.

At least, I intuitively know that `nil` is not an iterable regardless of what iterable really means in janet.

sogaiu:
The current candidate appears to be:

> Check if iterable is empty.

IMO, this is a bit off because of the lack of an indefinite article ("an") since `iterable` is no longer the name of the argument.

I think one of the following would be preferrable:

> Check if an iterable `iter` is empty.

or:

> Check if `iter` is empty.

sogaiu:
> I think more core functions can have clearer docstring or clearer argument names, but in my memory, empty? was the only function that confused me.

Ok.

If some of us go through `boot.janet` and find other related instances and become convinced a different name is warranted, we may suggest changing to such a name.

> We can document iterables somewhere in the core documentation, but even without any core documentation on iterables, I think iter and iterable are far easier to understand than xs.

Perhaps that is true for you, but I think without some idea of what `iter` / `iterable` mean, I'm not sure it's easier to understand.  Just seeing a particular term is not always helpful if one does not have an appropriate concept that the term points to in one's mind.

> I don't think I'm the best person to document iterables in the core documentation.

Fair enough.

sogaiu:
> At least, I intuitively know that nil is not an iterable regardless of what iterable really means in janet.

You're going to love [Common Lisp where `nil` and `()` are synonmous](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node9.html).

amano-kenji:
How do you like it now?

sogaiu:
Looks better, thanks :+1:

sogaiu:
On a related note, the following also work with at least fibers: `geomean`, `mean`, `product`, `sum`, `distinct`.

They all use `xs` in their docstrings, e.g.

```
(geomean xs)
    
Returns the geometric mean of xs. If empty, returns NaN.
```

Except for `distinct`, they also all use the word `empty` in their docstrings.

---

Additionally, the following also seem to work with fibers: `all`, `reduce`, `reduce2`, `some`.

Note that these all use `ind` in their signatures and docstrings.

Issue #1180, opened 2023-06-04
peg-grammar is not available in threads


zevv:
This snippet:
```janet
(ev/spawn-thread       
  (peg/match '{ :main :s* } "")) 
```
results in the following error:
```
error: grammar error in :s*, unknown rule
  in peg/match [src/core/peg.c] on line 1694
  in _spawn-thread [/tmp/t.janet] on line 2, column 3
```

sogaiu:
Yes, although not reported in an issue AFAIK, this has been noted in discussions at one or more of the gitter / matrix channels.

It seems to me that either it should be documented behavior (along with any other caveats [1] when spawning a thread) or the current behavior should be adjusted.

---

[1] I noticed that `eval` didn't give quite the same results in my testing, and I suppose it may be a related thing.

zevv:
I must admit I have not looked deeply at the memory model for threads so I'm not sure what the original design decisions are, but I think conceptually it might make sense to have new threads inherit the parent's dynamic bindings?

sogaiu:
For `coro`, the fiber that's created with `fiber/new` has `:yi` as part of its sigmask, and `:i` is "inherit the environment from the current fiber".

So if I do:

```janet
(ev/thread (coro (pp (peg/match {:main :s*} ""))))
```

I get:

```
@[]
nil
```

Currently `ev/spawn-thread` and `ev/do-thread` don't take any flags, but may be it would be nice if they did?  Something like:

```janet
(ev/spawn-thread :i (peg/match {:main :s*} ""))
```

I don't suppose that would break any code -- an initial keyword in a function body presumably doesn't do much.

bakpakin:
So the `ev/thread` function, which is what does all of the heavy lifting of setting up a new thread, takes some flags to determine what state to copy from the current thread into the new thread. I suppose it would make sense to copy over dynamic bindings as an option if specified - however, copying over the entire environment table as returned by `(curenv)`, table proto types and all, is prohibitively expensive.

> I noticed that eval didn't give quite the same results in my testing, and I suppose it may be a related thing.

Yes, eval needs to have available function definitions in it's  environment, so in order to see all available functions they would need to all be copied to the new thread.

For efficiency, the default behavior is to copy nothing that isn't needed.

bakpakin:
[...]

PR #1703, opened 2026-01-24
Make peg-grammar available in all threads for peg/compile

amano-kenji:
It fixes https://github.com/janet-lang/janet/issues/1180.

Although I don't know how to write C, who knew the pull request is going to be easy for a person who knows almost nothing about C? Today, I'm your code monkey who types shakespeare. The code monkey God gave me an insight for this pull request.

I tested it with

```
(ev/do-thread
  (pp (peg/match '{:main ':s*} " ")))
```

which prints

```
@[" "]
```

Your code monkey also found that sogaiu replaced `default-peg-grammar` with `(dyn :peg-grammar)`. This change broke peg/compile.

### DIFF
--- a/src/core/peg.c
+++ b/src/core/peg.c
@@ -1832,6 +1832,8 @@ static JanetPeg *compile_peg(Janet x) {
         Janet default_grammarv = janet_dyn("peg-grammar");
         if (janet_checktype(default_grammarv, JANET_TABLE)) {
             builder.default_grammar = janet_unwrap_table(default_grammarv);
+        } else {
+            builder.default_grammar = janet_get_core_table("default-peg-grammar");
         }
     }
     builder.tags = janet_table(0);
@@ -1854,8 +1856,8 @@ static JanetPeg *compile_peg(Janet x) {
 JANET_CORE_FN(cfun_peg_compile,
               "(peg/compile peg)",
               "Compiles a peg source data structure into a <core/peg>. This will speed up matching "
-              "if the same peg will be used multiple times. Will also use `(dyn :peg-grammar)` to supplement "
-              "the grammar of the peg for otherwise undefined peg keywords.") {
+              "if the same peg will be used multiple times. `(dyn :peg-grammar)` replaces "
+              "`default-peg-grammar` for the grammar of the peg.") {
     janet_fixarity(argc, 1);
     JanetPeg *peg = compile_peg(argv[0]);
     return janet_wrap_abstract(peg);


sogaiu:
> found that sogaiu replaced default-peg-grammar with (dyn :peg-grammar). This change broke peg/compile.

Perhaps you wouldn't mind indicating the specifics?

amano-kenji:
Ouch. I discovered that bakpakin replaced `default-peg-grammar` with `(dyn :peg-grammar)` in 85155bb2b4b30870eee583a6bf635c6f521dd5cc.

I was careless in monkey mode.

sogaiu:
Thanks for the details.

Errors happen.  No harm done.

sogaiu:
FWIW, according to the commit message for https://github.com/janet-lang/janet/commit/85155bb2b4b30870eee583a6bf635c6f521dd5cc, the change in question was allegedly part of activity supporting [this endeavor](https://github.com/janet-lang/janet/pull/478).

Note that this change has been in place for just about 5 years (from 2021-01-24).

amano-kenji:
Perhaps, default-peg-grammar should become immutable? You can change `(dyn :peg-grammar)` even if default-peg-grammar is immutable.

amano-kenji:
@bakpakin Should `default-peg-grammar` become a struct instead of a table? https://github.com/janet-lang/janet/pull/478 can be an outdated optimization at this point.

> We can avoid loading the full core environment if the mutable parts of core are not reachable from the main function.

It seems to me that you don't really need to modify default-peg-grammar during runtime.... In practice, people just want the default. If people wanted more shared grammar, then they can carry it outside default-peg-grammar. This is also certainly more functional and referentially transparent. I decrease complexity by sticking to referential transparency unless I have to deviate from it.

Or, perhaps this pull request is good enough because most small scripts don't launch a new thread....

bakpakin:
There was some concern originally about accidentally pulling in the entire core library just to compile PEGs, but I think that is not a realistic problem for most users - the current architecture was originally done to decouple Pegs from core. There are plenty of ways to optimize this but I think this is the simplest solution for now.

sogaiu:
> In practice, people just want the default. If people wanted more shared grammar, then they can carry it outside default-peg-grammar.

FWIW, I've added to `default-peg-grammar` before.  That lets you provide things to more than one grammar by only making changes in one place.

bakpakin:
Hmm, I just realized some other issues with this change, perhaps I merged too soon. Let me test some stuff locally, but I think may want to revert this. Sorry @amano-kenji

Issue #986, opened 2022-06-09
os/date should be able to handle a given timezone.


amano-kenji:
For, example

```
(os/date (os/time) "America/New_York")
```

zevv:
On POSIX systems this might work for you:

```
(os/setenv "TZ" "America/New_York")
(os/date (os/time) true)
```

CosmicToast:
This is non-trivial while maintaining strict ISO C99 compatibility (or, indeed, strict ISO C-anything compatibility), as it does not specify any way of doing this.
On POSIX, setting TZ is indeed the correct approach though.

amano-kenji:
@zevv I think `os/date` should document `TZ` environment variable, then.

PR #1704, opened 2026-01-24
Document how to set local timezone for os/date and os/strftime.

amano-kenji:
It turns out that localtime_s for windows and localtime_r for POSIX-compatible operating systems support TZ environment variable for setting local timezone.

It closes https://github.com/janet-lang/janet/issues/986

### DIFF
--- a/src/core/os.c
+++ b/src/core/os.c
@@ -1893,9 +1893,8 @@ static struct tm *time_to_tm(const Janet *argv, int32_t argc, int32_t n, struct 
 JANET_CORE_FN(os_date,
               "(os/date &opt time local)",
               "Returns the given time as a date struct, or the current time if `time` is not given. "
-              "Returns a struct with following key values. Note that all numbers are 0-indexed. "
               "Date is given in UTC unless `local` is truthy, in which case the date is formatted for "
-              "the local timezone.\n\n"
+              "the local timezone. Returns a struct with following key values. Note that all numbers are 0-indexed.\n\n"
               "* :seconds - number of seconds [0-61]\n\n"
               "* :minutes - number of minutes [0-59]\n\n"
               "* :hours - number of hours [0-23]\n\n"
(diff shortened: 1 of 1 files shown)


sogaiu:
Seems a little odd to move the first sentence:

> Returns the given time as a date struct, or the current time if `time` is not given. 

That seems like a good introduction to me, preferrable to starting with:

> Date is given in UTC unless `local` is truthy, in which case the date is formatted for the local timezone.

amano-kenji:
I improved the description.

sogaiu:
Thanks for restoring the first sentence.

---

In the docstring for `file/open` we make the following kind of reference:

> See fopen (<stdio.h> C99) for futher details

AFAIU, `TZ` is specified by POSIX (see [here](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html)) so I think if we figure out a specific label [1] for the appropriate standard we could do something similar here.  This would give a pointer to further information for those who want to learn more details.

Also, in general we have been mostly avoiding putting examples in the docstrings.

I favor there being examples somewhere though and have been involved with others [2] in gradually adding examples to janet-lang.org as described [here](https://github.com/janet-lang/janet-lang.org/blob/0cb794ee62676e894dc9a35c578c4c768a965dd6/README.md#adding-examples).

Thus I suggest moving the specific examples in the docstrings so that they show up at janet-lang.org's examples (which end up visible on the [Core API page](https://janet-lang.org/api/index.html)).

---

[1] Perhaps we can label as "POSIX.1-2024 Chapter 8.3 Other Environment Variables".

[2] There is a record of some of this activity [here](https://github.com/sogaiu/janet-examples/blob/53c79048fe69bd655151dedfb5d238c49cc0cb74/doc/janetuary-examples-2025.md), FWIW.

amano-kenji:
Is it okay now?

sogaiu:
Seems better thanks.

We can add additional info at another point if needed.

sogaiu:
When I do `man localtime_r`, I see `ctime(3)` and that doesn't mention the `TZ` environment variable.  Possibly this is an issue with my system.

Do you see `TZ` mentioned in your documentation?

---

I don't see the `TZ` environment variable in the current version of the PR, are you ok with that?

amano-kenji:
I finally came up with the right docstring.

amano-kenji:
https://github.com/janet-lang/janet-lang.org/pull/318 adds examples.

Issue #1707, opened 2026-01-28
`ev/sleep` is not actually cancelled by `ev/cancel`.


amano-kenji:
```
(def fib (coro
           (try
             (ev/sleep 3)
             ([err f]
              (print "fib's catch")))))

(ev/go
  |(try
     (do
       (resume fib)
       (print "after resume"))
     ([err f]
      (pp err)
      (print "task's catch"))))

(ev/sleep 1)
(ev/cancel fib :cancel)
```

The above code prints

```
fib's catch
"cannot resume fiber with status :dead"
task's catch
```

It is supposed to print

```
fib's catch
after resume
```

* `ev/cancel` seems to turn `fib` into a task.
* `ev/cancel` causes `try` to catch an error, but `ev/sleep` is still not cancelled. `ev/sleep` tries to revive the dead fiber.
* I suspect `ev/cancel` don't actually cancel other `ev` functions which will try to revive dead fibers later. `ev/select` had a similar issue.
* This is how structured concurrency constructs will cancel child tasks in the background. A task resumes a fiber, and `ev/cancel` cancels the fiber. The task must not be cancelled by `ev/cancel` because it needs to run the cleanup code which includes cancelling fibers that are being resumed by child tasks. These fibers may even run on other threads if threads share memory.

sogaiu:
I got the same results and confirm that after `ev/cancel`, `fib` shows up in the return value of `ev/all-tasks`.

bakpakin:
So the bug here is that we are not disallowing this usage of the initial fiber - ev/cancel schedules a fiber to be canceled, even if it is not actually a task fiber.

However, we are not reviving dead fibers, there are multiple guards in place to prevent this from happening. A fiber can be scheduled, canceled for some other reason, and the scheduler will need to deal with the now-dead fiber.

We may want to take greater care not to send duplicate messages to the supervisor channel, which does cause some confusion for anyone trying to coordinate this.

amano-kenji:
If I can only cancel tasks, then how can I prevent `ev/cancel` from cancelling a task in the middle of the clean-up code in `try` and `defer`? I want to tell a task that it should clean up everything including child tasks without cancelling the clean-up code.

If the clean-up code is written by the end user, then the end user can manually receive the clean-up message from a channel, but if a library handles the clean-up in the background, I don't know how it receives the clean-up message via a channel.

amano-kenji:
```
(def fib (coro (ev/sleep 3)))

(def task (ev/go
[...]

PR #1708, opened 2026-01-29
Tweak some fiber-related docstrings

sogaiu:
This PR suggests some changes to some docstrings that mention fibers and/or tasks.

While investigating `ev/cancel` via [this Zulip topic](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/How.20do.20I.20cancel.20a.20fiber.20that's.20being.20resumed.20by.20a.20task.3F/with/570685332), it seemed that the docstring for `ev/cancel` might be adjusted a bit to match the modified behavior hinted at [here](https://github.com/janet-lang/janet/issues/1707#issuecomment-3814766305).

The examination process suggested that possibly some other fiber / task-related docstrings might be changed a bit as well.

The functions / macros in question include:

* `ev/all-tasks`
* `ev/call`
* `ev/cancel`
* `ev/gather`
* `ev/go`
* `ev/spawn`
* `fiber/root`

Except for those of `ev/gather`, the changes are about mentioning tasks more explicitly.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -3865,13 +3865,16 @@
   (defn ev/call
     ```
     Call a function asynchronously.
-    Returns a fiber that is scheduled to run the function.
+    Returns a task fiber that is scheduled to run the function.
     ```
     [f & args]
     (ev/go (fn :call [&] (f ;args))))
 
   (defmacro ev/spawn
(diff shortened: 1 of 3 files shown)


bakpakin:
LGTM - there are certainly a number of places where we can both clarify this and enforce it in code.

PR #1711, opened 2026-02-14
Use snprintf instead of sprintf

sogaiu:
This PR replaces calls to `sprintf` with calls to `snprintf` instead.

None of the `sprintf` calls look problematic but since `snprintf` is used extensively in the rest of the codebase it seemed worth making these changes.


bakpakin:
LGTM!

Issue #1714, opened 2026-02-17
Out of bounds `put` can return uninitialized memory


edsrzf:
After updating to Janet 1.41.1, I'm seeing some strange behavior with code like `(put @"" 10 1)`. I expect this to increase the size of the buffer to 10 and return:

```
@"\0\0\0\0\0\0\0\0\0\0\x01"
```

However with 1.41.1 I sometimes see other results, such as:

```
@"\xFF\0\0\0\0\0\0\0\0\0\x01"
```

The behavior isn't consistent and I'll sometimes see different values if I rearrange the code.

I've looked through the changes in 1.41.1 and https://github.com/janet-lang/janet/commit/8fe284b5eb13479edcf083da6304f88447a339c2 stands out to me. `janet_buffer_setcount` calls `memset` to zero-initialize the new buffer memory, but after that commit, `put` changed to call `janet_buffer_ensure` directly, which does not initialize the memory returned by `realloc`. There's a similar situation with arrays.

edsrzf:
Should be fixed now. Could we please get a patch release for this?

bakpakin:
Yup, I will wait a day or two for some other issues and a bit more QA for the patch release of 1.41.2

PR #1715, opened 2026-02-17
Initialize memory allocated by `put`

edsrzf:
Proposed fix for #1714. One thing I'm unsure about is `janet_putindex`, since that code hasn't changed for years.


bakpakin:
Good catch, the changes here all started due to a discussion about the pathological cases of inserting elements at array[length] or buffer[length] to dynamically grow an array, which is not encouraged in Janet anyway. I would be a little surprised but not completely if there was a long standing issue there.

bakpakin:
So this bug is not present in earlier releases mostly due to the fact the putindex was unused by the compiler in all dynamic cases. However, the implementation was still erroneous, and could be triggered by using custom assembly, e.g. with `asm`.

PR #1718, opened 2026-02-25
Tweak a docstring and a comment

sogaiu:
This PR contains a couple of minor tweaks.

The docstring was tweaked so that the order of the keywords matches that of [what's currently in `util.c`](https://github.com/janet-lang/janet/blob/2602bec0176504e05c956a8ac7675067acd8c231/src/core/util.c#L100-L118) and the comment was a minor modification to the condition for when 0 is returned.


pepe (approved):
LGTM

PR #1720, opened 2026-02-27
Fix GC collecting active fiber during nested janet_continue

brettatoms:
This PR fixes a use-after-free error in Janet when using nested fibers.  There's also a example for how to reproduce the error in `test/test-gc-pcall.c` but since the tests don't compile C files, and this has to be reproduced in C b/c it only happens with `janet_pcall`, I'm not sure if this is the right place for it.

The main issues is that `janet_collect()` marks `janet_vm.root_fiber` but not `janet_vm.fiber`. When `janet_pcall` (or `janet_continue`) is called from a C function, the inner fiber becomes `janet_vm.fiber` while `root_fiber` still points to the outer fiber. If GC triggers inside the inner fiber's execution, the inner fiber is not in any GC root set and can be collected, including its stack memory, while actively running.

This also affects deeply nested cases: F1 -> C func -> `janet_pcall` ->
F2 -> C func -> `janet_pcall` -> F3, where F2 is saved only in a C stack local (`tstate.vm_fiber`), invisible to GC.

Fix: in `janet_continue_no_check`, root the fiber with `janet_gcroot` when this is a nested call (`root_fiber` already set). Each nesting level roots its own fiber, handling arbitrary depth. Top-level calls (event loop, REPL) skip the root/unroot entirely since root_fiber is NULL.

Disclaimer: I used AI to help discover this bug, find the fix and write the test cases to reproduce it.

### DIFF
--- a/src/core/vm.c
+++ b/src/core/vm.c
@@ -1559,6 +1559,15 @@ static JanetSignal janet_continue_no_check(JanetFiber *fiber, Janet in, Janet *o
         }
     }
 
+    /* If this is a nested continue (root_fiber already set), root the fiber
+     * so it survives GC. janet_collect only marks root_fiber, so without
+     * this a nested fiber (e.g., from janet_pcall in a C function) would be
+     * invisible to GC and could be collected while actively running. */
+    int fiber_rooted = (janet_vm.root_fiber != NULL);
+    if (fiber_rooted) {
+        janet_gcroot(janet_wrap_fiber(fiber));
+    }
+
     /* Save global state */
     JanetTryState tstate;
     JanetSignal sig = janet_try(&tstate);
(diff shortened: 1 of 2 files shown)


bakpakin:
Thanks @brettatoms , this looks like a real issue - the assumption the code was making that all relevant fibers will be connected to the root via the ->child pointer, but this isn't always the case, as in pcall. I think it would be a bit much to ask that users call gcroot themselves.

Still, I think we can just unilaterally add a gc root to all fibers and make the logic a bit simpler even than this. I don't think there would be much perf impact.

brettatoms:
Thanks for merging @bakpakin.   Part of this PR were two C programs to reproduce the error.  I put them in the test/ directory from lack of a better place but since they're written in C they don't get run automatically with `make test`.  Should I create a followup PR to make sure these get compiled and run with the rest of the tests?

bakpakin:
That would be appreciated. I ran them locally and was able to repro the bug and the fix with valgrind - presumably address sanitizer would work, so our newly added sanitizer builds should be able to catch a regression here

PR #1721, opened 2026-03-04
Align items in multiline pretty format (%m and %p)

McSinyx:
Indentation levels are not sufficiently clear for deeply nested data structures.  Instead, align items in a collection with its opening bracket, also known as hanging indentation.

To avoid guessing the length of a "short" collection to print it on one line, items are now always printed on separate lines.

### DIFF
--- a/src/core/pp.c
+++ b/src/core/pp.c
@@ -72,7 +72,7 @@ static int count_dig10(int32_t x) {
     }
 }
 
-static void integer_to_string_b(JanetBuffer *buffer, int32_t x) {
+static int32_t integer_to_string_b(JanetBuffer *buffer, int32_t x) {
     janet_buffer_extra(buffer, BUFSIZE);
     uint8_t *buf = buffer->data + buffer->count;
     int32_t neg = 0;
@@ -80,7 +80,7 @@ static void integer_to_string_b(JanetBuffer *buffer, int32_t x) {
     if (x == 0) {
         buf[0] = '0';
         buffer->count++;
-        return;
+        return 1;
     }
     if (x > 0) {
         x = -x;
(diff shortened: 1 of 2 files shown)


bakpakin:
Ok, the copilot review notwithstanding, looks pretty good, thanks @McSinyx! I'm just trying out the copilot for first pass reviews in dense C code especially, mostly as a helper to look for errors that may cause any sort of corruption or logic issues. 

I do generally like the new formatting more than the old.

McSinyx:
Thanks!

BTW will the use of LLMs stops at static analysis?  It would be nice if Janet will not incorporate non-trivial (say, 15+ LoC) snippet generated by LLMs, as they have been demonstrated to output verbatim and near-verbatim copies of their training input, unless the original works are found and their licenses are complied with.  In downstream distributions we are failing to reach consensus on whether projects including LLM output to be considered free software, so it would be kind of you to refrain from doing so (and gently ask contributors to do the same) until there's an official guidance from the FSF or FSFE.

bakpakin:
I have no plans right now to include any significant amount of LLM generated code in Janet, _especially_ anything non-trivial and haven't done so. However, these code review tools have been surprisingly useful at spotting certain kinds of indexing bugs, user after free bugs, etc, especially in complicated C code.

We did recently have some AI generated bug reproduction that found a fiber bug, so I do wonder if that is considered. I don't believe we have had any other significant contributions generated by AI.

McSinyx:
Thanks!  I am not a lawyer, hence the request pending statements from the FSF(E).  From my understanding of copyright though, it only governs artistic _expressions_, so motivations like bug reports or high level ideas are not covered.

bakpakin:
Starting a Zulip discussion here: https://janet.zulipchat.com/#narrow/channel/399615-general/topic/Poll.20for.20Policy.20on.20AI.20contributions.20to.20Janet/with/577653330


This hasn't been even on my radar for Janet until very recently, but things are ramping up fast. Thankfully, Janet is small enough we haven't been inundated by AI slop requests yet

Issue #1725, opened 2026-03-09
Some defmacro* hygiene cases in boot.janet


sogaiu:
Some of the macros in `boot.janet` predate `as-macro` and appear to have hygiene issues.

For example, if one was not aware of the existence of `loop` (or decided to shadow it for some reason):

```
$ janet
Janet 1.41.3-dev-8f446736 linux/x64/gcc - '(doc)' for help
repl:1:> (tabseq [i :in [:a :b]] i i)
@{:a :a :b :b}
repl:2:> (defn loop [&] nil)
<function loop>
repl:3:> (tabseq [i :in [:a :b]] i i)
repl:3:9: compile error: unknown symbol i
```

Note that the second call to `tabseq` has been affected by the new definition of `loop`.

It looks like in this case, it has to do with `tabseq`'s [current definition](https://github.com/janet-lang/janet/blob/8f446736edb302ce57d554d41a81a783e8b36872/src/boot/boot.janet#L696-L701):

```janet
(defmacro tabseq
  ``Similar to `loop`, but accumulates key value pairs into a table.
  See `loop` for details.``
  [head key-body & value-body]
  (def $accum (gensym))
  ~(do (def ,$accum @{}) (loop ,head (,put ,$accum ,key-body (do ,;value-body))) ,$accum))
```

It appears the issue might be addressed by changing the last form of the body to:

```janet
~(do (def ,$accum @{}) (as-macro ,loop ,head (,put ,$accum ,key-body (do ,;value-body))) ,$accum)
```

That is, make use of `as-macro`. 

Looking over the rest of `boot.janet`, I think I see other similar cases (see [next comment](https://github.com/janet-lang/janet/issues/1725#issuecomment-4028296649)).

---

On a side note, the [website macro docs](https://janet-lang.org/docs/macros.html) don't currently mention `as-macro`.  Perhaps some additional text is in order.

---

Thanks to @veqqq for bringing up @ianthehenry's [article](https://ianthehenry.com/posts/janet-game/the-problem-with-macros/).  Although it doesn't mention `as-macro` [1], it discusses macro hygiene in the context of Janet (and Common Lisp) in a particularly accessible fashion.

---

[1] veqqq mentioned ianthehenry's book has a [second chapter on macros](https://janet.guide/macro-mischief/), and `as-macro` is mentioned there.

sogaiu:
[...]

PR #1727, opened 2026-03-10
Macro hygiene tweaks

sogaiu:
This PR is a follow-up to a0cc867f14a8518e2387ef07546f6dce373aeb20 and an attempt to partially address #1725.

Touched items include:

* `when-with`, `if-with`, `when-let` - `if-let` and an associated form was moved to earlier in the file to make things work.
* `loop1`
* `repeat`
* `seq` and `catseq`
* `-?>` and `-?>>`
* `varfn`
* `ev/with-deadline`
* `ffi/defbind-alias` and `ffi/defbind`

I tried to group things into separate commits in case there are problems and some things need to be reverted.

In addition to janet's tests I ran tests across most of my projects and didn't see any issues, but may be other folks could have a look / test bits out?

---

@ianthehenry I broke something before having to do with macros and I think you noticed -- if you have a chance, may be you could take a look at this?

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -708,7 +708,7 @@
   [head & body]
   (def $accum (gensym))
   (check-empty-body body)
-  ~(do (def ,$accum @[]) (loop ,head (,array/push ,$accum (do ,;body))) ,$accum))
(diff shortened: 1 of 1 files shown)


bakpakin:
LGTM

PR #1729, opened 2026-03-12
Fix: add missing :symbolmap keyword to disasm function

PinieP:
The docstring of `disasm` says that `:symbolmap` is a valid keyword arg but this was missing in the implementation.


sogaiu:
I tried the suggested changes and they seem to work:

```
$ ./build/janet
Janet 1.41.3-dev-b160f4f5 linux/x64/gcc - '(doc)' for help
repl:1:> (disasm (fn [] (let [x 1] x)) :symbolmap)
@[(0 2 0 x)]
```

I wonder if it would make sense to align the names though, i.e. the keyword is `:symbolmap`, but the corresponding function is `janet_disasm_symbolslots`.

If so, not sure which direction though :)

PR #1747, opened 2026-05-11
Hide JanetMarshalContext implementation details

iceghost:
This hides JanetMarshalContext behind an opaque pointer, and splits it into two separate Context type for marshal and unmarshal.

The motivation was my confusion while working with this type because the same context type was used for both marshal/unmarshal context, despite containing almost nothing in common, and you cannot certainly use a context that was created for marshal operation for unmarshalling.

This will break current users with compile errors due to the different pointer type on the vtable now, but the ABI should remain the same with the pointer argument.

---

I hope this is not a very controversal change :D I noticed this when trying to stub out the implementation in my Zig port, so thought I would try to also upstream this while I can

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -433,7 +433,7 @@ static int janet_stream_getter(void *p, Janet key, Janet *out) {
 
 static void janet_stream_marshal(void *p, JanetMarshalContext *ctx) {
     JanetStream *s = p;
-    if (!(ctx->flags & JANET_MARSHAL_UNSAFE)) {
+    if (!(janet_marshal_flags(ctx) & JANET_MARSHAL_UNSAFE)) {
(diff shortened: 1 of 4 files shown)


bakpakin:
So this looks functionally fine, but I don't want to break backwards compatibility at the API level. I think some kind of migration at the API level would be desirable to enable this though, perhaps an alias like `typedef JanetUnmarshalContext JanetMarshalContext` for the meantime. Essentially, I want a simple way to write C that will compile correctly for all recent Janet versions.

iceghost:
I see, then I think I will revert my `MarshalContext`/`UnmarshalContext` switch, and only keep the opaque `MarshalContext` part. Making `MarshalContext` no longer part of the ABI is valuable enough for me.

PR #1756, opened 2026-05-25
plan9 port maintenance

pixelherodev:
Fixes a few regressions since the port initially landed

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -424,7 +424,11 @@ static void io_file_marshal(void *p, JanetMarshalContext *ctx) {
         if (iof->flags & JANET_FILE_NOT_CLOSEABLE) {
             fno = fileno(iof->file);
         } else {
+#ifdef JANET_PLAN9
+			fno = dup(fileno(iof->file), -1);
+#else
             fno = dup(fileno(iof->file));
+#endif
         }
 #endif
         janet_marshal_int(ctx, fno);
--- a/src/core/marsh.c
+++ b/src/core/marsh.c
@@ -30,13 +30,6 @@
 #include "util.h"
 #endif
 
-struct JanetMarshalContext {
-    void *m_state;
-    void *u_state;
-    int flags;
-    const uint8_t *data;
-    const JanetAbstractType *at;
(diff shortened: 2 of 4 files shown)


pixelherodev:
This partly reverts the opaquing of JanetMarshalContext, so cc @iceghost.

I'd like to find a way to fix compilation on plan9 without reverting that, since that seems to be an intentional change?

iceghost:
I thought CI also covers Plan9. I'm not sure how to build for Plan9 yet, but I will take a look when I get home in like 9 hours.

(it would be nice if you could include why it failed, since it is not apparent on the commit diff)

iceghost:
Ah, downloading the ISO from https://9front.org is such a pain in the ass with all the rate limiting on their side. My downloads keep on timing out.

I would recommend putting the compile errors here so people who don't have access to a Plan 9 installation could help review this pull request for you. Otherwise the change does not make sense to me.

(Also I think you should fix your system clock...? The commits appear to be authored in 2023)

iceghost:
@bakpakin am I allowed to ask what was exactly the compile error here that led to my change being reverted?

bakpakin:
@iceghost I don't think this reverted anything?

pixelherodev:
It did. One of the changes in 9aadb6775c53191699d8e1230d29ef5db2e8a711 moved the JanetMarshalContext back into the public API.

I think I have a simpler fix;

```c
#ifdef JANET_PLAN9
#pragma incomplete JanetMarshalContext
#endif 
```

in the header should work

pixelherodev:
> I would recommend putting the compile errors here so people who don't have access to a Plan 9 installation could help review this pull request for you. Otherwise the change does not make sense to me.

Whoops, my bad! The tl;dr is that the linker complains that the types of JanetMarshalContext on the internal API mismatch because the incomplete type of the public API is treated as a different signature from the private type.

> (Also I think you should fix your system clock...? The commits appear to be authored in 2023)

Bad CMOS battery. I was fixing it on every reboot but that was a PITA so I stopped >_<

PR #1760, opened 2026-06-08
Fix documentation for os_stat function

taoeffect:


sogaiu:
The suggested change seems like an improvement to me, though I agree with the basic idea suggested by the review comment as well.

taoeffect:
Pushed.

sogaiu:
LGTM

PR #1762, opened 2026-06-10
OSS-Fuzz: Add new fuzzer targets unmarshal

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer for the unmarshal functions, together with the build fix in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15657).


bakpakin:
Thanks @arthurscchan, this looks helpful. I will watch this over the next few days to see if it finds anything

Issue #272, opened 2020-01-22
Comparison operators in inttypes 'mostly dead' code


andrewchambers:
If you browse the implementation of s64 and u64 you will find that most of the comparison methods are implemented, but are actually nearly dead code because the builtin comparison operators do not use them. Instead deferring to janet_compare.

But the question is should we do anything about it? My simple suggestion is that we could remove 'nearly dead' methods from s64 and u64 such as := :> :< :>= :<=.

I'm guessing this is somewhat related to the removal of order< etc.

Overall I find it somewhat of an oddity that comparison can never do coercion but numeric operators sometimes can. I do think this is definitely a corner of the language likely to confuse people and potentially cause subtle errors, though I don't know a better solution without reintroducing something like what was removed recently.

bakpakin:
This an unfortunate consequence of adding a compare method to abstract types. It’s also why a full, weakly typed numeric tower is hard to implement.

This is the reason order>, etc. existed in the first place (strict total ordering, vs type flexible comparison). I removed them for simplicity, but this is a case that needs to be considered.

As for ‘nearly dead’ code, I say keep it. One can use the methods to do comparison when it the type of the left hand side is know.

Another possibility is to allow using these method forms of operators on numbers and other types. This would allow polymorphism when desired. If this were the case, I would say it would be good to remove the method fallbacks for built-in operators as it seems to be a source of confusion. Even pre-change, there were subtle issues like commutivity

andrewchambers:
I see, I definitely am not trying to say what is the best solution, it is very subtle and hard. Overall I ran into this trying to decide which methods were 'worth' implementing for a library:

https://github.com/andrewchambers/janet-bigcalc

bakpakin:
Cool library, looks like there is progress! Not sure what the best solution is. Perhaps it was a mistake to remove order< and friends, although I think it is nice to be able to use <, > to compare strings, tuples, and other types. The previous order functions let us keep a strict total order, while also allowing more friendly, weakly typed operations with the normal operators.

More generally, this is a problem of a good way to do multiple dispatch (multimethods), or at least multiple dispatch for two operands. Janet’s methods really only support single dispatch.

Janet’s model for polymorphism has really been to try and model abstract types as data structures with a few common properties, as associative arrays, and now ordered sequence - this makes some sense as a relatively small functional language. But there are limits, especially for overloading the built in binary operators.

andrewchambers:
[...]

PR #1764, opened 2026-06-11
boot.janet docstring improvements

andrewrothman:
Hi! First contribution to Janet. It has been a very nice language to use.

Please let me know if you have any thoughts on these proposed changes or if changes like this are welcome / unwelcome.

## For third person indicative mood...

This style is used by both Ruby and Lua and to me reads clearer:
- https://docs.ruby-lang.org/en/4.0/Array.html#method-i-to_a
- https://www.lua.org/manual/5.5/manual.html

## For polymorphic compare...

Janet for Mortals chapter 8 discusses this behavior for some built-in functions.

https://janet.guide/tables-and-polymorphism/

> In fact the only built-in functions that use polymorphic compare are zero?, pos?, neg?, one?, even?, and odd? ...

Later in that chapter...

> Tables and abstract types are the only polymorphic values in Janet.

However, tested working for structs:

```janet
# create a prototype with compare function

(def my-proto
	{:compare (fn [self other] (compare (self :v) other))})

# define a function to create a struct with the above prototype

(defn make-test-struct [v]
	(struct/with-proto my-proto :v v))

# make some test structs

(def test-struct-1 (make-test-struct 0))
(def test-struct-2 (make-test-struct 5))

# test polymorphic compare on structs

(print (zero? test-struct-1)) # => true
(print (zero? test-struct-2)) # => false
```

Thanks!

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -96,25 +96,25 @@
   nil)
 
 # Basic predicates
-(defn nan? "Returns true if x is NaN, false otherwise." [x] (not= x x))
-(defn number? "Returns true if x is a number, false otherwise." [x] (= (type x) :number))
-(defn fiber? "Returns true if x is a fiber, false otherwise." [x] (= (type x) :fiber))
-(defn string? "Returns true if x is a string, false otherwise." [x] (= (type x) :string))
-(defn symbol? "Returns true if x is a symbol, false otherwise." [x] (= (type x) :symbol))
-(defn keyword? "Returns true if x is a keyword, false otherwise." [x] (= (type x) :keyword))
-(defn buffer? "Returns true if x is a buffer, false otherwise." [x] (= (type x) :buffer))
-(defn function? "Returns true if x is a function (not a cfunction), false otherwise." [x] (= (type x) :function))
-(defn cfunction? "Returns true if x is a cfunction, false otherwise." [x] (= (type x) :cfunction))
-(defn table? "Returns true if x is a table, false otherwise." [x] (= (type x) :table))
-(defn struct? "Returns true if x is a struct, false otherwise." [x] (= (type x) :struct))
-(defn array? "Returns true if x is an array, false otherwise." [x] (= (type x) :array))
-(defn tuple? "Returns true if x is a tuple, false otherwise." [x] (= (type x) :tuple))
-(defn boolean? "Returns true if x is a boolean, false otherwise." [x] (= (type x) :boolean))
-(defn truthy? "Returns true if x is truthy, false otherwise." [x] (if x true false))
-(defn true? "Returns true if x is true, false otherwise." [x] (= x true))
-(defn false? "Returns true if x is false, false otherwise." [x] (= x false))
-(defn nil? "Returns true if x is nil, false otherwise." [x] (= x nil))
-(defn empty? "Returns true if an iterable, `iter`, is empty, false otherwise." [iter] (= nil (next iter nil)))
+(defn nan? "Checks if x is NaN." [x] (not= x x))
+(defn number? "Checks if x is a number." [x] (= (type x) :number))
+(defn fiber? "Checks if x is a fiber." [x] (= (type x) :fiber))
+(defn string? "Checks if x is a string." [x] (= (type x) :string))
+(defn symbol? "Checks if x is a symbol." [x] (= (type x) :symbol))
+(defn keyword? "Checks if x is a keyword." [x] (= (type x) :keyword))
+(defn buffer? "Checks if x is a buffer." [x] (= (type x) :buffer))
+(defn function? "Checks if x is a function (not a cfunction)." [x] (= (type x) :function))
+(defn cfunction? "Checks if x is a cfunction." [x] (= (type x) :cfunction))
+(defn table? "Checks if x is a table." [x] (= (type x) :table))
+(defn struct? "Checks if x is a struct." [x] (= (type x) :struct))
+(defn array? "Checks if x is an array." [x] (= (type x) :array))
+(defn tuple? "Checks if x is a tuple." [x] (= (type x) :tuple))
+(defn boolean? "Checks if x is a boolean." [x] (= (type x) :boolean))
+(defn truthy? "Checks if x is truthy." [x] (if x true false))
+(defn true? "Checks if x is true." [x] (= x true))
+(defn false? "Checks if x is false." [x] (= x false))
+(defn nil? "Checks if x is nil." [x] (= x nil))
+(defn empty? "Checks if an iterable, `iter`, is empty." [iter] (= nil (next iter nil)))
 
 # For macros, we define an incomplete odd? function that will be overridden.
 (defn odd? [x] (= 1 (mod x 2)))
@@ -128,7 +128,7 @@
    :struct true})
 
 (defn idempotent?
-  "Returns true if x is a value that evaluates to itself when compiled, false otherwise."
+  "Checks if x is a value that evaluates to itself when compiled."
   [x]
   (not (in non-atomic-types (type x))))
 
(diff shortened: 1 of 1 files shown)


sogaiu:
Regarding the text:

> Tables and abstract types are the only polymorphic values in Janet.

Please note that on the same [linked page](https://janet.guide/tables-and-polymorphism/), the disclaimer from the author:

> Remember that tables come in both mutable (“table”) and immutable (“struct”) varieties, but I’m going to use the term “table” loosely to mean “one of Janet’s associative data structures.” They’re exactly the same, apart from the mutability bit.

Also please note that the author of the text says:

> I’m just a fan of Janet; I am not affiliated with the language in any way. I have no real qualifications to be writing a book about it, and nothing that you read here should be considered authoritative, idiomatic, or educational.

via: https://janet.guide/

On a side note, the term "dictionary" is used on at least [one page](https://janet-lang.org/docs/data_structures/index.html) in the website docs to refer to both tables and structs.  There is also a function named `dictionary?`.  It doesn't seem like that term is used much in the existing docstrings though.

---

It is true that the existing docstrings for `zero?` and friends leave things on the underspecified side.  I'm not sure what the intention is there.  I think [this issue](https://github.com/janet-lang/janet/issues/272) might give some hints about why things are the way they are regarding behavior though.

andrewrothman:
> Please note that on the same [linked page](https://janet.guide/tables-and-polymorphism/), the disclaimer from the author:

Ah I did miss that note, good catch, thanks.

> Also please note that the author of the text says:

I should mention I'm a bit fan of the book - the author did a fantastic job (despite their claims that nothing there should be considered "educational" xD)

> It is true that the existing docstrings for zero? and friends leave things on the underspecified side. I'm not sure what the intention is there. I think https://github.com/janet-lang/janet/issues/272 might give some hints about why things are the way they are regarding behavior though.

Thanks for the link, interesting. It would seem best to confirm whether or not that's intended behavior for those symbols before changing their documentation.

> On a side note, the term "dictionary" is used on at least [one page](https://janet-lang.org/docs/data_structures/index.html) in the website docs to refer to both tables and structs. There is also a function named dictionary?. It doesn't seem like that term is used much in the existing docstrings though.

This is also something I was looking into - but figured I'd start small and leave it for future discussion.

In general (personal opinion) I find these qualities really help when using reference documentation:
- Be brief
  - I saw this [commit](https://github.com/janet-lang/janet/commit/5de889419ff26b710b706958bf99e180d084f564) by Calvin mentioned shortening some docstrings to reduce RAM usage, where he noted that docstrings aren't the same as tutorials... This to me makes sense, docstrings seem to be more a form of "reference documentation" which should be brief.
- Focus on "what it's for" and "why you'd want to use it" (basically focus on what observable side-effects it produces or what it returns - what it will do for me, not how it does it)
  - ie. "Check if x is zero" -> "Returns true if x is zero, false otherwise."
- Briefly note important edge-cases and error conditions.
  - Haven't dug into this yet... But ie. `string/split` seems to error on delimiter as empty string - that seems to me like something that should either be considered a bug, or briefly noted in the docstring.
    - ie. "The delimiter must be a non-empty string."

Another thing on indicative mood:
- Seems Python and Julia use imperative mood (as opposed to indicative, which is common in other ecosystems) - there is a good discussion on the Julia [forums](https://discourse.julialang.org/t/docstrings-imperative-or-indicative-mood/121524/20)
  - I would still argue that indicative is the clearer choice - because it's more clearly a description of what the symbol is for vs. a command / suggestion of what the user must do...
  - ie. "Map a function `f` over every value in a data structure `ind`" vs "Maps a function `f` over every value in a data structure `ind`"
    - "Maps a function" makes it clear at a glance that the docstring is describing what the function does when applied, whereas "Map a function" could be read as a command to me the user - that I must "map a function" for some reason, or before use.
  - This could just be the style I'm used to (but it is a broadly popular style) - maybe others have thoughts on this?

McSinyx:
Other here, I prefer imperative mood as the docs can then be used in place of the calls or function signatures.

sogaiu:
I can appreciate wanting consistency for certain aspects of the docstrings, but practically speaking I'm not sure how well that can be made to hold up over time.

FWIW, my sense regarding what seems to have been valued for docstrings in janet itself include (not necessarily in this order):

1. Compactness (I believe this is partly due to the goal of embedding.  For reference, see [this commit message](https://github.com/janet-lang/janet/commit/5de889419ff26b710b706958bf99e180d084f564).)
2. Correctness

This is just what comes to mind when I think back over the PRs I've seen.  Ultimately it's not my call though :)

bakpakin:
I am ambivalent to stylistic changes like tense here, but the point about structs is correct. Older versions of Janet did not have polymorphic structs, but they were later extended to behave more like tables.

sogaiu:
> ... the point about structs is correct. Older versions of Janet did not have polymorphic structs, but they were later extended to behave more like tables.

May be this PR should be just about the structs bit then?

andrewrothman:
Just pushed commits reverting the tense / mood.

Better to rebase it out of the history instead of revert? Let me know.

That leaves only the change to docstrings for polymorphic compare (isn't currently mentioned for tables or structs).

sogaiu:
I don't know how others feel about using the term "dictionary" to cover "table or struct", but FWIW some points in favor:

* it is more concise
* [used in the website docs](https://janet-lang.org/docs/data_structures/index.html)
* there is a function named `dictionary?`
* it is used in the C source in various places

Against might be:

* currently not used in too many docstrings [1]
* folks might not be as familiar with the term

Would like to hear what folks think.

---

[1] [`lengthable?`](https://janet-lang.org/api/index.html#lengthable?) is a notable exception, though we could make an effort to start using it more

andrewrothman:
Some variants to propose:

```
(dictionary? x)
    Check if x is a table or struct.
```

```
(dictionary? x)
    Check if x is a dictionary.
```

```
(dictionary? x)
    Check if x is a dictionary (table or struct).
```

```
(dictionary? x)
    Check if x is a dictionary data structure (table or struct).
```

```
(kvs dict)
    Takes a table or struct and returns a new array of key value pairs
    like @[k v k v ...].
```

```
(kvs dict)
    Takes a dictionary and returns a new array of key value pairs
    like @[k v k v ...].
```

```
(kvs dict)
    Takes a dictionary (table or struct) and returns a new array of key value pairs
    like @[k v k v ...].
```

```
(kvs dict)
    Takes a dictionary data structure (table or struct) and returns a new array of key value pairs
    like @[k v k v ...].
```

```
(kvs dict-ds)
    Takes a dictionary data structure (table or struct) and returns a new array of key value pairs
    like @[k v k v ...].
```

The "dictionary data structure (table or struct)" / "dict-ds" version is a bit more verbose, but does have two advantages:
- Good on familiarity.
- Works cleanly for arrays and tuples as well...

```
(tuple/slice idx-ds [,start=0 [,end=(length idx-ds)]])
    Take a sub-sequence of an indexed data structure (array or tuple) from index start inclusive
    to index end exclusive...
```

For "indexed" - "a dictionary" reads well but "an indexed" not as cleanly - at least to me, though it's probably something one could get used to after reading enough docstrings with that format. (though "idx" as the variable name often means the "index" integer)

Anyway - just some random thoughts for y'all to ponder.

sogaiu:
For `dictionary?` I think:

```
Check if x is a table or struct (aka a dictionary).
```

could work as well.

Upon further reflection, I think it might be better to hold off on whether to make changes that use the term "dictionary".  I'll make a separate issue.

Regarding "indexed" (when just referring to arrays / tuples) -- currently there are at least four variants:

* indexed (e.g. in `lengthable?`'s docstring)
* indexed collection (e.g. in `find`'s docstring)
* indexed type (e.g. in `find-index`'s docstring)
* indexed data structure (e.g. in `frequencies`'s docstring)

Unifying these might make sense, but this seems like something to make another issue from so I'll do that.

In this PR, I suggest we:

* focus on getting the content / spirit of what was referred to in [this comment](https://github.com/janet-lang/janet/pull/1764#issuecomment-4805819113) committed, and
* stick with `table or struct` for this PR

sogaiu:
> Upon further reflection, I think it might be better to hold off on whether to make changes that use the term "dictionary". I'll make a separate issue.

As mentioned [here](https://github.com/janet-lang/janet/issues/1780#issuecomment-5230034703), I'm moving ahead with using the term "dictionary" (or "dictionary type" depending on the situation) along with `dict` and `dicts` for appropriate parameter names.

veqqq:
I second and strongly support the term "dictionary".

Issue #1769, opened 2026-06-24
Closing a stream does not unregister it from EPOLL


Samuel-Martineau:
I am running into what I believe is a bug in the EPOLL implementation of the event loop.

Here is my understanding of the issue. Below is a minimum reproducible example.

This works as I would expect:
1. A socket pair is created with file descriptors `A` / `B`.
2. `A` is `dup`ed to a new file descriptor `X`.
3. `X` is registered as a `JanetStream`.
5. `A` is `dup`ed into a new file descriptor `Y`.
6. `Y` is registered as a `JanetStream`.
7. `X`/`Y` are closed with `janet_stream_close`.
8. `A` / `B` are closed with `close`.

This, to my surprise, fails:
1. A socket pair is created with file descriptors `A`/`B`.
2. `A` is `dup`ed to a new file descriptor `X`.
3. `X` is registered as a `JanetStream`.
4. `X` is closed with `janet_stream_close`.
5. `A` is `dup`ed into a new file descriptor. Since `X` is now free, it is reused by the kernel.
6. Attempting to register `X` as a `JanetStream` throws a "File exists" error.

In short, closing a FD later causes a "File exists" error.

I believe that this is because `EPOLL` is never properly notified of the destruction of the first FD, and so it sees the second FD as a duplicate.

If my understanding is correct, the fix would be to call something akin to
```c
epoll_ctl(janet_vm.epoll, EPOLL_CTL_DEL, stream->handle, NULL);
```
when `janet_stream_close` is called.

---

```c
Janet cfun_epoll_bug(int argc, Janet *argv)
{
    (void)argv;
    janet_fixarity(argc, 0);

    int sockets[2];
    if (socketpair(AF_UNIX, SOCK_STREAM, 0, sockets) == -1)
    {
        janet_panic("socketpair failed");
    }

    int first_dup = dup(sockets[0]);
    if (first_dup == -1)
    {
        close(sockets[0]);
        close(sockets[1]);
        janet_panic("first dup failed");
    }

    JanetStream *first = janet_stream(first_dup, JANET_STREAM_SOCKET | JANET_STREAM_READABLE | JANET_STREAM_WRITABLE, NULL);

    // Closing the first stream causes a "File exists" error when trying to register the second one.
    // If it is left open, the second stream can be registered successfully.
    janet_stream_close(first);

    int second_dup = dup(sockets[0]);
    if (second_dup == -1)
    {
        close(sockets[0]);
        close(sockets[1]);
[...]

PR #1770, opened 2026-06-24
Unregister stream from EV upon closure

Samuel-Martineau:
I am exploring possible fixes to #1769, assuming that it is indeed a bug and not intended behaviour.

**EV Backends:**
- [x] `poll`: Already implemented correctly
- [x] `epoll`: Added and tested
- [x] `kqueue`: Added, but not yet tested (I currently don't have access to a BSD machine...)
- [ ] Windows: I believe is already correct, but I don't have a Windows machine to verify

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -371,15 +371,33 @@ static void janet_stream_close_impl(JanetStream *stream) {
     }
 #else
     if (stream->handle != -1) {
-#ifdef JANET_EV_EPOLL
         if ((stream->flags & JANET_STREAM_UNREGISTERED) == 0) {
+#ifdef JANET_EV_EPOLL
             int status;
             do {
                 status = epoll_ctl(janet_vm.epoll, EPOLL_CTL_DEL, stream->handle, NULL);
             } while (status == -1 && errno == EINTR);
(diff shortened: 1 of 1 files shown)


pepe (commented):
I do not see anything wrong with your PR, but I am not C lang expert, unfortunately. I am sure, it gets reviewed by the owner in a short time.

Anyway, I have the suspicion, that this behavior, you are describing, is what causes some intermittent problems with my applications on Linux. I will think about the best way to test your version, and come back.

bakpakin:
We have some failures on various mac systems due to something off with the kqueue implementation

Samuel-Martineau:
~~Any logs you can share with me?~~ My roommate has a Mac, so I might be able to run some tests myself this weekend...

I just saw the GitHub actions. I'll look into it

Samuel-Martineau:
I am also quite surprised to see a run failure for Windows... I didn't expect this change to have any impact outside of EPOLL/KQUEUE

bakpakin:
I've been investigating the MINGW + Wine failure, as expected it is completely unrelated. LGTM, thanks again @Samuel-Martineau !

Issue #1772, opened 2026-06-26
Garbage collected fiber can still prevent event loop from finishing program


Samuel-Martineau:
When a fiber is suspended manually (ie, by `(yield)`ing, not by waiting for IO), it increments the event loop's reference count.

https://github.com/janet-lang/janet/blob/0e0cbc3ab0787c67e1ca81ca6cf6b97e9e378b4d/src/core/ev.c#L1547-L1550

However, when such a fiber is garbage collected, it does not decrement the reference count, since it does not hold an `ev_state`.

https://github.com/janet-lang/janet/blob/0e0cbc3ab0787c67e1ca81ca6cf6b97e9e378b4d/src/core/gc.c#L333-L341

This has the effect that it is possible for a Janet program to never terminate, since it is "waiting" for a fiber that no longer exists.

---

Here is a minimum reproducible example of this phenomenon.

```c
JANET_FN(cfun_mre, "(mymod/mre)", "...")
{
    janet_fixarity(argc, 0);

    Janet func;
    janet_dostring(janet_core_env(NULL),
                   "(fn [] (forever (yield)))",
                   "mre.janet", &func);

    JanetFiber *fiber = janet_fiber(janet_unwrap_function(func), 64, 0, NULL);

    janet_schedule(fiber, janet_wrap_nil());

    janet_printf("scheduled fiber %v\n", janet_wrap_fiber(fiber));

    return janet_wrap_nil();
}
```

```janet
(mymod/mre)

(ev/sleep 1)

(gccollect)
```

PR #1773, opened 2026-06-26
Do not wait on freed fiber

Samuel-Martineau:
Possible fix for #1772.


bakpakin:
This looks at first glance correct, but the fact that this was missed tells me we should simplify the refcounting logic a bit.

Samuel-Martineau:
That is my feeling too... Let's just say that I am not very confident that this is the only issue of this kind

PR #1778, opened 2026-07-03
support compile with C++ compiler (MinGW G++)

guan-das-jounior:
Hi.

I would be able to compile with G++ on win 11
please keep the PR (if you do not plan to merge it) for future people who face this issue too. 
-- thanks in advance

-----

**summary of work**:
- mostly removed `goto`s
- implicit `enum` -> explicit `enum`
- `template` -> `template_` since in C++ `template` is a keyword

### DIFF
--- a/build_mingw.bat
+++ b/build_mingw.bat
@@ -36,6 +36,8 @@ if not "%JANET_BUILD%" == "" (
     @set CFLAGS=%CFLAGS% -DJANET_BUILD="\"%JANET_BUILD%\""
 )
 
+@if "%1"=="fresh" goto FRESH
+
 @rem Create build directories
 if not exist build mkdir build
 if not exist build\core mkdir build\core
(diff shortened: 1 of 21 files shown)


bakpakin:
Welcome to github! I appreciate the effort, but I think there are some barriers on how to contribute code that are getting in the way.

First, why so many unrelated changes like removing gotos? Generally, the better way to do this sort of thing is to make a minimal change (the only thing that seems needed would be the build script and the renaming of template -> _template), as well as a test build so it doesn't regress.

Second, it's confusing if you keep pushing changes to the same pull request with lots of unrelated changes. Either mark this as a draft, do development in your own local copy only or your own fork, and only make a PR when you are confident enough to get feedback and have the change merged as is.

Last, I don't think being able to compile with C++ is that much of a goal, although I appreciate it might be good for some workflows. Usually you can link C and C++ code together with a common linker.

bakpakin:
Also, please configure your git client to have a commit author name as something other than "Your Name". Perhaps I am wrong, but you seem new to this and I'm not too confident on merging unsolicited changes from a brand new account.

PR #1787, opened 2026-07-30
Fix sorted-by failure when value changes

### LINKED ISSUE #1785: Arrays are not valid `sort(ed)-by` keys
Despite everything being comparable normally, arrays cannot be used as the output of the key function in `sort-by` or `sorted-by`:
```janet
(pp (< [4] [5] [6]))              # true
(pp (sorted-by |[$] [6 5 4]))     # @[4 5 6]

(pp (< @[4] @[5] @[6]))           # true
(pp (sorted-by |@[$] [6 5 4]))    # error: expected integer key for array in range [0, 3), got 3
```

HParker:
In cases where the function provided to sorted-by changes the value every time, we need to protect against walking off the edge of the list. In this case every @[$] creates a new array with a new pointer used for comparison. This means the first sort condition never fails through the whole list and we reach the end.

This prevents us from failing, but since the values are effectively new every time sort sees them, they are not actually sorted.

Fixes: #1785

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -937,8 +937,8 @@
 (defmacro- sort-partition-template
   [ind before? left right pivot]
   ~(do
-     (while (,before? (in ,ind ,left) ,pivot) (++ ,left))
-     (while (,before? ,pivot (in ,ind ,right)) (-- ,right))))
+     (while (and (,before? (in ,ind ,left) ,pivot) (< ,left (- (length ,ind) 1))) (++ ,left))
+     (while (and (,before? ,pivot (in ,ind ,right)) (> ,right 0)) (-- ,right))))
 
 (defn- sort-help [a lo hi before?]
   (when (< lo hi)
--- a/test/suite-boot.janet
+++ b/test/suite-boot.janet
@@ -357,6 +357,8 @@
 (assert (deep= @[1 2 3 4 5] (sorted [5 3 4 1 2])) "sort 3")
 (assert (deep= @[{:a 1} {:a 4} {:a 7}]
                (sorted-by |($ :a) [{:a 4} {:a 7} {:a 1}])) "sort 4")
+(assert (deep= @[1 3 2]
+               (sorted-by |@[$] [1 3 2])) "sort 5")
 
 # Sort function
 # 2ca9300bf


sogaiu:
I wonder if it makes sense to do the ordering comparison before invoking `before?`.

---

Minor note about formatting -- the lines are getting long and it's in a macro with unquoting, so may be breaking things on to multiple lines is worth it?

Something like this?

```janet
(defmacro- sort-partition-template
  [ind before? left right pivot]
  ~(do
     (while (and (< ,left (dec (length ,ind)))
                 (,before? (in ,ind ,left) ,pivot))
       (++ ,left))
     (while (and (< 0 ,right)
                 (,before? ,pivot (in ,ind ,right)))
       (-- ,right))))
```

I may be in the minority about this kind of thing though :sweat_smile: 

---

Appreciate the addition of a regression test.

sogaiu:
I'm not sure whether this change should be made.

Possibly worth seeing the discussion starting around [here](https://github.com/janet-lang/janet/issues/1785#issuecomment-5127346768).

HParker:
acknowledging that this won't get merged, I am still curious if there is a faster way to do this, so far I am learning I have a lot to learn to know what is fast vs. slow in Janet.

```
(def loops 1000000)
(var i 0)
(def loop-total 1000)
(def loop-cond-end 500)

(var start-time (os/clock :monotonic))
(loop [_i :range [0 loops]]
  (set i 0)
  (while (< i loop-cond-end) (++ i))
  )
(var finish-time (os/clock :monotonic))
(var diff (- finish-time start-time))
(print "ran " loops " iterations in " diff " " (/ loops diff) " ips")

(var start-time (os/clock :monotonic))
(loop [_i :range [0 loops]]
  (set i 0)
  (label left-found
         (loop [:repeat loop-total]
           (if (< i loop-cond-end)
             (++ i)
             (return left-found))))

  )
(var finish-time (os/clock :monotonic))
(var diff (- finish-time start-time))
(print "ran " loops " iterations in " diff " " (/ loops diff) " ips")
```

returns:

```
ran 1000000 iterations in 11.2602790002711 88807.7462357658 ips
ran 1000000 iterations in 15.433199999854 64795.3762025673 ips
```

This avoids the conditional, but is almost 30% slower. I wonder if `:repeat` can be made to be as fast as a `while` loop while maintaining a max loops... Maybe there is another way to do this that I am unaware of. I'll think more. Maybe we can discuss in Zulip.

This code is also not very pretty...

sogaiu:
For future reference wrt profiling, possibly of interest might be [these bits in spork](https://github.com/janet-lang/spork/blob/d11159797de554f32c63912bdd73c35e45a38f6b/spork/test.janet#L82-L124).  Some background for `timeit-loop` is [here](https://github.com/janet-lang/spork/issues/127).  Possibly there is some intersection of interest with the author ;)

Also, perhaps this is already know, but FWIW, I recommend having a look at [this interview with bakpakin](https://alexalejandre.com/interviews/interview-with-bakpakin/) to get a sense of where the project is at, what priorities might be, etc.

PR #1799, opened 2026-08-14
int8_t, uint8_t and float getters

dmitrys99:
# What's included

For consistency with other numeric types and for better usability during natives creation this PR adds type getter for:
* `int8_t`
* `uint8_t`
* `float`

### DIFF
--- a/src/include/janet.h
+++ b/src/include/janet.h
@@ -940,6 +940,9 @@ JANET_API Janet janet_nanbox32_from_tagp(uint32_t tag, void *pointer);
 /* End of tagged union implementation */
 #endif
 
+JANET_API int janet_checkfloat(Janet x);
+JANET_API int janet_checkint8(Janet x);
+JANET_API int janet_checkuint8(Janet x);
 JANET_API int janet_checkint16(Janet x);
 JANET_API int janet_checkuint16(Janet x);
 JANET_API int janet_checkint(Janet x);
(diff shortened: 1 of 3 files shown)


bakpakin:
LGTM. Surprised we are missing these but I suppose most of these APIs were added as needed rather than holistically.

src/include/janet.h
```diff
@@ -2209,9 +2215,12 @@ JANET_API void *janet_getpointer(const Janet *argv, int32_t n);
 
 JANET_API int32_t janet_getnat(const Janet *argv, int32_t n);
 JANET_API int32_t janet_getinteger(const Janet *argv, int32_t n);
+JANET_API float janet_getfloat(const Janet *argv, int32_t n);
```
McSinyx, line 2218: Hi, the float interface reversion in ec48673f7c66c04076cdb2591f440c97472db480 misses this declaration.

sogaiu: May be good to make an issue or PR.

McSinyx: Thanks, I've filed https://github.com/janet-lang/janet/pull/1874 (thought it'd be faster for Calvin to commit the deletion but I forgot we use the merge button here).

sogaiu: Thanks for making an issue :+1: 

FWIW, in my experience with this project, I think it has usually worked better to create new issues or PRs.

PR #1802, opened 2026-08-16
Tweak type docstring

sogaiu:
This PR is a suggestion to adjust the docstring for `type`.

Specifically, it includes the following changes:

* adds the missing `:pointer` result
* reorders the keywords to reflect [the order in `janet.h`](https://github.com/janet-lang/janet/blob/291e2222bbdf4d107b1d71e5a7af28f127a9340d/src/include/janet.h#L518-L535) for easier reasoning when viewing the docstring
* adjusts the number of newlines per bullet item [1] to be 1 instead of 2 - my impression is that this results in a nicer viewing experience though YMMV

---

[1] ...except for the last one.


pepe (approved):
LGTM

PR #1804, opened 2026-08-16
Tweak flatten* docstrings

sogaiu:
This PR is an attempt to clarify the terminology in some docstrings along with information about the handled types of the corresponding functions.

* `flatten-into` https://github.com/janet-lang/janet-lang.org/pull/382
* `flatten` https://github.com/janet-lang/janet-lang.org/pull/381

---

A common thread among these docstrings was the use of the phrase "nested array (tree)". It turns out that the functions support tuples as well, so the term "indexed type" was used instead along with renaming the parameter `xs` as `ind`.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1867,20 +1867,24 @@
   ret)
 
 (defn flatten-into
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1805, opened 2026-08-16
Improve completeness of *-in docstrings

sogaiu:
This PR is an attempt to clarify the terminology in some docstrings along with information about the handled types [1] of the corresponding functions.

* `get-in` https://github.com/janet-lang/janet-lang.org/pull/383
* `put-in` https://github.com/janet-lang/janet-lang.org/pull/384
* `update-in` https://github.com/janet-lang/janet-lang.org/pull/385

A common thread among these docstrings was the use of the phrase "nested data structure", which though can be convenient for casual communication might not be clear enough when details are under consideration. Instead, as with some other recent docstring PRs, an attempt was made to spell out more specifically what types appear to be intentionally supported.

Also of note might be that `ks` can be an abstract type, assuming it has appropriate methods.

---

[1] `put-in` and `update-in` use `length`, which prevents fibers from being used for `ks` (though `get-in` doesn't rely on `length`).  If we really wanted to, it seems like the requirement of `length` could be relaxed, but I thought that it would be better to just wait for a request rather than try to add the functionality proactively.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1675,21 +1675,38 @@
     (put res (in ks kk) (in vs vk)))
   res)
 
-(defn get-in
-  ``Access a value in a nested data structure. Looks into the data structure via
-  a sequence of keys. If value is not found, and `dflt` is provided, returns `dflt`.``
-  [ds ks &opt dflt]
-  (var d ds)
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1814, opened 2026-08-18
Add tests for partition

sogaiu:
This PR adds some tests for `partition`. The tests are meant to reflect some intended use cases.

PR #1813, opened 2026-08-18
Add tests for interpose

sogaiu:
This PR adds some tests for `interpose`. The tests are meant to reflect some intended use cases.

PR #1815, opened 2026-08-18
Improve docstrings of filter and friends

sogaiu:
This PR is an attempt to clarify the terminology in some docstrings along with information about the handled types of the corresponding functions.

(This PR was constructed under the assumption that the following PRs are merged first:

* #1812
* #1813
* #1814

The aforementioned 3 PRs all have tests that would probably be better to have in place before this PR gets merged.)

* `filter` https://github.com/janet-lang/janet-lang.org/pull/390
* `interpose` https://github.com/janet-lang/janet-lang.org/pull/392
* `partition` https://github.com/janet-lang/janet-lang.org/pull/393
* `partition-by` https://github.com/janet-lang/janet-lang.org/pull/391

These functions all used `ind` as a parameter name and had docstrings that didn't mention that they could handle more types.  Instead of `ind`, I chose to use the name `x` as we've been doing elsewhere to indicate support for more types like bytes, fibers, and certain abstract types [1].

I also chose to rename some local names in `partition-by` (i.e. `is-new` -> `new?` and `category` -> `categ`) because the result seemed easier to perceive overall without introducing much extra puzzlement (e.g. I think `categ` is pretty obviously an abbreviation for "category").  `partition-by` had a few places where multiple forms were being placed on single lines so the shorter names seemed more in line with that.

---

[1] `filter` seems to be the only one of the four for which it made some sense to indicate that dictionaries are supported (I think sorting the results would still give results that could be useful).  The other three functions can technically take a dictionary but it was more questionable to me whether this should be advertised (so I didn't) because the order of the elements in the returned array seemed to be something one might care about (i.e. sorting seems like it could lose a fair bit of information) and value-order for dictionaries might change depending on the version of `janet` in use.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1152,13 +1152,17 @@
   res)
 
 (defn filter
-  ``Given a predicate, take only elements from an array or tuple for
-  which `(pred element)` is truthy. Returns a new array.``
-  [pred ind]
+  ``
+  Given a predicate, take only elements from `x` for which `(pred
+  element)` is truthy. Returns a new array.
+
+  `x` can be a bytes, indexed, dictionary, fiber, or abstract type
+  with suitable `get` and `next` methods.
+  ``
+  [pred x]
   (def res @[])
-  (each item ind
-    (if (pred item)
-      (array/push res item)))
+  (each v x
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1818, opened 2026-08-19
Improve docstrings for reverse and reverse!

sogaiu:
This PR is an attempt to expand on handled types in the docstrings of some functions.

* `reverse` https://github.com/janet-lang/janet-lang.org/pull/400
* `reverse!` https://github.com/janet-lang/janet-lang.org/pull/401

It turns out that these functions can work with suitable abstract types so this PR tries to reflect some of those details in the docstrings.

Also, the parameter name of `t` was replaced with `x` to be more consistent with the rest of `boot.janet` and follow what we've been doing recently in other docstrings to reflect that the functions can handle a variety of types.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -1626,32 +1626,41 @@
   res)
 
 (defn reverse!
-  `Reverses the order of the elements in a given array or buffer and returns it
-  mutated.`
-  [t]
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1822, opened 2026-08-20
Clarify float check

dmitrys99:
`FLT_MIN` is not opposite to `FLT_MAX`, but it is rather smallest float, close to zero, which can be written to variable of type `float`.

Also, `janet_checkfloatrange()` is called with `x` of type `double`, so comparison `(x) == (float)(x)` does not true for non-integer values.

This PR fixes an issue.


bakpakin:
We really shouldn't be checking floating point ranges at all, and instead simply converting any double to a float and accepting the loss of precision

PR #1824, opened 2026-08-21
Improve docstrings for sum and friends

sogaiu:
This PR is an attempt to expand on handled types in the docstrings of some functions.

* `sum` https://github.com/janet-lang/janet-lang.org/pull/411
* `mean` https://github.com/janet-lang/janet-lang.org/pull/412
* `geomean` https://github.com/janet-lang/janet-lang.org/pull/413
* `product` https://github.com/janet-lang/janet-lang.org/pull/414

It turns out that these functions can work with values that are:

> bytes, indexed. dictionary, fiber, or abstract type with suitable get and next methods

The docstrings were updated to spell these details out.

Also, the parameter name of `xs` was replaced with `x`, to be more consistent with the rest of `boot.janet` and follow what we've been doing recently in other docstrings to reflect that the functions can handle a variety of types.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -760,40 +760,62 @@
   (tuple fiber/new (tuple 'fn :fiber-fn '[] ;body) flags))
 
 (defn sum
-  "Returns the sum of xs. If xs is empty, returns 0."
-  [xs]
+  ``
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1826, opened 2026-08-22
Improve docstring for describe

sogaiu:
This PR is an attempt to spell out more precisely which types of values yield strings that have pointer values in them.

Note there is a companion PR for associated examples: https://github.com/janet-lang/janet-lang.org/pull/417


pepe (approved):
LGTM

PR #1828, opened 2026-08-23
Improve docstring for length

sogaiu:
This PR is an attempt to spell out more fully which types of values can be handled by `length`, namely:

> bytes, indexed, dictionary, or abstract type with a suitable `length` method

Note there is a companion PR for associated examples: https://github.com/janet-lang/janet-lang.org/pull/419


pepe (approved):
LGTM

PR #1829, opened 2026-08-23
Improve docstring for last

sogaiu:
This PR is an attempt to spell out more fully which types of values can be handled by `last`, namely:

> bytes, indexed, or abstract type with a suitable length method

The parameter name `xs` was replaced with `x` to be more in line with other recent docstring PRs.

Note there is a companion PR for associated examples: https://github.com/janet-lang/janet-lang.org/pull/420

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -899,9 +899,14 @@
   (get x 0))
 
 (defn last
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1830, opened 2026-08-23
Improve docstring for loop

sogaiu:
This PR is an attempt to spell out more fully which types of values can be handled by `loop` as well as make some minor tweaks.

Some changes include:

* Which values can be used with `:keys`, `:pairs`, and `:in` are more explicitly described.
* For `:down`, that `start` and `end` are interpreted as endpoints of a half open interval is mentioned (similar to what was already the case for `:range`).  This was part of making the text a bit more similar to `:range`.
* Backticks were used to surround a few more terms (e.g. `def`)
* Some reflow was done to make future comparison a bit easier

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -647,75 +647,85 @@
     (maclintf :normal "empty loop body")))
 
 (defmacro loop
-  ```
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1831, opened 2026-08-25
Improve docstring for table/to-struct

sogaiu:
This PR is an attempt to spell out the behavior of `table/to-struct` more fully.

Specifically, it mentions the optional `proto` argument and that it should be a struct.  Also, I added text pointing out that not specifying the optional argument will result in a struct with no prototype [1].

Note there is a companion PR for associated examples: https://github.com/janet-lang/janet-lang.org/pull/423

---

[1] One of the examples in the companion PR demonstrates that passing in a table with a prototype results in a struct with no prototype.

### DIFF
--- a/src/core/table.c
+++ b/src/core/table.c
@@ -384,7 +384,11 @@ JANET_CORE_FN(cfun_table_setproto,
 
 JANET_CORE_FN(cfun_table_tostruct,
               "(table/to-struct tab &opt proto)",
-              "Convert a table to a struct. Returns a new struct.") {
+              "Create and return a struct based on a table `tab`. "
+              "If given, the optional argument `proto` specifies "
+              "a struct to be used as the newly created struct's "
+              "prototype. Note that if `proto` is not specified, "
+              "the newly created struct will not have a prototype.") {
     janet_arity(argc, 1, 2);
     JanetTable *t = janet_gettable(argv, 0);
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

bakpakin:
Overall, this looks good to me, if a little redundant. However, I would like to do some analysis at the end of this and see what the impact of extending all of these doc strings is for the final binary size.

sogaiu:
> if a little redundant

Fair enough.  I think I wasn't as careful with this one as I was with some of the other ones.

I'll submit a shorter attempt.

sogaiu:
> However, I would like to do some analysis at the end of this and see what the impact of extending all of these doc strings is for the final binary size.

This sounds great.  If you are able to make this procedure something I can carry out, I'm happy to (at least attempt to) incorporate it into the process of attempting future updates...that is, unless it's just looking at compilation results, which I presume I can do now, at least for my environment.

bakpakin:
My "procedure" is pretty simple, basically build janet first normally, and then with `#define JANET_NO_DOCSTRINGS` in janetconf.h (or the corresponding meson option) and find the difference. Last I recall, it was about 200kB but likely a bit larger now

sogaiu:
Thanks!  I'll give it a go here.

PR #1832, opened 2026-08-25
Improve docstring for struct/with-proto

sogaiu:
This PR is an attempt to spell out some more detail for the `struct/with-proto` function.

Specifically, the `proto` and `kvs` arguments are mentioned.  The purpose of `proto` is explained and a reference is made to the `struct` function for `kvs`.

Note there is a companion PR for associated examples: https://github.com/janet-lang/janet-lang.org/pull/425


pepe (approved):
LGTM

Issue #1516, opened 2024-10-30
Built in `errorf`-like construct for `assert`


sogaiu:
`error` has a nice related construct called `errorf`.

I've found myself wanting a similar construct for `assert` [1] on multiple occasions.  To show an example of what code ends up looking like now:

```janet
(assert (and (check-1 a) (check-2 x y z))
        (string/format "A %s format %s that has %d arguments"
                       "nice" "string" 3))
```

with `assertf` this could be something like:

```janet
(assertf (and (check-1 a) (check-2 x y z))
         "A %s format %s that has %d arguments" "nice" "string" 3))
```

Without something like `assertf`, I find myself:

* not wanting to use `assert` as much
* not being as motivated to provide relevant details in the associated message and/or pass fewer arguments to `string/format`

Since `string/format` itself takes up 13 characters (that's more than 15% of my line budget and combined with a space and opening delimiter, it's 15 characters),  the message string template ends up starting a fair bit to the right.

This puts more than a little pressure on the length of the arguments which can lead to cutting of meaningful content, additional lines, and/or "unsavory" hacks [2].

So I think at least in some cases, there may be secondary effects (often adverse?) as well.

I also think that verbose "checking" code tends to make code harder to read because of additional clutter, so some might think this is an additional significant consideration.

---

There has been [some discussion on Zulip](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/assertf.3F/near/479218350) about `assertf`.  My sense is that, so far, folks who have responded [3] are favorable to the idea of inclusion into Janet.

IIUC, ianthehenry has hinted that this [version](https://github.com/ianthehenry/bauble/blob/45e77c5616a97bb77caf176e85cf38753f650de0/src/util/init.janet#L5-L6) might be worth looking at:

```janet
(defmacro assertf [x & args]
  ~(as-macro ,assert ,x (,string/format ,;args)))
```

(Note: text above updated based on feedback.)

Perhaps it would be a good candidate or may be can inspire something appropriate?

For reference, the current implementation of `errorf` is [here](https://github.com/janet-lang/janet/blob/5e443cd29da23efd4454acce82abafc79252e87f/src/boot/boot.janet#L202-L205):

```janet
(defn errorf
  "A combination of `error` and `string/format`. Equivalent to `(error (string/format fmt ;args))`."
  [fmt & args]
  (error (string/format fmt ;args)))
```

---

[...]

PR #1835, opened 2026-08-27
Extend the scope of truthy `if-let` bindings to the `fal` branch

gahr:
The `if-let` construct - together with the fact that only `nil` and `false` are falsey - is very handy to express monadic chains over maybe/optional values:

```janet
(if-let [resp (make-http-request)       # is nil on failure
         {:status stat :body body} resp # always truthy, binds `stat` and `body`
         check (= 200 stat)             # continue only on 200 OK
         json (json/decode body true)   # is nil on failure
         {:foo foo} json]
  foo)
```

However, none of the bindings are available in the false branch, which could be helpful to have for error reporting or fallback logic:

```janet
(if-let [resp (make-http-request)
         {:status stat :body body} resp
         check (= 200 stat)
         json (json/decode body true)
         {:foo foo} json]
  foo
  (if (= 401 stat) # unauthorized, refresh the token
    (try-again)
    nil))
```

This PR makes all symbols introduced by the bindings of an `if-let` available in all branches.

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -451,38 +451,33 @@
   the `fal` form. Bindings have the same syntax as the `let` macro,
   are initially bound to `nil`, and are in scope in both branches.``
   [bindings tru &opt fal]
-  (def len (/ (length bindings) 2))
+  (def len (length bindings))
   (if (= 0 len) (error "expected at least 1 binding"))
-  (if-not (int? len) (error "expected an even number of bindings"))
+  (if (odd? len) (error "expected an even number of bindings"))
   (def fal2 (if macexvar (macexvar fal) fal))
-  (def syms (array/new len))
-  (defn fill-syms [i]
-    (when (< i len)
-      (let [bl (in bindings (* i 2))
-            sym (if (symbol? bl) bl (gensym))]
-        (put syms i sym))
-      (fill-syms (+ 1 i))))
-  (defn aux [i]
+  (def syms (array/new (/ len 2)))
+  (defn make-ifs [i]
     (if (>= i len)
       tru
       (do
-        (def bl (in bindings (* 2 i)))
-        (def br (in bindings (+ 1 (* 2 i))))
+        (def bl (in bindings i))
+        (def br (in bindings (+ 1 i)))
         (if (symbol? bl)
-          ~(if (def ,bl ,br) ,(aux (+ 1 i)) ,fal2)
-          (do
-            (def sym (in syms i))
-            ~(do (def ,sym ,br) (if (def ,bl ,sym) ,(aux (+ 1 i)) ,fal2)))))))
-  
-  (defn wrap [i]
-    (if (>= i len)
-      (aux 0)
-      ~(do
-         (def ,(in syms i) ,nil)
-         ,(wrap (+ 1 i)))))
+          ~(if (def ,bl ,br) ,(make-ifs (+ 2 i)) ,fal2)
+          ~(if (def ,(def sym (in syms (/ i 2))) ,br)
+             (do (def ,bl ,sym) ,(make-ifs (+ 2 i)))
+             ,fal2)))))
 
-  (fill-syms 0)
-  (wrap 0))
+  (defn make-defs [i]
+    (if (>= i len)
+      (make-ifs 0)
+      ~(do
+         ,(let [bl (in bindings i) sym (if (symbol? bl) bl (gensym))]
+            (put syms (/ i 2) sym)
+            ~(def ,sym ,nil))
+         ,(make-defs (+ 2 i)))))
+
+  (make-defs 0))
 
 (defmacro when-with
   ``Similar to with, but if binding is false or nil, returns
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -450,22 +450,38 @@
   evaluate the `tru` form. If any are false or nil, evaluate
   the `fal` form. Bindings have the same syntax as the `let` macro.``
   [bindings tru &opt fal]
-  (def len (length bindings))
+  (def len (/ (length bindings) 2))
   (if (= 0 len) (error "expected at least 1 binding"))
-  (if (odd? len) (error "expected an even number of bindings"))
+  (if-not (int? len) (error "expected an even number of bindings"))
   (def fal2 (if macexvar (macexvar fal) fal))
+  (def syms (array/new len))
+  (defn fill-syms [i]
+    (when (< i len)
+      (let [bl (in bindings (* i 2))
+            sym (if (symbol? bl) bl (gensym))]
+        (put syms i sym))
+      (fill-syms (+ 1 i))))
   (defn aux [i]
     (if (>= i len)
       tru
       (do
-        (def bl (in bindings i))
-        (def br (in bindings (+ 1 i)))
+        (def bl (in bindings (* 2 i)))
+        (def br (in bindings (+ 1 (* 2 i))))
         (if (symbol? bl)
-          ~(if (def ,bl ,br) ,(aux (+ 2 i)) ,fal2)
-          ~(if (def ,(def sym (gensym)) ,br)
-             (do (def ,bl ,sym) ,(aux (+ 2 i)))
-             ,fal2)))))
-  (aux 0))
+          ~(if (def ,bl ,br) ,(aux (+ 1 i)) ,fal2)
+          (do
+            (def sym (in syms i))
+            ~(do (def ,sym ,br) (if (def ,bl ,sym) ,(aux (+ 1 i)) ,fal2)))))))
+  
+  (defn wrap [i]
+    (if (>= i len)
+      (aux 0)
+      ~(do
+         (def ,(in syms i) ,nil)
+         ,(wrap (+ 1 i)))))
+
+  (fill-syms 0)
+  (wrap 0))
 
 (defmacro when-with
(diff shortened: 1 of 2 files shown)


sogaiu:
The existing behavior does seem surprising, given:

```janet
(if-let [a 8
         b nil]
  :hi
  [a b])
# compile error: unknown symbol b
```

because:

```janet
(def b 2)

(if-let [a 8
         b nil]
  :hi
  [a b])
# =>
[8 nil]
```

Trying to think if changing things would break any code...

gahr:
This surely is a change in behaviour (for code that used to compile even with the old code).

As an example, this evaluateds to `"B"` on master and `nil` on this branch.

```janet
(let [b "B"]
  (if-let [a false
           b false]
    "whatever"
    b))
```

I'm unsure how concerned we should be, though.

pepe:
I would not change it. It seems to me, both modes of the operation are valid, imo. I'm on the phone, but also the change seems to me, to be quite involved.

The other thing is that it may be a good thing to mention this behavior in the docstring.

sogaiu:
Looking at the original example...

```janet
{:status stat :body body} resp # always truthy, binds `stat` and `body`
```

Did I misunderstand this?  It doesn't seem to me that "always truthy" holds when `resp` is `nil`.

---

It does seem to me that the `if-let` bindings make the happy path clearer.  However, when there is a failure, it seems more likely it is harder to be confident that failures are being handled properly because there is a group of bindings one must reason about regarding what might have failed.

The following obscures the happy path, but handling the failures seems much clearer to me as there are fewer things one must consider together at any given time:

```janet
(when-let [resp (make-http-request)
           {:status stat :body body} resp]
  (cond
    (= 200 stat)
    (when-let [json (json/decode body true)]
      (get json :foo))
    #
    (= 401 stat)
    (try-again)
    #
    nil))
```

I don't know if there is inherently a trade-off involved.

---

> the change seems to me, to be quite involved.

I'm also concerned by this.

gahr:
The "always truthy" binding only happens if resp is not nil, so it's always truthy whenever it happens.

gahr:
Some of the diffs are due to my changing the implementation to use the number of bindings instead of the number of items in the bindings list. I can undo that part and see if you find it less involved.

Alternatively, if you agree that there's value in this behaviour but would rather not change `if-let`, I can propose to implement my thing as `if-let*`. I can improve the docs too.

You tell me :)

pepe:
I would prefer a new macro, and update the docs for the `if-let`. My 2 cents, as always.

sogaiu:
> if you agree that there's value in this behaviour but would rather not change if-let, I can propose to implement my thing as if-let*.

I do think that the proposed behavior seems less surprising, but like @pepe am wary of changing an existing construct in this fashion.  My vote would be for something under a different name (perhaps like you suggested), if something were to be added.

Since we don't have a way of marking things as explicitly experimental, I think there is some value in having it tried out elsewhere (e.g. in personal projects or possibly `spork`) first [1].  It seems to me that breakage and change of behavior would likely be less of a concern.  That may just be me though :sweat_smile:

The ultimate decision will be by @bakpakin.

---

[1] For reference, some time back [we added `assertf`](https://github.com/janet-lang/janet/issues/1516), a version of which happened to have been independently arrived at by at least two people.  In that case, [we discussed it a bit on Zulip](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/.E2.9C.94.20assertf.3F/near/479218350) before moving forward with an addition, but the discussion took place after one party had already been using a version so it had some kind of track record.

bakpakin:
The original logic for this decision was that if the binding from if-let is known to be falsey or nil, do you really need access to the binding in the falsey branch? I'm sure it's more convenient for certain macros or other styles of code, but for most direct use I don't think it is all that practical.

However, with multiple bindings support in if-let, I suppose it could make some sense. However, the chain logic seems confusing - in the first example, only `resp` would be available, not the the other bindings? The provided test cases also don't exercise some edge cases, such as failure in the middle of such a chain or even executing the falsey case.

More concretely, let's look at the proposed example:

```
(if-let [resp (make-http-request)
         {:status stat :body body} resp
         check (= 200 stat)
         json (json/decode body true)
         {:foo foo} json]
  foo
  (if (= 401 stat) # unauthorized, refresh the token
    (try-again)
    nil))
```

If `resp` is nil, what happens? In the falsey case, is stat nil because it hasn't been set yet? Or do we just raise an error, perhaps defeating the point of the chaining? In short, while chaining is a useful operation, I don't think this proposal is well-formed as is.

gahr:
I think the proposal is well-defined: all bindings are defined at the start, and are bound to nil, so if resip is nil, if-let short-circuits, and all other bindings remain nil. Status is not 401, and we don't reauth, which is the intended behaviour.

Do you think I should spec out the semantics in a clearer way, or is the intended behaviour not logical / intuitive / correct?

gahr:
> The provided test cases also don't exercise some edge cases, such as failure in the middle of such a chain or even executing the falsey case.

oh, thanks for spotting it. I was supposed to have additional `a a b b` bindings in those tests. I'll rewrite them in a clearer manner.

gahr:
I have tried to reduce the differences against master and to simplify the implementation. I have given the helper functions more talking names: `make-defs` for the wrapping function that sets the bindings to nil, and `make-ifs` to the inner function (was `aux`). Hopefully the code is more readable now.

bakpakin:
I see, in the falsey branch we bind everything to nil. I guess it makes some sense. For linting purposes, we should probably add `:unused` to all of the `nil` bindings to avoid linting errors.

bakpakin:
Hmm, code-gen could be better. Setting things to `nil` before is kind of slow, and a lot spurious linting message with strict lint turned on.

```
Janet 1.41.3-dev-8fb4c4f4 linux/x64/gcc - '(doc)' for help
repl:1:> (setdyn *lint-warn* :strict)
:strict
repl:2:> (if-let [x 1 y 2] x x)
repl:2:1: compile warning (strict): binding x is shadowing a binding
repl:2:1: compile warning (strict): binding y is shadowing a binding
repl:2:1: compile warning (strict): dead code, consider removing x
repl:2:1: compile warning (strict): binding y is unused
repl:2:1: compile warning (strict): binding x is unused
1
repl:3:> (macex1 '(if-let [x 1 y 2] x x)
repl:4:(> )
(do (def x nil) (do (def y nil) (if (def x 1) (if (def y 2) x x) x)))
```

gahr:
I'll look into that, thanks!

Edit: I can probably def the subsequent bindings in each falsey branch, so we don't pre-define bindings that end up being truthy.

gahr:
I have changed the approach to avoid shadowing bindings. This also simplifies the logic because I don't need a support table for symbols: now, as soon as I bind a symbol to false/nil, I call `bind-nills` to bind the subsequent symbols in the bindings list.

This made a point clearer, though: I cannot bind to nil symbols that appear inside destructuring bindings, so in my example,

```janet
(if-let [resp (make-http-request)
         {:status stat :body body} resp]
  tru
  fal)
```

I cannot have `stat` and `body` in the false branch, short of analyzing non-symbol bindings and look for symbols in there, recursively, which I'd rather not do... The solution is

```janet
(if-let [resp (make-http-request)
         {:status stat :body body} resp
         stat stat
         body body]
  tru
  fal)
```

which is not super-nice...

What do you think, is this a big-enough drawback to invalidate the whole proposal?

bakpakin:
> What do you think, is this a big-enough drawback to invalidate the whole proposal?

The new changes look good (still a few places to add `:unused` to remove some strict lints), but I was experimenting locally and came to some similar conclusions. I think this macro could be useful but agree with @pepe that this should probably be a new macro, maybe in spork or a utility library.

I was looking at Common Lisp for inspiration here, and I see that the Alexandria utility library has similar macros `if-let`, `when-let`, `when-let*`, and previously had `if-let*` (which corresponds to this proposal), although it was later removed for similar reasons - see https://gitlab.common-lisp.net/alexandria/alexandria/-/commit/d670d4fbe0abc3bb16d9f4c67e19e4b265882ee9 . I couldn't find any discussion on this with a cursory search but I would be curious to see what was said. In that library, `if-let` avoids this issue as all the assignments happen unconditionally, and with `when-let*` there is no else case so this whole issue is avoided.

In general, I think part of the issue is that coercing unbound things to `nil` is very error prone for a number of reasons, and isn't something the rest of the standard library does (save for the `match` macro, which is quite quirky).

sogaiu:
For reference, the [mentioned commit](https://gitlab.common-lisp.net/alexandria/alexandria/-/commit/d670d4fbe0abc3bb16d9f4c67e19e4b265882ee9) seems to be where `if-let*` got added.

I think [this one](https://gitlab.common-lisp.net/alexandria/alexandria/-/commit/130ac5c40c1295f5291162e94bcc9b444d050ed9) is where it got removed.

No luck tracking down any discussions :sweat_smile:

gahr:
Thanks @bakpakin and @sogaiu for your comments and research. I eventually came to this: https://gitlab.common-lisp.net/alexandria/alexandria/-/work_items/10#note_13450

I think I'll drop this proposal. I'll keep it in my source for a while and see how it goes. I might eventually come back and open a PR to have it in spork/misc. What do you think?

sogaiu:
Thanks for the reference, I guess this bit was relevant?

> Because the semantics for IF-LET* are less clear (re. the second leg), I think it should be dropped.

If an alternative `if-let*` does end up working well, may be it would be a good addition to spork.

PR #1836, opened 2026-08-29
free realpath() result with libc allocator

iceghost:
`dest` is allocated with `realpath` in this case, which needs to be deallocated with `free` and not user overridable `janet_free`.


bakpakin:
Good catch!

PR #1837, opened 2026-08-29
Add tests for nested declarations

yohannd1:
As per [discussion on the janet forums](https://janet.zulipchat.com/#narrow/channel/399615-general/topic/Definitions.20in.20expressions/with/619684513), this PR adds regression tests for two special cases that are supported by Janet at the moment, but are not explicitly mentioned anywhere:

```janet
(if (def x ...)
  (<expr that uses x>)
  (<another expr that uses x>))
```

```janet
(def x (def y ...))
(<expr that uses x and y>)
```

### DIFF
--- a/test/suite-specials.janet
+++ b/test/suite-specials.janet
@@ -206,4 +206,43 @@
 (def x (if (= nil x) "B" x))
 (assert (= x "A"))
 
+# Nested declarations
+(do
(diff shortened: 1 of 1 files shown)


bakpakin:
Thanks, this looks good, especially the nested declarations.

PR #1842, opened 2026-09-02
add long options to shell completion scripts

tranzystorekk:
Long options typically make completion (especially iteractive) more readable.

I tested the change for fish, for other shells I'm not a regular user, so help would be appreciated, but I still followed example completion scripts on my system as well as shell documentation.


bakpakin:
Thanks for the improvements, I can confirm the bash works.

I did notice we are missing a few flags, including all of the bundle flags as well as the new --tool flag

tranzystorekk:
I can add missing flags in this PR or another, just assumed we wanted to get already existing options up to speed.

bakpakin:
This is fine as is, we can merge

PR #1843, opened 2026-09-03
FreeBSD should now use inotify

dmarker:
I noticed a question about filewatch over on janet.zulipchat.com, which was actually about macos filewatch.

As promised I verified inotify(2) works on FreeBSD15. In theory this should now be all supported versions of FreeBSD. I did the meson build and `samu -C build test` and everything passes. I don't have linux/windows so they are not tested, ideally they were unchanged.

This did require turning off flags that FreeBSD maybe should ignore but isn't?
linking @markjdb as he woud know.

### DIFF
--- a/src/core/filewatch.c
+++ b/src/core/filewatch.c
@@ -29,7 +29,16 @@
 #ifdef JANET_EV
 #ifdef JANET_FILEWATCH
 
-#ifdef JANET_LINUX
+#if defined(JANET_APPLE) || defined(JANET_BSD)
+#define JANET_KQUEUE 1
+#endif
+#if defined(JANET_LINUX) || defined(__FreeBSD__)
+/* pull FreeBSD off JANET_KQUEUE as it has more functionality with inotify */
+#undef  JANET_KQUEUE
+#define JANET_INOTIFY 1
+#endif
+
+#ifdef JANET_INOTIFY
 #include <sys/inotify.h>
 #include <unistd.h>
 #endif
@@ -38,7 +47,7 @@
 #include <windows.h>
 #endif
 
-#if defined(JANET_APPLE) || defined(JANET_BSD)
+#ifdef JANET_KQUEUE
 #include <sys/event.h>
 #include <sys/stat.h>
 #include <unistd.h>
@@ -60,7 +69,7 @@ typedef struct {
     int is_watching;
 } JanetWatcher;
 
-#ifdef JANET_LINUX
+#ifdef JANET_INOTIFY
 
 #include <sys/inotify.h>
 #include <unistd.h>
@@ -96,7 +105,7 @@ static uint32_t decode_watch_flags(const Janet *options, int32_t n) {
             sizeof(JanetWatchFlagName),
             keyw);
         if (!result) {
-            janet_panicf("unknown linux flag %v", options[i]);
+            janet_panicf("unknown inotify flag %v", options[i]);
         }
         flags |= result->flag;
     }
@@ -122,7 +131,8 @@ static void janet_watcher_add(JanetWatcher *watcher, const char *path, uint32_t 
     if (watcher->stream == NULL) janet_panic("watcher closed");
     int result;
     do {
-        result = inotify_add_watch(watcher->stream->handle, path, flags);
+        /* FreeBSD is not forgiving about returned-only flags, just make sure off */
+        result = inotify_add_watch(watcher->stream->handle, path, flags & ~(IN_IGNORED|IN_ISDIR|IN_Q_OVERFLOW|IN_UNMOUNT));
     } while (result == -1 && errno == EINTR);
     if (result == -1) {
         janet_panicv(janet_ev_lasterr());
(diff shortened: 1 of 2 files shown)


markjdb:
> In theory this should now be all supported versions of FreeBSD.

Not quite: 14.4 does not have inotify, but 14.5 will. Per https://www.freebsd.org/security/ 14.4 won't be EoL until the end of the year.

> This did require turning off flags that FreeBSD maybe should ignore but isn't?
> linking @markjdb as he woud know.

Yes, when I implemented inotify I tried to stay compatible with Linux, but I went off of documented behaviour and third-party test cases, so it's not quite perfect. FreeBSD is just being a bit stricter with invalid flags. I could change it, but that won't appear in any releases for a while, so at this point I'm not sure it's worth bothering.

dmarker:
Moved this to a draft, I'll look into fixing in ports tree for any version that has inotify(2) and come back to this next year when it won't break anybody.

[port review](https://reviews.freebsd.org/D59342) up.

bakpakin:
Interesting, I didn't know inotify was on other systems besides Linux! Thanks for the improvement, I agree inotify provides a better interface and is more useful for the kinds of things I do with Janet.

@dmarker Is there perhaps a compile-time flag we can test for for inotify availability on FreeBSD, or is this something that is simply version 14.5+? We may also want to provide the ability for the runtime to introspect which implementation it is running (inotify, kqueue, or windows) if it's not obvious from `(os/which)`.

Happy to leave this around until we can merge.

dmarker:
> @dmarker Is there perhaps a compile-time flag we can test for for inotify availability on FreeBSD, or is this something that is simply version 14.5+?

The build system for the ports tests for the presence of `/usr/include/sys/inotify.h`, that file doesn't exist on the system unless it has inotify(2). 

Another option is to look at `__FreeBSD_version` in [/usr/src/sys/sys/param.h](https://github.com/freebsd/freebsd-src/blob/4099b5c828800776fda3eccd6537ceed638a2a8b/sys/sys/param.h#L79). It can be tested to be >= 1405000, which is what it was incremented to for FreeBSD14.5.

But I have almost zero `meson` knowledge so I don't know how doable either is.

> Happy to leave this around until we can merge.

That is probably the easiest, and the wait won't be that long. Also if the port update is merged FreeBSD users will have access to the inotify(2) version (except on FreeBSD14.4).

src/core/filewatch.c
```diff
@@ -29,7 +29,16 @@
 #ifdef JANET_EV
 #ifdef JANET_FILEWATCH
 
-#ifdef JANET_LINUX
```
gahr, line 32: how about this?

```suggestion
#if defined(JANET_LINUX)
#define JANET_INOTIFY 1
#elif defined(__FreeBSD__)
#  include <sys/param.h>
#  if __FreeBSD_version >= 1500068
#    define JANET_INOTIFY 1
#  else
#    define JANET_KQUEUE 1
#  endif
#elif defined(JANET_BSD) || defined(JANET_APPLE)
#define JANET_KQUEUE 1
#endif

dmarker: The version would be `1405000` and we need `JANET_APPLE` along with `JANET_BSD`. But that is more elegantly done than my janky way with `undef`.

But there remains fixing the test, `:freebsd` needs to be either with `is-kqueue` or `is-inotify` depending on that same variable.

gahr: Hah yes, I got the version wrong. Also, ITYM 1500050, which was the version at the time of https://reviews.freebsd.org/D50315. So, we probably want the *next* version, which was bumped in [65ed1a035ceb8f7c323c0bc0c7344a7fa7cadc9f](https://cgit.freebsd.org/src/commit/?id=65ed1a035ceb8f7c323c0bc0c7344a7fa7cadc9f), i.e., 1500051.

gahr: Amended to include `JANET_APPLE`

gahr: > But there remains fixing the test, `:freebsd` needs to be either with `is-kqueue` or `is-inotify` depending on that same variable.

Plausible approaches that come to mind:
1. expose a new `(os/features)` API, which returns a list of features among which could be `:inotify`
2. expose a new `(os/version)` API that returns a version string, and check that... for FreeBSD, we could return whatever `freebsd-version -u` returns, e.g., `15.0-RELEASE-p8`, and do a coarse-grained check for > 15.0.

dmarker: Actually [D50315](https://reviews.freebsd.org/D50315) was MFC-ed to stable/14 and first appears with FreeBSD 14.5. The second comment for this pull request is from the author of inotify(2) for FreeBSD! Notice he corrected me on that too so don't feel bad 😄 

But you can check as well by looking for [inotify.h](https://github.com/freebsd/freebsd-src/blob/releng/14.5/sys/sys/inotify.h) in each `releng` branch. Sure enough 14.5 has it.

As of now this is handled in the ports tree: see [commit](https://github.com/freebsd/freebsd-ports/commit/5c01d7a). This means in a week or so latest pkg will have janet 1.42.0 with inotify(2) for FreeBSD15. Even though 14.5 could have it, it won't because packages are build on the oldest major release still supported (so 14.4 until January).

But any FreeBSD14.5 user can get the inotify(2) support by just building themselves the usual way `cd /usr/ports/lang/janet && make install clean`.

Given that users have a solution available, I'm still inclined to just wait until January when all supported versions of FreeBSD will have inotify(2).

Issue #1780, opened 2026-07-12
Some docstrings mention `table or struct` while some say `dictionary`


sogaiu:
This issue is a follow-up to #1764 where it was observed that there is a term "dictionary" which covers both tables and structs, but it is not used much in the existing docstrings.

May be we can explore if it would be worth modifying the existing docstrings to use "dictionary" more extensively and if so, how it might be good to do so.

Some relevant comments on this topic include:

* https://github.com/janet-lang/janet/pull/1764#issuecomment-4871509385
* https://github.com/janet-lang/janet/pull/1764#issuecomment-4878367747
* https://github.com/janet-lang/janet/pull/1764#issuecomment-4950719552

sogaiu:
As mentioned indirectly in one of the linked comments above, the docstring of `kvs` uses `dict` already in a parameter name.

That seems sufficiently short and likely not so confusing [1].

It looks like `tab` has been used for "table" and `st` for "struct" so another option is to use `tab-or-st`.  That's not too long, but I think `ind` [2] has been used for "indexed" (or "indexed type") in a few places [3] (as compared to `tup-or-arr` which I have not seen anywhere [4]).  So use of `dict` seems more consistent with the usage of `ind`.

---

[1] Unlike `str` for "struct", since `str` is used for "string" in multiple places already.

[2] `ind` / `inds` appears to have been used to mean more than "indexed type" in a few places as well.  For example, `drop`'s data structure parameter name is `ind`, but [its implementation](https://github.com/janet-lang/janet/blob/0fea20c82182fe661f75b00a8889d801fe2d79b6/src/boot/boot.janet#L1271-L1284) clearly covers more than just "indexed type" (as interpreted as "tuple or array" like the website docs suggest).

[3] `find-index`'s docstring is:

> Find the index of indexed type for which pred is true. Returns dflt 
> if not found.

Though as noted in [this comment](https://github.com/janet-lang/janet/issues/1781#issuecomment-5020388752), `find-index` can work with more than indexed types so perhaps its data structure parameter name should be changed from `ind` to something else.

[4] But...I just came across `arrtup` in `array/slice` and `tuple/slice`...looks like `arrtup` came into existence [quite a while back](https://github.com/janet-lang/janet/commit/4d119e4e03bffb2c83e208e9be7bf26e01a127db), but I think there are only two instances.

sogaiu:
[...]

PR #1846, opened 2026-09-06
Improve docstring for run-context

sogaiu:
This PR is an attempt to adjust the docstring for `run-context` in a variety of ways.

Some changes include:

* Use of the term "dictionary" instead of "table or struct" (see #1780).
* Reordering of the options for `opts` to match the way they are introduced in the source.  I think this will make understanding and maintaining the source is a little easier.
* Modifying and moving the description of the return value to after the options so that `env` can be used as part of the return value and it can be used after being introduced.
* Used backticks around some more of the text.
* Unified statements about default values and signatures.
* Removed the word "optional" from a few places as all of the options are optional.
* Whitespace and reflow changes

### DIFF
--- a/src/boot/boot.janet
+++ b/src/boot/boot.janet
@@ -2877,41 +2877,48 @@
    :all math/inf})
 
 (defn run-context
-  ```
-  Run a context. This evaluates expressions in an environment,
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1847, opened 2026-09-07
Improve docstring for getproto

sogaiu:
This PR is an attempt to adjust the docstring for `getproto`.

Primarily, it's following up on https://github.com/janet-lang/janet/issues/1780, but it also includes a bit of rewording and reflow as well.


pepe (approved):
LGTM

Issue #1205, opened 2023-06-29
Doc: file/open docs a bit confusing


ChanderG:
This is more of a comment than an issue, but the docstring for the `file/open` function is a bit unclear:
https://github.com/janet-lang/janet/blob/3a4d56afca2eee0e4dbca9a6400757d198e6d6f3/src/core/io.c#L133

I wanted to read/write a file and none of the obvious combinations of "w" and "a" with flags seemed to work. Seeing that the code was directly using the mode keyword with `fopen` and looking at the man page of `fopen` made it immediately clear that what I wanted was `r+` - but this is not clear from the janet function documentation.

sogaiu:
Yes, I agree.

One idea to improve the situation might be to modify `file/open`'s docstring so that it mentions looking at `fopen`'s documentation for further details.

sogaiu:
Perhaps a PR can be created to add something like the following as the last bit of the docstring:

```
See fopen (<stdio.h>, C89, C99) for further details.
```

CosmicToast:
Probably C99 only, since Janet explicitly targets C99.
C89 is quite different on the overall.

sogaiu:
@CosmicToast Thanks for taking a look and the feedback.

I wrote `C89` because I found the following in fopen(3):

```
CONFORMING TO
       fopen(), freopen(): POSIX.1-2001, POSIX.1-2008, C89, C99.
```

Good point about Janet targetting C99 -- not mentioning C89 also makes things shorter :)

Then if folks are good with:

> See fopen (<stdio.h>, C99) for further details.

I'll open a PR.

sogaiu:
@ChanderG It looks like #1231 was merged.

Does it look like this issue has been resolved?  If so, may be it can be closed :)

PR #1848, opened 2026-09-07
Fix a couple of imprecisions in the file/ docs

gahr:
The `+` flag to `(file/open)` was incorrectly documented as being for appending. Instead, this allows to read to files opened with `:w` or `:a`, and to write to files opened with `:r`. This is the same as the `+` flag in `fopen(3)`.

The `(file/seek)` function was documented as having both `whence` and `n` as optional arguments. It also didn't say what the default for `n` was. Anyway, this doesn't make sense, as a default of `:cur` and `0` would effectively make `(file/seek f)` a no-op. I also amended a test so both `(file/seek f :set 0)` and `(file/seek f :set)` are covered and checked to move to the beginning of the file.

### DIFF
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -356,20 +356,18 @@ JANET_CORE_FN(cfun_io_fseek,
         janet_panic("file is closed");
     int64_t offset = 0;
     int whence = SEEK_CUR;
-    if (argc >= 2) {
-        const uint8_t *whence_sym = janet_getkeyword(argv, 1);
-        if (!janet_cstrcmp(whence_sym, "cur")) {
-            whence = SEEK_CUR;
-        } else if (!janet_cstrcmp(whence_sym, "set")) {
-            whence = SEEK_SET;
-        } else if (!janet_cstrcmp(whence_sym, "end")) {
-            whence = SEEK_END;
-        } else {
-            janet_panicf("expected one of :cur, :set, :end, got %v", argv[1]);
-        }
-        if (argc == 3) {
-            offset = (int64_t) janet_getinteger64(argv, 2);
-        }
+    const uint8_t *whence_sym = janet_getkeyword(argv, 1);
+    if (!janet_cstrcmp(whence_sym, "cur")) {
+        whence = SEEK_CUR;
+    } else if (!janet_cstrcmp(whence_sym, "set")) {
+        whence = SEEK_SET;
+    } else if (!janet_cstrcmp(whence_sym, "end")) {
+        whence = SEEK_END;
+    } else {
+        janet_panicf("expected one of :cur, :set, :end, got %v", argv[1]);
+    }
+    if (argc == 3) {
+        offset = (int64_t) janet_getinteger64(argv, 2);
     }
     if (fseek(iof->file, offset, whence)) janet_panic("error seeking file");
     return argv[0];
--- a/src/core/io.c
+++ b/src/core/io.c
@@ -150,7 +150,7 @@ JANET_CORE_FN(cfun_io_fopen,
               "* a - append to the file\n\n"
               "Following one of the initial flags, 0 or more of the following flags can be appended:\n\n"
               "* b - open the file in binary mode (rather than text mode)\n\n"
-              "* + - append to the file instead of overwriting it\n\n"
+              "* + - open the file for both reading and writing\n\n"
               "* n - error if the file cannot be opened instead of returning nil\n\n"
               "See fopen (<stdio.h>, C99) for further details.") {
     janet_arity(argc, 1, 3);
@@ -342,12 +342,12 @@ JANET_CORE_FN(cfun_io_fclose,
 
 /* Seek a file */
 JANET_CORE_FN(cfun_io_fseek,
-              "(file/seek f &opt whence n)",
+              "(file/seek f whence &opt n)",
               "Jump to a relative location in the file `f`. `whence` must be one of:\n\n"
               "* :cur - jump relative to the current file location\n\n"
               "* :set - jump relative to the beginning of the file\n\n"
               "* :end - jump relative to the end of the file\n\n"
-              "By default, `whence` is :cur. Optionally a value `n` may be passed "
+              "By default, `n` is 0. Optionally a value `n` may be passed "
               "for the relative number of bytes to seek in the file. `n` may be a real "
               "number to handle large files of more than 4GB. Returns the file handle.") {
     janet_arity(argc, 2, 3);
--- a/test/suite-io.janet
+++ b/test/suite-io.janet
@@ -57,8 +57,8 @@
   (assert (= 0 (file/tell f)) "start of file again")
   (assert (= (string (file/read f :all)) "foo\n") "temp files work")  
   (file/seek f :set)
-  (assert (= 0 (file/tell f)) "start of file again")
-  (assert (= (string (file/read f :all)) "foo\n") "temp files work"))
+  (assert (= 0 (file/tell f)) "start of file again 2")
+  (assert (= (string (file/read f :all)) "foo\n") "temp files work 2"))
 
 # issue #1055 - 2c927ea76
 (let [b @""]
--- a/test/suite-io.janet
+++ b/test/suite-io.janet
@@ -53,10 +53,12 @@
   (file/write f "foo\n")
   (assert (= 4 (file/tell f)) "after written string")
   (file/flush f)
-  (each arg [[:set 0] [:set]]
-    (file/seek f ;arg)
-    (assertf (= 0 (file/tell f)) "start of file again - args: %j" arg)
-    (assertf (= (string (file/read f :all)) "foo\n") "temp files work - args: %j" arg)))
+  (file/seek f :set 0)
+  (assert (= 0 (file/tell f)) "start of file again")
+  (assert (= (string (file/read f :all)) "foo\n") "temp files work")  
+  (file/seek f :set)
+  (assert (= 0 (file/tell f)) "start of file again")
+  (assert (= (string (file/read f :all)) "foo\n") "temp files work"))
 
 # issue #1055 - 2c927ea76
 (let [b @""]
--- a/test/suite-io.janet
+++ b/test/suite-io.janet
@@ -55,8 +55,8 @@
   (file/flush f)
   (each arg [[:set 0] [:set]]
     (file/seek f ;arg)
-    (assert (= 0 (file/tell f)) "start of file again")
-    (assert (= (string (file/read f :all)) "foo\n") "temp files work")))
+    (assertf (= 0 (file/tell f)) "start of file again - args: %j" arg)
+    (assertf (= (string (file/read f :all)) "foo\n") "temp files work - args: %j" arg)))
 
 # issue #1055 - 2c927ea76
 (let [b @""]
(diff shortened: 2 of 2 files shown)


sogaiu:
A quick comment about the tests, hope to take a closer look at the rest soon.

The way they are expressed now, can we tell which one(s) failed from the error messages?

May be it's ok to have a slight amount of repetition here.

sogaiu:
For `file/open`'s `+`, I went through `io.c` looking at where `JANET_FILE_UPDATE` is used and did some brief testing and I think the suggested change for the corresponding portion of this PR makes sense.

---

On a side note, I noticed:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L449

and:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L452-L453

I wonder if it makes sense to be checking for `JANET_FILE_UPDATE` here since [`JANET_FILE_*` seem to be mutually exclusive](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/include/janet.h#L2274-L2282).

sogaiu:
> The (file/seek) function was documented as having both whence and n as optional arguments. 

`(file/seek <f>)` errors because of:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L353

So the suggested change to make `whence` non-optional, i.e. `(file/seek f whence &opt n)`, seems correct.

> It also didn't say what the default for n was.

That did seem to be missing and the suggested change here looks good to me.

---

On a side note, since:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L353

may be:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L359

is unneeded?

gahr:
> On a side note, I noticed:
> 
> https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L449
> 
> and:
> 
> https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L452-L453
> 
> I wonder if it makes sense to be checking for `JANET_FILE_UPDATE` here since [`JANET_FILE_*` seem to be mutually exclusive](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/include/janet.h#L2274-L2282).

Yes, `JANET_FILE_READ`, `JANET_FILE_APPEND`, and `JANET_FILE_WRITE` are mutually exclusive. I think this should be an `else if`

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L450

sogaiu:
Regarding the tests:

```janet
  (each arg [[:set 0] [:set]]
    (file/seek f ;arg)
    (assert (= 0 (file/tell f)) "start of file again")
    (assert (= (string (file/read f :all)) "foo\n") "temp files work")))
```

I think this way of doing things means that if there is an error, we can't tell which combination resulted in the error.  Does that seem right?

Since there is hardly any repetition, I wonder if it would be reasonable to not loop and just spell out the invocations, e.g.:

```janet
(file/seek f :set 0)
(assert (= 0 (file/tell f)) "start of file again 1")
(assert (= (string (file/read f :all)) "foo\n") "temp files work 1"))
(file/seek f :set)
(assert (= 0 (file/tell f)) "start of file again 2")
(assert (= (string (file/read f :all)) "foo\n") "temp files work 2"))
```

sogaiu:
Regarding:

> Yes, JANET_FILE_READ, JANET_FILE_APPEND, and JANET_FILE_WRITE are mutually exclusive. I think this should be an else if

Sorry, may be I wrote in an unclear and/or incorrect fashion earier.  It's true that the values are mutually exclusive when checking using bitmasks, but in terms of the logic of the code, it seems to me that whenever one is checking for `JANET_FILE_WRITE`, there is also a check for `JANET_FILE_UPDATE` (and similarly for `JANET_FILE_READ` + `JANET_FILE_UPDATE`).

For example:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L194

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L262

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L282

So `JANET_FILE_UPDATE` being set effectively gives the behavior of having both `JANET_FILE_READ` and `JANET_FILE_WRITE` at the same time?

---

For:

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L449-L454

it doesn't seem like we are accounting for something that originally had `JANET_FILE_UPDATE` set...but it's quite possible I'm misunderstanding how all of this code works :sweat_smile:

gahr:
Either that, or we dump the arguments in the asserts.. I have no strong opinion.

gahr:
> So `JANET_FILE_UPDATE` being set effectively gives the behavior of having both `JANET_FILE_READ` and `JANET_FILE_WRITE` at the same time?

Yes, exactly the same semantics as in fopen, which is where that `+` ends up, cause we're passing it the flags verbatim:
https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L161

But I don't understand your comment: are you suggesting a change, pointing out an error, or what? :D

sogaiu:
Sorry, I think what I'm trying to point out should lead to a new issue / PR.

What I'm trying to say is that when unmarshalling, we don't seem to be accounting for something that had `JANET_FILE_UPDATE` set originally (before marshalling).

There are checks for `JANET_FILE_READ`, `JANET_FILE_WRITE`, and `JANET_FILE_APPEND` in [this function which seems to be involved in unmarshalling](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L442-L474), but I don't see any accounting for `JANET_FILE_UPDATE` (nor for `JANET_FILE_BINARY` nor `JANET_FILE_NONIL`).

Shouldn't there be?  It seems like information is lost.

Even if so, this PR is probably not the place to address it.

gahr:
> Sorry, I think what I'm trying to point out may be a separate issue.
> 
> What I'm trying to say is that when unmarshalling, we don't seem to be accounting for something that had `JANET_FILE_UPDATE` set originally (before marshalling).
> 
> There are checks for `JANET_FILE_READ`, `JANET_FILE_WRITE`, and `JANET_FILE_APPEND` in [this function which seems to be involved in unmarshalling](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/src/core/io.c#L442-L474), but I don't see any accounting for `JANET_FILE_UPDATE`.
> 
> Shouldn't there be? It seems like information is lost.
> 
> Even if so, this PR is probably not the place to address it.

Ahhh got you! I am not familiar with marshalling, but it surely looks like it can be handled in a separate PR ;) I'll take a look.

sogaiu:
I think janet's [tests override `assert`](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/helper.janet#L43-L50) but possibly not `assertf`.  I'm not sure the changes will work as expected.

It's may be simplest just to "unroll the loop" (as suggested in the latter half of [this comment](https://github.com/janet-lang/janet/pull/1848#issuecomment-5581848620)) but perhaps there is some other approach.

sogaiu:
> I am not familiar with marshalling, but it surely looks like it can be handled in a separate PR ;) I'll take a look.

I'm not really sure how things should work here, may be an issue is better at this point?  I can do that unless you would prefer to.

gahr:
> I think janet's [tests override `assert`](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/helper.janet#L43-L50) but possibly not `assertf`. I'm not sure the changes will work as expected.
> 
> It's may be simplest just to "unroll the loop" (as suggested in the latter half of [this comment](https://github.com/janet-lang/janet/pull/1848#issuecomment-5581848620)) but perhaps there is some other approach.

Oh, I tested it and it worked, when I changed a test to fail.. but sure, let me unroll if you prefer.

gahr:
> > I am not familiar with marshalling, but it surely looks like it can be handled in a separate PR ;) I'll take a look.
> 
> I'm not really sure how things should work here, may be an issue is better at this point? I can do that unless you would prefer to.

Go ahead, thanks!

sogaiu:
> Oh, I tested it and it worked, when I changed a test to fail.. but sure, let me unroll if you prefer.

I think the "custom" `assert` calls [`assert-no-tail`](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/helper.janet#L15-L35) which looks like it does some kind of book-keeping that mutates [these "globals"](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/helper.janet#L6-L11).  I think using `assertf` may bypass all of that.  Also, `assert-no-tail`'s behavior can be tweaked via [is-verbose](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/helper.janet#L13) which is influenced by the `VERBOSE` environment variable.

I think one safer approach is to rewrite using the "custom" `assert`.  Possibly there's some other way, but I don't know what that would be off-hand.

gahr:
Good catch about `assert` not being the real `assert` :) I have unrolled the loops

sogaiu:
@gahr With the way things are currently, the messages end up being the same for some of the cases so we won't be able to tell apart 1 from 3 and 2 from 4:

```janet
  (file/seek f :set 0)
  (assert (= 0 (file/tell f)) "start of file again")                   # 1
  (assert (= (string (file/read f :all)) "foo\n") "temp files work")   # 2
  (file/seek f :set)
  (assert (= 0 (file/tell f)) "start of file again")                   # 3
  (assert (= (string (file/read f :all)) "foo\n") "temp files work"))  # 4
```

I think it's worth having the strings associated with the tests being different (just appending different numbers seems to be a convention being followed [1]), e.g. like the code in the latter half of [this comment](https://github.com/janet-lang/janet/pull/1848#issuecomment-5581848620).

---

[1] For example, see [these lines](https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/suite-vm.janet#L99-L102):

https://github.com/janet-lang/janet/blob/359d9f9d496f5a9081ac3a9a930b69fd76ad626b/test/suite-vm.janet#L99-L102

sogaiu:
LGTM

I tried locally and the tests seemed fine too.

On a historical note, I think one of the changes in this PR further addresses https://github.com/janet-lang/janet/issues/1205.

PR #1849, opened 2026-09-08
Don't hardcode 127.0.0.1 in tests

gahr:
How about we resolve through the special host name "localhost." instead of assuming that the loopback interface is bound to 127.0.0.1 (which might not be the case, especially in FreeBSD jails)?

See RFC6761 for how "localhost." is treated, especially:

Name resolution APIs and libraries SHOULD recognize localhost names as special and SHOULD always return the IP loopback address for address queries and negative responses for all other query types.  Name resolution APIs SHOULD NOT send queries for localhost names to their configured caching DNS server(s).


bakpakin:
Not a bad idea, but I've definitely tested on hosts where DNS is setup up oddly or wrong and localhost. does not resolve as you would expect.

The CI error is concerning though on windows:

```
error: Failed to get localname on <core/stream [fd=348]>: The system detected an invalid pointer address in attempting to use a pointer argument in a call.
```

bakpakin:
I cannot reproduce locally or fix this in CI - Usual trouble-shooting did find anything interesting other than than getsockname is failing with the win32 error code WSAEFAULT, which the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getsockname) suggests is:

> [WSAEFAULT](https://learn.microsoft.com/en-us/windows/desktop/WinSock/windows-sockets-error-codes-2)
	The name or the namelen parameter is not a valid part of the user address space, or the namelen parameter is too small.
	
Adding plenty of extra buffer space does not fix this, though. The name buffer should also trivially be part of the address space, we pass an address to a local stack variable.  I suspect it may have something to do with how we are using AcceptEx.

gahr:
I'm sorry, I have zero knowledge of Windows. Would you rather we dropped this?

bakpakin:
We can leave this for now - I overall like it but breaking CI is obviously an issue. I've tagged this PR as a bug to avoid  pening a new issue. We should at least figure out what is going on

gahr:
I'll give it a shot

gahr:
I managed to get a Windows instance running in EC2, and I discovered that this is related to IPv6, i.e., when `localhost.` resolves to `::1`. In particular, the issue is when a server does a `(net/localname s)` within a `(net/server)` handler.

Small reproducer:

```janet
(def *host* "::1")
(def *port* 12345)

(defn dump [what [host port]]
 (print what ": " host ", " port))

(defn handler [s]
 (dump "server peer" (net/peername s))
 (dump "server local" (net/localname s))  # this line has issues
 (ev/write s " "))

(with [s (net/server *host* *port* handler)]
  (with [c (net/connect *host* *port*)]
    (dump "client local" (net/localname c))
    (dump "client peer" (net/peername c))
    (ev/read c 1)))
```

```
C:\janet>janet test.janet
client local: ::1, 51523
client peer: ::1, 12345
server peer: ::, 51523
error: Failed to get localname on <core/stream [fd=248]>: The system detected an invalid pointer address in attempting to use a pointer argument in a call.
  in net/localname [src/core/net.c] on line 946
  in handler [test.janet] on line 9, column 23
```

There's something fishy going on in the `server peer` call already, as I would expect `::1`, not `::`. Then, the expected `server local` line is where it explodes.

This PR causes the issue because `localhost` resolves to `::1`. This breaks just the same:

```
C:\janet>git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.

C:\janet>set JANET_TEST_HOST ::1
JANET_TEST_HOST=::1

C:\janet>janet test\suite-ev.janet
Starting suite test\suite-ev.janet...
error: Failed to get localname on <core/stream [fd=768]>: The system detected an invalid pointer address in attempting to use a pointer argument in a call.
  in net/localname [src/core/net.c] on line 946
  in defer [test\suite-ev.janet] on line 226, column 22
  in names-handler [test\suite-ev.janet] on line 223, column 3
...
```

I'll play more with the Windows EC2 instance later...

gahr:
I have a hunch that hardcording `AF_INET` here might not be right...

https://github.com/janet-lang/janet/blob/716d9f7b5b03c37a126d92cd4ec72678679c4773/src/core/net.c#L304

but I can't check it now anymore.. stay tuned..

bakpakin:
Oh perfect, thank you so much! I had managed to get the repro locally, but yes, AF_INET was the issue. Replacing it with AF_UNSPEC seems to fix it!

That said, windows docs seem to discourage it, but that is certainly the issue

gahr:
Thanks, rebased.

PR #1851, opened 2026-09-08
added math/get-seed extracting a rng's seed

Arteneko:
it stores the seed on an extra field for later retrieval

two tests were added and i've built and tested this by hand locally (macos m3)


veqqq:
By default, the seed is 0. So this would only ever give you a 0, unless you provide another seed (and thus know it, not needing this.)

So I don't think there's any use case for this change.

Arteneko:
as discussed on [the forum](https://janet.zulipchat.com/#narrow/channel/409517-help/topic/Getting.20the.20seed.20from.20a.20rng.20object/with/622587737) this PR isn't needed anymore

PR #1856, opened 2026-09-09
Use backticks for nicer site rendering

sogaiu:
This PR mostly tweaks some markup in some docstrings to improve the rendering on the website's core api page.

Specifically, backticks were used to surround some constructs and some backslashes were removed in the process.

I'll leave some links below to the 1.42.0 versions of the docs so that the current states can be observed:

* [ffi/trampoline](https://janet-lang.org/1.42.0/api/index.html#ffi/trampoline)
* [os/open](https://janet-lang.org/1.42.0/api/index.html#os/open)

Note the backslash "leakages".

PR #1857, opened 2026-09-09
Unify assignment operator docstrings

sogaiu:
This PR attempts to "unify" the assignment operator (e.g. `+=`) function docstrings.

Suggested changes include:

* Having the docstrings refer to the `ns` parameter with splicing rather than `n`.
* Alignment of two of the docstrings (i.e. for `+=` and `-=`) to the remaining ones.
* Tweaking of markup to prevent backslash leakage (similar to #1856).


pepe (approved):
This way, it is much better. LGTM

PR #1858, opened 2026-09-10
Modify some buffer docstrings

sogaiu:
This PR contains some suggested changes to a number of `buffer/` docstrings.

Some changes include:

* Make `buffer/format-at`'s docstring different from that of `buffer/format`'s by mentioning its `at` parameter and its purpose.
* Use `buf` instead of `buffer` to avoid using an existing function name as a parameter name [1].
* Use `n` instead of `count` (for reasons similar to those described in [1]).
* Use the `buf` parameter in place of "the buffer" in a number of locations to reduce docstring size.
* Use "needed" instead of "necessary" to reduce docstring size.
* Use "bytes type" instead of "byte sequence" to more closely align with recent docstring changes.
* Mention some more parameter names in some docstrings.
* Surround some parameter names with backticks.
* Some reflow.

---

[1] Although this is in the C source code, I think it's nicer to adopt the convention of avoiding collisions between parameter names and existing construct names.  This is less confusing but can also be nicer when performing searches.

### DIFF
--- a/src/core/buffer.c
+++ b/src/core/buffer.c
@@ -204,9 +204,9 @@ JANET_CORE_FN(cfun_buffer_new,
 }
 
 JANET_CORE_FN(cfun_buffer_new_filled,
-              "(buffer/new-filled count &opt byte)",
-              "Creates a new buffer of length `count` filled with `byte`. By default, `byte` is 0. "
-              "Returns the new buffer.") {
+              "(buffer/new-filled n &opt byte)",
+              "Creates a new buffer of length `n` filled up with `byte`. "
+              "By default, `byte` is 0. Returns the new buffer.") {
     janet_arity(argc, 1, 2);
     int32_t count = janet_getinteger(argv, 0);
     if (count < 0) count = 0;
@@ -236,9 +236,9 @@ JANET_CORE_FN(cfun_buffer_frombytes,
 }
 
 JANET_CORE_FN(cfun_buffer_fill,
-              "(buffer/fill buffer &opt byte)",
-              "Fill up a buffer with bytes, defaulting to 0s. Does not change the buffer's length. "
-              "Returns the modified buffer.") {
+              "(buffer/fill buf &opt byte)",
+              "Fill up a buffer, `buf`, with bytes, defaulting to 0s. Does "
+              "not change `buf`'s length. Returns `buf`.") {
     janet_arity(argc, 1, 2);
     JanetBuffer *buffer = janet_getbuffer(argv, 0);
     int32_t byte = 0;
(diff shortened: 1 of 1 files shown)


src/core/buffer.c
```diff
@@ -555,8 +559,9 @@ static void bitloc(int32_t argc, Janet *argv, JanetBuffer **b, int32_t *index, i
 }
 
 JANET_CORE_FN(cfun_buffer_bitset,
-              "(buffer/bit-set buffer index)",
-              "Sets the bit at the given bit-index. Returns the buffer.") {
+              "(buffer/bit-set buf index)",
+              "Sets the bit at the given bit-index in a buffer, `buf`. "
```
pepe: Maybe rename the argument to bit-index? Or the other way around?

src/core/buffer.c
```diff
@@ -566,8 +571,9 @@ JANET_CORE_FN(cfun_buffer_bitset,
 }
 
 JANET_CORE_FN(cfun_buffer_bitclear,
-              "(buffer/bit-clear buffer index)",
-              "Clears the bit at the given bit-index. Returns the buffer.") {
+              "(buffer/bit-clear buf index)",
+              "Clears the bit at the given bit-index in a buffer, `buf`. "
```
pepe: Dto. for renaming

src/core/buffer.c
```diff
@@ -577,8 +583,9 @@ JANET_CORE_FN(cfun_buffer_bitclear,
 }
 
 JANET_CORE_FN(cfun_buffer_bitget,
-              "(buffer/bit buffer index)",
-              "Gets the bit at the given bit-index. Returns true if the bit is set, false if not.") {
+              "(buffer/bit buf index)",
+              "Gets the bit at the given bit-index in a buffer, `buf`. "
```
pepe: Dto. for renaming

src/core/buffer.c
```diff
@@ -587,8 +594,9 @@ JANET_CORE_FN(cfun_buffer_bitget,
 }
 
 JANET_CORE_FN(cfun_buffer_bittoggle,
-              "(buffer/bit-toggle buffer index)",
-              "Toggles the bit at the given bit index in buffer. Returns the buffer.") {
+              "(buffer/bit-toggle buf index)",
+              "Toggles the bit at the given bit-index in a buffer, `buf`. "
```
pepe: Dto. for renaming

src/core/buffer.c
```diff
@@ -641,9 +649,9 @@ JANET_CORE_FN(cfun_buffer_blit,
 }
 
 JANET_CORE_FN(cfun_buffer_format,
-              "(buffer/format buffer format & args)",
-              "Snprintf like functionality for printing values into a buffer. Returns "
-              "the modified buffer.") {
+              "(buffer/format buf fmt & args)",
+              "Functionality like snprintf for printing values into a "
```
pepe: Maybe:

Formats `fmt`, similar to `sprintf`, with `args` into a buffer `buf`. Returns `buf`.

But not too sure, here. Just a maybe.

pepe (commented):
All comments are more about the taste, than the validity.

src/core/buffer.c
sogaiu: I like your version better :+1:

src/core/buffer.c
sogaiu: Good idea about renaming the argument.  I'll do that for all four.

Thanks!

sogaiu:
@pepe Thanks!

Your comments helped a lot :+1:

pepe:
It is my honor and my pleasure to help you, as always

pepe (approved):
LGTM

sogaiu:
I totally missed that there were conflicts :sweat_smile: 

Should be ok now :crossed_fingers:

PR #1859, opened 2026-09-10
Tweak some buffer parameter docstrings

sogaiu:
This PR has suggestions for some functions that currently have `buffer` as a parameter name (but are not `buffer/` functions as there is a [separate PR](https://github.com/janet-lang/janet/pull/1858) for those).

Some changes include:

* As with https://github.com/janet-lang/janet/pull/1858, the parameter name `buf` is used instead of `buffer`.  This is slightly shorter but also avoids a string that is used for a built-in function [1].
* Some other parameter names were also shortened (e.g. `value` -> `val`).
* `val` is now mentioned in a docstrings for a bit more clarity.
* "end of stream" was changed to "end-of-stream" so that `ev/read` and `ev/chunk` were a little more consistent.
* Wording was altered in a few locations to decrease some docstring lengths (e.g. "Will raise" -> "Raises")
* Ordering and formatting of `int/to-bytes` was changed a bit to make reading the source a bit easier.  Also, `-` was used in the list as a separator between names and descriptions to be more consistent with most of the other docstrings.

---

[1] As mentioned in #1858, this can be nicer for searching as well.

### DIFF
--- a/src/core/ev.c
+++ b/src/core/ev.c
@@ -3408,13 +3408,14 @@ JANET_CORE_FN(janet_cfun_stream_close,
 }
 
 JANET_CORE_FN(janet_cfun_stream_read,
-              "(ev/read stream n &opt buffer timeout)",
-              "Read up to n bytes into a buffer asynchronously from a stream. `n` can also be the keyword "
-              "`:all` to read into the buffer until end of stream. "
-              "Optionally provide a buffer to write into "
-              "as well as a timeout in seconds after which to cancel the operation and raise an error. "
-              "Returns the buffer if the read was successful or nil if end-of-stream reached. Will raise an "
-              "error if there are problems with the IO operation.") {
+              "(ev/read stream n &opt buf timeout)",
+              "Read at most `n` bytes into a buffer asynchronously from a "
+              "stream. `n` can also be `:all` to read into the buffer until "
+              "end of stream. Optionally provide a buffer, `buf`, to write "
+              "into and `timeout` in seconds after which to cancel the "
+              "operation and raise an error. Returns the buffer if the read "
+              "was successful or nil if end-of-stream reached. Raises "
+              "an error if there were problems with the IO operation.") {
     janet_arity(argc, 2, 4);
     JanetStream *stream = janet_getabstract(argv, 0, &janet_stream_type);
     janet_stream_flags(stream, JANET_STREAM_READABLE);
@@ -3431,9 +3432,10 @@ JANET_CORE_FN(janet_cfun_stream_read,
 }
 
 JANET_CORE_FN(janet_cfun_stream_chunk,
-              "(ev/chunk stream n &opt buffer timeout)",
-              "Same as ev/read, but will not return early if less than n bytes are available. If an end of "
-              "stream is reached, will also return early with the collected bytes.") {
+              "(ev/chunk stream n &opt buf timeout)",
+              "Same as `ev/read`, but does not return early if less than "
+              "`n` bytes are available. If an end-of-stream is reached, "
+              "returns early with the collected bytes.") {
     janet_arity(argc, 2, 4);
     JanetStream *stream = janet_getabstract(argv, 0, &janet_stream_type);
     janet_stream_flags(stream, JANET_STREAM_READABLE);
--- a/src/core/inttypes.c
+++ b/src/core/inttypes.c
@@ -235,14 +235,17 @@ JANET_CORE_FN(cfun_to_number,
 }
 
 JANET_CORE_FN(cfun_to_bytes,
-              "(int/to-bytes value &opt endianness buffer)",
-              "Write the bytes of an `int/s64` or `int/u64` into a buffer.\n"
-              "The `buffer` parameter specifies an existing buffer to write to, if unset a new buffer will be created.\n"
+              "(int/to-bytes val &opt endianness buf)",
+              "Write the bytes of an `int/s64` or `int/u64`, `val`, into a "
+              "buffer. The optional `buf` parameter is an existing buffer "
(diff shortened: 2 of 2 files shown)


src/core/marsh.c
```diff
@@ -1679,13 +1679,14 @@ JANET_CORE_FN(cfun_env_lookup,
...
 
 JANET_CORE_FN(cfun_marshal,
-              "(marshal x &opt reverse-lookup buffer no-cycles)",
-              "Marshal a value into a buffer and return the buffer. The buffer "
-              "can then later be unmarshalled to reconstruct the initial value. "
-              "Optionally, one can pass in a reverse lookup table to not marshal "
-              "aliased values that are found in the table. Then a forward "
-              "lookup table can be used to recover the original value when "
-              "unmarshalling.") {
+              "(marshal x &opt rev-lookup buf no-cycles)",
```
pepe, line 1682: Maybe add the information about `no-cycles` removing the possibility of cyclic relations in marshaled value?

pepe (commented):
Just for your consideration.

src/core/marsh.c
sogaiu: Yeah, I agree it would be good to cover `no-cycles`.

Unfortunately, I don't know how it works and didn't manage to figure it out, so I was leaving it alone :)

pepe: The best I am able I already came with :-D: removes the possibility of cyclic relations in `x`, maybe?

sogaiu: I'll make an issue at least.

sogaiu: Ok, #1861 has been made, so I'm going to resolve this for now.

sogaiu:
@pepe Thanks!

I agree with your comment but didn't figure out what to do for this PR so I've made #1861 as a reminder to try to increase the chances that `no-cycles` gets documented.

sogaiu:
Added some documentation for the `no-cycles` parameter based on feedback in [this comment](https://github.com/janet-lang/janet/issues/1861#issuecomment-5626109901).

src/core/marsh.c
```diff
@@ -1686,7 +1686,10 @@ JANET_CORE_FN(cfun_marshal,
               "specify aliased values to not marshal. Then a forward lookup "
               "table can be used to recover the original value when "
               "unmarshalling. If optional argument `buf` is provided, "
-              "marshal into and return that instead.") {
+              "marshal into and return that instead. The optional "
```
pepe, line 1689: This does not sound quite right to me. I would like to know what it actually does, rather than what it helps with when invoking another function.

I’ll think about it some more and try to come up with my own proposal later today.

sogaiu: Good luck :)

FWIW, when I looked into it earlier, I found [this](https://github.com/janet-lang/janet/blob/e6194a81880dfdda5d3054914b886b0932c0fdd3/src/core/marsh.c#L1702):

```c
    janet_arity(argc, 1, 4);
    // ...
    uint32_t flags = 0;
    // ...
    if (argc > 3 && janet_truthy(argv[3])) {
        flags |= JANET_MARSHAL_NO_CYCLES;
```

that seems to have an effect [here](https://github.com/janet-lang/janet/blob/e6194a81880dfdda5d3054914b886b0932c0fdd3/src/core/marsh.c#L703):

```c
void janet_marshal(
    JanetBuffer *buf,
    Janet x,
    JanetTable *rreg,
    int flags) {
    // ...
    st.maybe_cycles = !(flags & JANET_MARSHAL_NO_CYCLES);
```

which in turn has effects [here](https://github.com/janet-lang/janet/blob/e6194a81880dfdda5d3054914b886b0932c0fdd3/src/core/marsh.c#L424-L428):

```c
#define MARK_SEEN() \
    do { if (st->maybe_cycles) { \
        janet_table_put(&st->seen, x, janet_wrap_integer(st->nextid++)); \
    } } while (0)
#endif
```

and:

[here](https://github.com/janet-lang/janet/blob/e6194a81880dfdda5d3054914b886b0932c0fdd3/src/core/marsh.c#L497-L504):

```c
        Janet check;
        if (st->maybe_cycles) {
            check = janet_table_get(&st->seen, x);
            if (janet_checkint(check)) {
                pushbyte(st, LB_REFERENCE);
                pushint(st, janet_unwrap_integer(check));
                return;
            }
        }
```

`MARK_SEEN` (the macro above) is called in quite a few places, mostly inside [`marshal_one`](https://github.com/janet-lang/janet/blob/e6194a81880dfdda5d3054914b886b0932c0fdd3/src/core/marsh.c#L468-L690).

sogaiu: @pepe Did you have any luck?

I'd like to not hold up the changes that are ok (i.e. the ones that don't have to do with `marshal`).

May be we can consider one of the following:

1. keep [this issue](https://github.com/janet-lang/janet/issues/1861) open as a reminder to further improve `marhsal`'s docstring, or
2. remove the portions of this PR that have to do with `marshal`, or
3. Something else.

What do you think?

sogaiu: I've gone for option 2, i.e. this PR no longer touches `marsh.c`.

sogaiu:
I've backed at all changes to `marsh.c` and will work on those bits some other time.

I've also backed out changes to `ffi.c` to facilitate work on another PR.

Thanks @pepe!

src/core/marsh.c
sogaiu: @pepe I made a [new PR](https://github.com/janet-lang/janet/pull/1873) with the in-progress changes to the functions in `marsh.c` that used to be in this PR.

I mentioned in the new PR that the description for the `no-cycles` parameter might be improved.

PR #1860, opened 2026-09-10
Modify os/clock and string/format docstrings

sogaiu:
This PR contains some suggested changes to the docstrings for `os/clock` and `string/format`.

`os/clock` changes include:

* Changing the parameter names to `clock` and `as`.  These are both shorter and seem more descriptive.  The original `source` and `format` happen to be used in other contexts with different meanings as well, so avoiding them seemed better.
* Using more backticks to surround some items.

`string/format` changes include:

* Changing the parameter names to `fmt` and `args`.  `fmt` and `args` seem more in line with other functions / macros that have similar arguments such as `buffer/format`, `assertf`, `errorf`, and the like.
* Modifying some wording, grammar, punctuation, and formatting.

In both cases, I modified the markup used for list items to be more aligned with what we're doing in other docstrings.


pepe (approved):
LGTM. I like the `string/format` a lot.

sogaiu:
Oh wait, may be I should go for `args` over `vals`...

sogaiu:
Ok, I went with `args` and updated the original description.

pepe (approved):
LGTM

pepe (approved):
LGTM

PR #1862, opened 2026-09-11
Surround more things with backticks

sogaiu:
This PR contains mostly suggestions regarding formatting of some docstrings.

Some changes include:

* Surround more things with backticks.  This includes Janet function names, file path-related things such as `<time.h>`, keywords, etc.
* Append parens to C functions [1].
* Adjust list item markup so some lists render a bit nicer on the core API page.
* `filewatch/new`'s `channel` parameter name was changed to `chan`.  I'm thinking to do this elsewhere as well, but thought I'd attend to this instance since there were other things to do too and merging of PRs seems a bit less problematic this way.
* Reflow and whitespace adjustments.

---

[1] If we follow this convention, I think it will help us to programmatically distinguish between Janet function / macro names and C constructs.  This in turn may help us create cross-references.


bakpakin:
This sounds good to me. The original intention of the backticks were to be exclusively for function arguments or links to other bindings, but really any pattern that we can unambiguously convert to hyperlinks in the documentation would be good.

PR #1863, opened 2026-09-11
Modify string/find* docstrings

sogaiu:
This PR suggests changes to the docstrings of `string/find` and `string/find-all`.

Changes include:

* Uses the parameter `start` instead of `start-index`.  This name is used elsewhere in a similar fashion (e.g. `array/slice`, `for`, `peg/find`, `string/split`, etc.) and it's shorter :)
* How `start` is used in the code is also described.


pepe (approved):
LGTM

PR #1864, opened 2026-09-11
Modify debug/unbreak and tuple/setmap docstrings

sogaiu:
This PR suggests changes to the docstrings of `debug/unbreak` and `tuple/setmap` docstrings.

Changes include:

* Uses the parameter `col` instead of `column`.  This name is used elsewhere in a similar fashion (e.g. `bad-compile`, `debug/break`, `parser/where`, and `warn-compile`) and it's shorter :)
* `tup` is now mentioned in `tuple/setmap`'s docstring.
* A number of items were surrounded in backticks.

Regarding the `source` parameter (hi @pepe!), what this can be is not described in this docstring, but it's somewhat demonstrated in that of `debug/break`.  I intend to return to `debug/unbreak`'s docstring after or around the time I work on `debug/break`, so for the moment, my plan is to not explain what `source` is in this PR.

### DIFF
--- a/src/core/debug.c
+++ b/src/core/debug.c
@@ -232,10 +232,10 @@ JANET_CORE_FN(cfun_debug_break,
 }
 
 JANET_CORE_FN(cfun_debug_unbreak,
-              "(debug/unbreak source line column)",
-              "Remove a breakpoint with a source key at a given line and column. "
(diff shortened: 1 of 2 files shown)


pepe (approved):
LGTM

PR #1865, opened 2026-09-11
Rename channel parameter as chan

sogaiu:
This PR suggests changes to some `ev/` functions.

Changes include:

* Rename `channel` parameter as `chan`.  It is shorter and seems unambiguous.  It was also being used already in a few spots.
* Rename `value` parameter as `val`.  This is part of an ongoing effort.
* Spell out a bit more what `ev/full` returns.
* Mention the `chan` parameter in docstrings.
* Shorten some text in a few spots (e.g. replace "Returns the channel" with "Returns `chan`").
* Some reflow.


pepe (approved):
LGTM

Issue #1861, opened 2026-09-10
marshal's no-cycles parameter appears undocumented


sogaiu:
While working on #1859, @pepe pointed out that the `no-cycles` parameter for `marshal` is not documented.

I made an attempt at understanding how it works, but did not come away with a confident answer.

This issue is a reminder to make another try to document it.

pepe:
Just for the record, I am marshaling everyday, all day and never used the parameter.

Yet, when I am thinking about it, maybe I should, actually :-D.

bakpakin:
The `no-cycles` parameter will make `unmarshal` faster if we know there are no cycles. If there are cycles, you will get a stack overflow (inside Janet, not a segfault). Useful trick for speeding things up in some cases.

PR #1873, opened 2026-09-12
Modify some marshal docstrings

sogaiu:
This PR suggests some changes to the docstrings of functions in `marsh.c`.

Changes include:

* Use `buf` instead of `buffer` as a parameter name.  This is part of the effort in #1858 and #1859.
* Use `rev-lookup` instead of `reverse-lookup`.  It is shorter and seem unambiguous.
* Surround the reference to `invert` with backticks.
* More parameter names are also now mentioned in the docstrings (e.g. `env`, `x`, `buf`, `no-cycles`, and `lookup`).

It's possible the `no-cycles` parameter's description could be improved.  See #1861 and [this comment](https://github.com/janet-lang/janet/pull/1859#discussion_r3987383169).

### DIFF
--- a/src/core/marsh.c
+++ b/src/core/marsh.c
@@ -1670,22 +1670,26 @@ Janet janet_unmarshal(
 
 JANET_CORE_FN(cfun_env_lookup,
               "(env-lookup env)",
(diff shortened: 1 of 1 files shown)


pepe (approved):
LGTM

PR #1874, opened 2026-09-12
Remove dangling janet_getfloat declaration

McSinyx:
It was not removed along with the function definition.

Fixes: ec48673f7c66 ("Revert float interface")
References: 23e4ad4bd3b4 ("Add ... float getters for consistency")


McSinyx:
Thanks!

PR #1877, opened 2026-09-14
Tweak docstrings for os/sleep, os/posix-chroot

sogaiu:
This PR suggests changes to the docstrings for `os/posix-chroot` and `os/sleep`.

`os/posix-chroot`:

* Use the parameter name `dir` instead of `dirname`.  `dir` is also used by `os/dir` for a similar purpose so having both be the shorter choice (that seemed unambiguous) seemed better.
* The accompanying prose was also modified to mention that `os/sleep` suspends the current thread instead of "the program".

`os/sleep`:

* Use the parameter name `sec` instead of `n`.  `sec` is used by `ev/deadline`, `ev/sleep`, and `ev/with-deadline` for a similar meaning anyway and `n` might have suggested integer values only (as hinted at by the clarifying text in the current docstring).
* The accompanying prose was also modified to explicitly point out that the value for `sec` can have a fractional part.


pepe (approved):
LGTM

PR #1878, opened 2026-09-14
Use ffi-native as parameter name

sogaiu:
This PR mostly suggests using the parameter name `ffi-native` over `native` in the docstrings of the functions `ffi/close` and `ffi/lookup`.

The name `native` is also the name of a built-in function and `ffi-native` isn't so choosing the latter may yield less problematic results in the context of making linking / referencing identifiers of our docstrings work.

For reference, we made a similar change in #1872 for `type` -> `ffi-type`.

Some other changes include:

* The parameter name `ffi-native` is mentioned in both docstrings.
* The parameter name `symbol-name` was also replaced with `sym-name` as this seemed unambiguous and shorter.  The parameter was also mentioned in the docsting.
* Attempts were made to shorten the prose a bit (e.g. "will have" -> "has").


pepe (approved):
LGTM

PR #1879, opened 2026-09-14
Use ksvs as parameter name

sogaiu:
This PR mostly suggests using the parameter name `ksvs` over `kvs` in the docstrings of the functions `table`, `struct`, and `struct/with-proto`.

The name `kvs` is also the name of a built-in function and `ksvs` isn't (see #1878 for additional reasoning).

The prose for the docstrings of `table` and `struct` was tweaked slightly to reduce length as well.


pepe (approved):
LGTM

PR #1880, opened 2026-09-17
Tweak some docstrings

sogaiu:
This PR collects together a number of small changes.

Changes include:

* `invert` - fix a markup typo (sorry!)
* `put-in` - use `val` instead of `v`.  This is both more consistent with recent parameter choices and there is an internal `v` which was shadowing the parameter (no operational issue).
* `interpose` - use `val` instead of `sep`.  `sep` was being used elsewhere as a string and for `interpose` the argument in question can be a value of any type so `val` seemed more fitting.
* `net/setsockopt` - align with other docstrings to use capitalization for ordinary sentences, use of `*` for indicating list items, and use `-` to separate names from descriptions.
* `os/clock` - do not use a comma as part of expressing a Janet tuple and reword slightly.


pepe (approved):
LGTM

PR #1881, opened 2026-09-18
Use square brackets for data tuples in docstrings

sogaiu:
This PR is mostly about suggesting the use of square bracket tuples for "data tuples" in docstrings.

`ev/select` and `os/clock` do this already and it seems less confusing than using "round paren tuples" as use of the latter can initially give the impression of a call.

Changes include:

* Use square bracket tuples in the docstrings for `module/find`, `compile`, `math/frexp`, and `tuple/sourcemap`.
* Mention some unmentioned parameters in some docstrings.
* For `compile`, mention and try to provide explanations for the `env` and `source` parameters.
* Remove a note in `compile`'s docstring "Pair the compile function with parsing functionality to implement eval." as it seems out of place.  I can put this back if it seems better to leave it.
* Surround some keywords in backticks for consistency [1].

---

[1] If it's better to switch to not having keywords being surrounded by backticks, I can try to move things in that direction.  My impression is that it aids readability via `doc` and eventually will help on the website when rendering of code spans is improved / implemented, but it does take a little more space.


pepe (approved):
LGTM

